PR #1252, opened 2014-12-31
minor improvements related to SyncFieldMixin and Device.save()

xor-xor:
The actual behavior of `Device.save()` is rather confusing, mostly because of the `SavePrioritized` mixin (some changes are silently dropped, which is hard to track if you're unaware of this mechanism) - therefore, I think it's good to provide some explaination for this in the source code.

Also, I've improved the description of `SyncFieldMixin` and moved some hard-coded stuff to settings.


zefciu:
:ok_hand:

PR #1258, opened 2015-01-05
Better scanning of Juniper stacks

ojagodzinski:
- different stack models based on used switches
- improvement in fetching mac addresses from CLI

### DIFF
--- a/src/ralph/scan/plugins/ssh_juniper.py
+++ b/src/ralph/scan/plugins/ssh_juniper.py
@@ -45,6 +45,16 @@ def _get_hostname(ssh):
                 pass  # unexpected result... maybe next line will be ok...
 
 
+def _get_model(ssh):
+    for line in _ssh_lines(ssh, 'show version'):
+        if line.lower().startswith('model'):
+            line_chunks = line.split(':')
+            try:
+                return line_chunks[1].strip()
+            except IndexError:
+                pass  # unexpected result... maybe next line will be ok...
+
+
 def _get_mac_addresses(ssh):
     mac_addresses = []
     for line in _ssh_lines(ssh, 'show chassis mac-addresses'):
@@ -111,6 +121,7 @@ def _ssh_juniper(ssh, ip_address):
         'management_ip_addresses': [ip_address],
     }
     hostname = _get_hostname(ssh)
+    model = _get_model(ssh)
     mac_addresses = _get_mac_addresses(ssh)
     if hostname:
         device['hostname'] = hostname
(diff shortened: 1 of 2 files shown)


xor-xor:
@ojagodzinski: please fix errors reported by Travis (two unit tests have failed - either fix your code or update them to reflect your changes). Thanks!

vi4m:
What about netconf protocol ? Are we gonna use it eventually?

ojagodzinski:
NETCONF communication with responses in XML/JSON (supported in latest Juniper firmwares) is only reasonable direction we can take. Having all response data structured would make parsing much easier. However my main concern is it is't tuned on by default and doesn't cover 100% of commands.

It can be some secondary mechanism compared with "cli scraping" results. As You can see it [here](https://github.com/Juniper/py-junos-eznc) getting basic facts can be done in 2 lines of code.

Or... just throw away any "cli scraping" and demand modern firmware with NETCONF service on devices to have reliable responses. As second thing is one command, first requirement can be much bigger problem.

src/ralph/scan/plugins/ssh_juniper.py
```diff
@@ -45,6 +45,16 @@ def _get_hostname(ssh):
                 pass  # unexpected result... maybe next line will be ok...
 
 
+def _get_model(ssh):
+    for line in _ssh_lines(ssh, 'show version'):
+        if line.lower().startswith('model'):
```
vi4m: Re expression is much better to document required input string, and for future updates 

re.match('model:(.*)', 'model: dupa').groups()

vi4m:
@ojagodzinski did you follow our hints with re ?

ojagodzinski:
```python
def _get_hostname(ssh):
    for line in _ssh_lines(ssh, 'show version'):
        compar = re.match('.*hostname: (.*)', line.lower())
        if compar is not None:
            return compar.groups[0]


def _get_model(ssh):
    for line in _ssh_lines(ssh, 'show version'):
        compar = re.match('.*model: (.*)', line.lower())
        if compar is not None:
            return compar.groups[0]


def _get_mac_addresses(ssh):
    mac_addresses = []
    for line in _ssh_lines(ssh, 'show chassis mac-addresses'):
        line = line.lower()
        compar = re.match('.*(base|public) address[ ]*(.*)', line.lower())
        if compar is not None:
            mac_addresses.append(MACAddressField.normalize(compar.groups()[1]))
    return mac_addresses
```

Something like this?

xor-xor:
@vi4m - can you reply to the previous comment..?

vi4m:
yes

vi4m:
Please fix and reopen

vi4m:
PING

PR #1260, opened 2015-01-07
Fixed correct PDU visualization

ar4s:
Changes:
- added directive (``pdu-item``) responsible for correct display of PDU,
- added ``.pdu .item`` class to CSS.


xor-xor:
:ok:

xliiv:
:ok:

PR #1266, opened 2015-01-12
added quasi-serializer for location data

xliiv:
Added Device method which serialize location data.

### DIFF
--- a/src/ralph/discovery/models_device.py
+++ b/src/ralph/discovery/models_device.py
@@ -968,6 +968,18 @@ class Device(
     def url(self):
         return reverse('search', args=('info', self.id))
 
+    def serialize_location_data(self):
+        data = {
+            'barcode': self.barcode,
+            'dc': self.dc,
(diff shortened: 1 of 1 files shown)


vi4m:
LGTM

src/ralph/discovery/models_device.py
```diff
@@ -968,6 +968,18 @@ def get_property_set(self):
     def url(self):
         return reverse('search', args=('info', self.id))
 
+    def serialize_location_data(self):
+        data = {
```
zefciu: Why not just: `return {...}`

src/ralph/discovery/models_device.py
```diff
@@ -968,6 +968,18 @@ def get_property_set(self):
...
+            'barcode': self.barcode,
+            'dc': self.dc,
+            # 'server_room': 'default_server_room',
```
quamilek: It should be commented?

PR #1268, opened 2015-01-13
Scan form create IPAddress object when does not exist.

quamilek:
Earlier when scan IPAddressed which does not exist in database, it was necessary to add it manually through the admin panel. 
Now, the scan view create automatically IPAddress when does not exist in database.


zefciu:
:ok:

xor-xor:
:ok:

PR #1274, opened 2015-01-16
added database to scrooge api

mkurek:
added databases to scrooge api to show them as components in scrooge components view and to make ability to charge for single database server.


kula1922:
:ok:

xor-xor:
:ok: - although according to @vi4m, we should follow new rules regarding descriptions of commits and pull requests (ask him for details).

PR #1276, opened 2015-01-18
New API entrypoint - change-ip-address

andrzej-jankowski:
Added new entry point to API to provide possibility to change IP address after deployment.

### DIFF
--- a/src/ralph/deployment/api.py
+++ b/src/ralph/deployment/api.py
@@ -8,15 +8,23 @@ from __future__ import unicode_literals
 
 from django.conf import settings
 from tastypie import fields
+from tastypie import http
 from tastypie.authentication import ApiKeyAuthentication
+from tastypie.authorization import DjangoAuthorization
+from tastypie.bundle import Bundle
 from tastypie.cache import SimpleCache
 from tastypie.constants import ALL, ALL_WITH_RELATIONS
+from tastypie.exceptions import ImmediateHttpResponse
 from tastypie.resources import ModelResource as MResource
+from tastypie.resources import Resource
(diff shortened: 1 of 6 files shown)


xor-xor:
:ok: - although according to @vi4m, we should follow new rules regarding descriptions of commits and pull requests (ask him for details).

quamilek:
:ok:

src/ralph/deployment/api.py
```diff
@@ -70,3 +78,69 @@ class Meta:
             timeframe=TIMEFRAME,
             expiration=EXPIRATION,
         )
+
+
+class IPAddressChange(object):
+
+    __slots__ = ['current_ip_address', 'new_ip_address']
```
zefciu: Do we really need to handle so much of these objects that we need slots? According to people smarter than me ``__slots__`` should be used only for performance reasons.

I believe namedtuple would suit better here:

    IPAddressChange = namedtuple(['current_ip_address', 'new_ip_address']

andrzej-jankowski: Tastypie needs class instance to create bundle object.

src/ralph/deployment/tests/functional/tests_api.py
```diff
@@ -68,7 +76,7 @@ class TestVMCreation(UserTestCase):
 
     fixtures=['vm_creation_setup']
 
-    def test_correct(self): 
+    def test_correct(self):
```
vi4m: All tests: 

Comply with test naming : test_[object]_should_[action]_when[something]

put it in the name instead of docstrings

PR #1277, opened 2015-01-19
Renamed tests to comply with test/should/when

zefciu:
Just a set of testcase renames to comply with our policies. Should clean up the Andrzej's PR.


xor-xor:
:ok:

vi4m:
LGTM

andrzej-jankowski:
:ok:

PR #1278, opened 2015-01-19
Added new entry point to API to provide possibility to change IP address

andrzej-jankowski:
* Added new API entry point - /api/v0.9/change-ip-address/
* Reason: Possibility to change IP address after server deployment.


xor-xor:
:ok:

quamilek:
:ok:

PR #1282, opened 2015-01-21
Fix button toogle on scan form.

quamilek:
Change the order of loading ajax_select.js file. 
jquery_ui.js loaded from ajax_select.js must be loaded before the bootstrap.js, another order causes error.


xor-xor:
:ok:

zefciu:
:ok:

PR #1283, opened 2015-01-21
Visualization improvments

quamilek:
Changes:
* include angular scripts if necessary
* reduced requests when user rotate rack
* removed unused device.css
* small refactor of variables name


xliiv:
:ok:

PR #1284, opened 2015-01-21
Fix in device admin panel.

andrzej-jankowski:
Parent and management fields are now readonly when an asset is assigned. Before this change we could't save form.

### DIFF
--- a/src/ralph/discovery/admin.py
+++ b/src/ralph/discovery/admin.py
@@ -27,10 +27,7 @@ from ralph.business.admin import RolePropertyValueInline
 from ralph.discovery import models
 from ralph.discovery import models_device
 from ralph.ui.forms.network import NetworkForm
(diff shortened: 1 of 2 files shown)


src/ralph/discovery/admin.py
```diff
@@ -399,6 +375,14 @@ class DeviceAdmin(ModelAdmin):
         'model': ['^name', ],
     }
 
+    def get_readonly_fields(self, request, obj=None):
+        ro_fields = super(DeviceAdmin, self).get_readonly_fields(request, obj)
+        if obj:
```
quamilek:
```
if obj and obj.get_asset()
    ro_fields = ro_fields + ('parent', 'management',)
return ro_fields
```
Looks better? :)

quamilek:
:ok:

xliiv:
:ok:

PR #1285, opened 2015-01-22
Fix for MAC detecting in IDRAC plugin.

andrzej-jankowski:
For some configurations IDRAC plugin does not work because an error in MAC detecting function occurs. This function was changed to provide more stable detection. Also our tests sample data was updated.


vi4m:
LGTM

quamilek:
:ok:

PR #1286, opened 2015-01-26
New CiscoSSHHandler added to provide better support for Cisco switches.

andrzej-jankowski:
* New connector added - CiscoSSHHandler
* Change ssh_cisco_catalyst_plugin to use new connetor
* ssh_cisco_catalyst_plugin now does not send `terminal length` command - it's now handled by connector class

### DIFF
--- a/src/ralph/util/network.py
+++ b/src/ralph/util/network.py
@@ -9,9 +9,11 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import logging
 import socket
-import sys
 import StringIO
+import sys
+import time
 
 from dns.exception import DNSException
 from lck.cache import memoize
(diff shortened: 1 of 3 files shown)


src/ralph/util/network.py
```diff
@@ -118,6 +120,63 @@ def connect_ssh(
...
+        self.channel.send("\n")
+
+    def cisco_command(self, command):
+        self.logger.info("Running command: {}".format(command))
+        self.channel.send(command + "\n")
+        time.sleep(2)
+        buff = ['']
+        while True:
+            chunk = self.channel.recv(9999)
+            lines = chunk.split("\n")
```
kula1922: self.channel.recv(9999).split("\n")

zefciu:
:ok:

xliiv:
:ok:

PR #1287, opened 2015-01-26
Xen dom0 is incorrectly classified as virtual host

XANi:
Xen dom0 should be classified as physical machine:

    xen3:~☠ facter -p  |grep -P '(virtual|^productname)' 
    is_virtual => false
    productname => IBM System x3550 -[7978ZBX]-
    virtual => xen0

btw. is there a reason fact 'is_virtual' is not used for that ? Both facts come from same place ( https://github.com/puppetlabs/facter/blob/master/lib/facter/virtual.rb ).

### DIFF
--- a/src/ralph/scan/plugins/puppet.py
+++ b/src/ralph/scan/plugins/puppet.py
@@ -213,7 +213,7 @@ def _get_ip_addresses_hostnames_sets(ip):
 
 def _is_host_virtual(facts):
     is_virtual = facts.get('virtual', 'physical') not in (
-        'physical', 'openvz', 'openvzhn',
+        'physical', 'openvz', 'openvzhn', 'xen0',
     )
     if facts.get('manufacturer') == 'Bochs' and not is_virtual:
         facts['virtual'] = 'virtual'


vi4m:
Please add space near ,'xen0', to be pep8 compliant :)

XANi:
Sure but... there was no trailing whitespace there before... did it passed validation before turning on validator ?

XANi:
mkay

    src/ralph/scan/plugins/puppet.py:216:41: E231 missing whitespace after ','
    src/ralph/scan/plugins/puppet.py:216:49: W291 trailing whitespace


what ?

XANi:
Ok, should be fixed now.

vi4m:
Good for me, but just ask @andrzej-jankowski why don't we use is_virtual?

andrzej-jankowski:
I don't know. It's legacy code adopted from old Ralph's discovery system. Maybe in the past fact is_virtual did not exist.

andrzej-jankowski:
:ok:

vi4m:
Thank you for your first pull request @XANi

XANi:
i was planning to do another for puppetdb v3 api but then I noticed you guys are already doing it on develop/ :)

vi4m:
Yes, in fact we are discussing the details behind the certificates storing. you can join the discussion.

vi4m:
Because, for example, there could be multiple puppet db nodes and we have some ideas  how to handle this. I will post some blueprint to collect some comments.

XANi:
On gitter ? Or you have some mailing list?

We generally have one (well, pair of HA nodes) DB per site so it would be either one queue per site, or iterating over all puppetDBs till one of them returns data on node.

Then if more than one have it:

* ignore old ( >1-2 days) ones
*  get one with latest data if difference of time is short, like few hours, meaning node was moved/reinstalled to other site.

Our env have unique certname for each node so there is no case of "we have 2 different nodes named same on different envs"

vi4m:
@andrzej-jankowski what do you think about it

PR #1289, opened 2015-01-27
Updated the installation with cautions.

vi4m:
It's hard to install Ralph by hand, so i added the Docker way
explanation.


andrzej-jankowski:
:ok:

kula1922:
:ok:

PR #1293, opened 2015-01-28
Added editable management IP in 'Addresses' tab

ar4s:
I added form where admins can edit IP management from device edit view in core.


zefciu:
:ok:

andrzej-jankowski:
You should also change this form: https://github.com/allegro/ralph/blob/develop/src/ralph/discovery/admin.py#L418

Remember to add some validation to prevent users adding many management IP addresses to one device. Only one management IP address to one device is possible.

andrzej-jankowski:
:ok:

PR #1296, opened 2015-02-02
Removed duplicated UserFactory

ar4s:
The CMDB use own UserFactory and UserFactory from /ui/tests/global_utils.py.
For this reason each factory have own sequence.


vi4m:
Pismo nieczytelne!

xliiv:
:ok:

andrzej-jankowski:
:ok:

PR #1297, opened 2015-02-03
Fixed ReadOnlyWidget and ReadOnlySelectWidget

andrzej-jankowski:
Widgets shouldn't paste "None" string as a value when value is None -
empty string is much better. Previous behavior produce many validation
errors. E.g. IntegerField validation said that it's not a number -
that's right because "None" it is not a number :)


xliiv:
:ok:

ar4s:
:ok:

PR #1298, opened 2015-02-03
Corrections for scan confirmation

zefciu:
* Added model_type to columns in subdevices
* Subdevice creation is done with priority=215
* Handled asset=None situation


andrzej-jankowski:
:ok:

xliiv:
:ok:

PR #1299, opened 2015-02-03
Management IP form always visible

ar4s:
Changes:
 * removed if statment from template
 * form hs moved to top


andrzej-jankowski:
:ok:

xliiv:
:ok:

PR #1300, opened 2015-02-03
Detaching logical subdevices not found by scan

zefciu:
Previously only physical subdevices not found by scan were detached.
Extended this feature on logical children

### DIFF
--- a/src/ralph/scan/data.py
+++ b/src/ralph/scan/data.py
@@ -58,6 +58,12 @@ UNIQUE_FIELDS_FOR_MERGER = {
 SAVE_PRIORITY = 215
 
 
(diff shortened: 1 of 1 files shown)


src/ralph/scan/data.py
```diff
@@ -58,6 +58,12 @@
 SAVE_PRIORITY = 215
 
 
+def has_logical_children(device):
+    """Returns True if this device has logical children of False if it has
```
andrzej-jankowski: "of False if..." -> OR?

andrzej-jankowski:
:ok:

ar4s:
:ok:

PR #1301, opened 2015-02-04
better scan of Juniper stacks

ojagodzinski:
Changes in Juniper SSH scan plugin to distinguish between different device models in stack.


vi4m:
LGTM

vi4m:
Please stash commits & replace commit message with better one - short line with summary, line empty and in more details what has changed and why.

andrzej-jankowski:
:ok:

PR #1302, opened 2015-02-04
Added two missing features to deploy/clean plugin

zefciu:
* The hostname is now copied from deployment
* Also copying venture/role from deployment
* Factories for venture/role moved to utils


andrzej-jankowski:
:ok:

xliiv:
:ok:

PR #1303, opened 2015-02-04
Improved Xen detection

zefciu:
The Xen servers > 6.5 don't contain 'xen' string in their snmp_name. We
need to detect them using their HTTP response (looking for 'XenServer'
string in the returned HTML document).

Also - renamed the existing Dell detection test, so new tests can be
added.


andrzej-jankowski:
:ok:

ar4s:
:ok:

PR #1304, opened 2015-02-04
Moved the settings in settings.py

zefciu:
Some settings must be put before pluggable apps initialization. If not,
then the plugins that use ralph will crash trying to import things from
ralph.


andrzej-jankowski:
:ok:

ar4s:
:ok:

PR #1305, opened 2015-02-06
Allowed ralph plugins without menu entry and urls

zefciu:
Some apps might not need an entry in menu or their own views (and hence
their own urls.py). This fix will cause them not to crash.


xliiv:
:ok:

ar4s:
:ok:

PR #1306, opened 2015-02-09
Moved property setting to a separate function

zefciu:
We want to set device properties from other places than just save() of
the device form. Thus we move the logic to a separate function.

### DIFF
--- a/src/ralph/discovery/models_device.py
+++ b/src/ralph/discovery/models_device.py
@@ -925,6 +925,30 @@ class Device(
         # the reason (for more, see the source code of SavePrioritized).
         return super(Device, self).save(*args, **kwargs)
 
+    def set_property(self, symbol, value, user):
+        from ralph.business.models import RoleProperty, RolePropertyValue
+        try:
+            p = self.venture_role.roleproperty_set.get(symbol=symbol)
+        except RoleProperty.DoesNotExist:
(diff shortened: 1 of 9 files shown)


andrzej-jankowski:
:ok:

src/ralph/discovery/models_device.py
```diff
@@ -925,6 +925,30 @@ def save(self, sync_fields=True, *args, **kwargs):
         # the reason (for more, see the source code of SavePrioritized).
         return super(Device, self).save(*args, **kwargs)
 
+    def set_property(self, symbol, value, user):
+        from ralph.business.models import RoleProperty, RolePropertyValue
+        try:
+            p = self.venture_role.roleproperty_set.get(symbol=symbol)
+        except RoleProperty.DoesNotExist:
+            p = self.venture.roleproperty_set.get(symbol=symbol)
```
ar4s: why not ``p, _ = self.venture.roleproperty_set.get_or_create(symbol=symbol)``?

andrzej-jankowski: self.venture_role vs self.venture

andrzej-jankowski: It's ok here.

zefciu: I don't know why. That's not my code :D

ar4s:
:ok:

PR #1308, opened 2015-02-10
Changes in deployments plugins order.

andrzej-jankowski:
Added clean plugin as required to each other plugins to make sure that
new venture and role will be assigned to deployed server.
Removed venture and role assignation form clean plugin - it's now not
required.


kula1922:
:ok:

xliiv:
:ok:

PR #1314, opened 2015-02-18
Custom fields mechanism

zefciu:
Added custom fields mechanism. It allows to bind custom attributes to a
content type and access them like normal attributes.

### DIFF
--- a/src/ralph/settings-test-ralph.py
+++ b/src/ralph/settings-test-ralph.py
@@ -18,3 +18,5 @@ PLUGGABLE_APPS = ['cmdb', 'assets']
 SOUTH_TESTS_MIGRATE = False
 
 ASSETS_AUTO_ASSIGN_HOSTNAME = True
+
+INSTALLED_APPS += ['ralph.util.tests']
--- a/src/ralph/util/models.py
+++ b/src/ralph/util/models.py
@@ -10,15 +10,18 @@ from __future__ import unicode_literals
 
 from collections import namedtuple
 
+from dj.choices import Choices
 from django.contrib.auth.models import User
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.contenttypes.generic import GenericForeignKey
 from django.db import models as db
 from django.db.utils import DatabaseError
 from django.dispatch import Signal
+from django.utils.translation import ugettext_lazy as _
 from tastypie.models import create_api_key
 
 from ralph.settings import SYNC_FIELD_MIXIN_NOTIFICATIONS_WHITELIST
 
-from django.contrib.auth.tests import models as auth_test_models
 ChangeTuple = namedtuple('ChangeTuple', ['field', 'old_value', 'new_value'])
 
 
@@ -32,10 +35,6 @@ def create_api_key_ignore_dberrors(*args, **kwargs):
 
 db.signals.post_save.connect(create_api_key_ignore_dberrors, sender=User)
 
-
-# workaround for a unit test bug in Django 1.4.x
-
-del auth_test_models.ProfileTestCase.test_site_profile_not_available
 
 # signal used by SyncFieldMixin for sending notifications on changed fields
 fields_synced_signal = Signal(providing_args=['changes', 'change_author'])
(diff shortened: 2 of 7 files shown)


src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+
+    def load_attributes_if_necessary(self, inst):
+        if self.name not in inst.__dict__:
+            inst.load_custom_attributes()
+
+    def is_option_based(self):
+        """Returns true if this attribute uses options"""
+        return self.type in {CustomAttributeTypes.SINGLE_CHOICE}
+
+    def __str__(self):
```
ar4s: please use ``__unicode__``

src/ralph/util/tests/test_custom_attributes.py
```diff
@@ -0,0 +1,104 @@
...
+        self.device.save()
+        device = MyDevice.objects.get(pk=self.device.pk)
+        self.assertEqual(device.ca2.value, 'Option 1')
+
+    def test_choice_value_raises_valueerror_when_nonexistent_option_used(self):
+        with self.assertRaises(ValueError):
+            self.device.ca2 = 'Option X'
+
+    def test_value_is_none_if_unset(self):
+        self.assertIsNone(self.device.ca2)
```
xliiv: where is "unset" part?

src/ralph/settings-test-ralph.py
```diff
@@ -18,3 +18,5 @@
 SOUTH_TESTS_MIGRATE = False
 
 ASSETS_AUTO_ASSIGN_HOSTNAME = True
+
+INSTALLED_APPS += ['ralph.util.tests']
```
xliiv: why this?

mkurek: It's probably added here to use `MyDevice` model in unit tests. Please look at Scrooge solution: https://github.com/allegro/ralph_pricing/blob/develop/src/ralph_scrooge/tests/models.py#L22 - maybe adding `Meta` with `app_label` will be sufficient here (instead of adding `ralph.util.tests` to `INSTALLED_APPS`)

src/ralph/util/tests/test_custom_attributes.py
```diff
@@ -0,0 +1,104 @@
...
+        self.opt2 = CustomAttributeOption(
+            custom_attribute=self.ca2,
+            value='Option 2',
+        )
+        self.opt2.save()
+        self.opt3 = CustomAttributeOption(
+            custom_attribute=self.ca2,
+            value='Option 3',
+        )
+        self.opt3.save()
```
ar4s: Where you use ``self.opt1``, ``self.opt2``, ``self.opt3``?

ar4s:
I don't see the benefits of the use of ``metaclass``.. Could you explain or give some example how it works?

src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+        verbose_name=_('Displayed name'),
+        max_length=255,
+    )
+    type = db.PositiveIntegerField(
+        choices=CustomAttributeTypes()
+    )
+    content_type = db.ForeignKey(ContentType)
+    required = db.BooleanField()
+
+    def load_attributes_if_necessary(self, inst):
```
andrzej-jankowski: Please use more descriptive vars names (inst).

src/ralph/settings-test-ralph.py
zefciu: Doesn't work

src/ralph/util/tests/test_custom_attributes.py
zefciu: Nowwhere. But I want this to be realistic. I need several opts here.

andrzej-jankowski:
Maybe we should create some feature branch for this...

xliiv:
+1 for seperated branch

PR #1316, opened 2015-02-23
Custom fields mechanism

zefciu:
Added custom fields mechanism. It allows to bind custom attributes to a
content type and access them like normal attributes.

### DIFF
--- a/src/ralph/util/models.py
+++ b/src/ralph/util/models.py
@@ -10,15 +10,18 @@ from __future__ import unicode_literals
 
 from collections import namedtuple
 
+from dj.choices import Choices
 from django.contrib.auth.models import User
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.contenttypes.generic import GenericForeignKey
 from django.db import models as db
 from django.db.utils import DatabaseError
 from django.dispatch import Signal
+from django.utils.translation import ugettext_lazy as _
 from tastypie.models import create_api_key
 
 from ralph.settings import SYNC_FIELD_MIXIN_NOTIFICATIONS_WHITELIST
 
-from django.contrib.auth.tests import models as auth_test_models
 ChangeTuple = namedtuple('ChangeTuple', ['field', 'old_value', 'new_value'])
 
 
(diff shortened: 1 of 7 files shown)


src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+    def content_type(cls):
+        return ContentType.objects.get_for_model(cls)
+
+    @property
+    def custom_attributes(cls):
+        if cls._custom_attributes is None:
+            cls.refresh_custom_attributes()
+        return cls._custom_attributes
+
+    def refresh_custom_attributes(cls):
```
kula1922: docs

src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+        verbose_name=_('Displayed name'),
+        max_length=255,
+    )
+    type = db.PositiveIntegerField(
+        choices=CustomAttributeTypes()
+    )
+    content_type = db.ForeignKey(ContentType)
+    required = db.BooleanField()
+
+    def load_attributes_if_necessary(self, inst):
```
kula1922: docs

src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+                attribute=self,
+                value=value,
+                object_id=inst.pk,
+                content_type=ContentType.objects.get_for_model(type(inst)),
+            )
+            value_object.dirty = True
+            inst.ca_values[self.type][self.pk] = value_object
+        else:
+            value_object.value = value
+            value_object.dirty = True
```
kula1922: value_object.dirty = True for each case?

src/ralph/util/models.py
```diff
@@ -109,3 +108,194 @@ def save(self, mute=False, visited=None, *args, **kwargs):
...
+            for value in values_set:
+                value.dirty = False
+                value.to_delete = False
+                self.ca_values[type_id][value.attribute.pk] = value
+        for ca in type(self).custom_attributes:
+            value = self.ca_values[ca.type].get(ca.pk)
+            if value is not None:
+                self.__dict__[ca.name] = value.value
+            else:
+                self.__dict__[ca.name] = None
```
kula1922: self.__dict__[ca.name] = getattr(value, 'value', None)

zefciu: I really really don't like this idea, as it would exploit the fact that getattr is broken.

vi4m:
Bye!

PR #1317, opened 2015-03-02
Rack Visualisation view changes.

andrzej-jankowski:
List of changes:
- Service Catalog, Hostname and Management IP address are now display in
  Info Box on the left
- Service Catalog and hostname are now display on rack visualisation
  (instead of model)
- Minor CSS changes.

Depends on: https://github.com/allegro/ralph_assets/pull/605


mkurek:
:ok:

xliiv:
:ok:

PR #1320, opened 2015-03-03
Validation for IPWithHostnameField

zefciu:
The IPWithHostname field should always require IP Address, or else the
IPAddress object creation will fail.

### DIFF
--- a/src/ralph/ui/forms/addresses.py
+++ b/src/ralph/ui/forms/addresses.py
@@ -8,6 +8,7 @@ import ipaddr
 
 from bob.forms import AutocompleteWidget
 from django import forms
+from django.core.exceptions import ValidationError
 from django.template import Context
 from django.template.loader import get_template
 from django.utils.translation import ugettext_lazy as _
@@ -388,6 +389,10 @@ class IPWithHostField(forms.MultiValueField):
     widget = IPWithHostWidget()
 
     def compress(self, value):
+        if value:
+            hostname, ip_number = value
+            if not ip_number:
+                raise ValidationError(_('IP Address is required'))
         return value
 
 
--- a/src/ralph/ui/tests/unit/test_fields.py
+++ b/src/ralph/ui/tests/unit/test_fields.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+"""Tests for special fields."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.forms import ValidationError
+from django.test import TestCase
+
+from ralph.ui.forms.addresses import IPWithHostField
+
+
+class TestIPWithHostnameField(TestCase):
+    """Tests for ``IPWithHostField``"""
+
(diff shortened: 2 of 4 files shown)


src/ralph/ui/forms/addresses.py
```diff
@@ -388,6 +389,9 @@ def __init__(self, *args, **kwargs):
     widget = IPWithHostWidget()
 
     def compress(self, value):
+        hostname, ip_number = value
+        if not ip_number:
+            raise ValidationError(_('IP Address is required'))
```
xliiv: I know this works, but more reasonable place for this code is clean(..), I think.

mkurek:
I've checked out your branch and test it and this form no longer throw 500 when saving None IPAddress, BUT - despite the fact that you are throwing ValidationError user don't get any message that something goes wrong and ip field is not mark with ex. red color to suggest that there was a problem with this field. What's worse, it's looks like this form is silently ignoring ValidationError and after submitting it, it present values before changes. 

ip_host_valid

Besides that when i'm trying to save empty form i get an error: ValueError: need more than 0 values to unpack: models_device.py:810

src/ralph/ui/forms/addresses.py
```diff
@@ -388,6 +389,10 @@ def __init__(self, *args, **kwargs):
     widget = IPWithHostWidget()
 
     def compress(self, value):
+        if value:
+            hostname, ip_number = value
+            if not ip_number:
+                raise ValidationError(_('IP Address is required'))
```
kula1922: why not
if value and not value[1]:

src/ralph/ui/tests/unit/test_fields.py
```diff
@@ -0,0 +1,21 @@
...
+from ralph.ui.forms.addresses import IPWithHostField
+
+
+class TestIPWithHostnameField(TestCase):
+    """Tests for ``IPWithHostField``"""
+
+    def test_field_invalid_when_ip_number_missing(self):
+        field = IPWithHostField()
+        with self.assertRaises(ValidationError):
+            field.compress(['example.com', ''])
```
kula1922: I am not sure but it should work too

with self.assertRaises(ValidationError):
            IPWithHostField().compress(['example.com', ''])

src/ralph/ui/forms/addresses.py
vi4m: No docstring is a much important issue!

kula1922: There is not need know about veriables naming.  _('IP Address is required') give you everything what you need to know. 

Are you thinking that two nested "if" statement are little low shitdifficult to understand?

zefciu: I prefer not to address tuples with index numbers as they are meaningless. I believe this is not overnested here.l

zefciu: The ``compress`` is a standard method on this field. It is defined by django and need not to be documented by me.

vi4m:
LGTM

mkurek:
:ok:

PR #1326, opened 2015-03-17
Fix install python-graph

kula1922:
This fix is only for few days. Everything need back to old version when graph core will release as pypi package.


ar4s:
:ok:

vi4m:
LGTM

PR #1327, opened 2015-03-17
Various tweaks for visualization

zefciu:
* The racks get narrower on narrow monitors
* The hover/activation is uniformly done with a white border
* The borders in info table are corrected (removed extra border under
  last element.

### DIFF
--- a/src/ralph/ui/static/partials/data_center/rack.html
+++ b/src/ralph/ui/static/partials/data_center/rack.html
@@ -1,6 +1,6 @@
 <div class="rack rotate-{{ rack.orientation }} x-{{ rack.visualization_col }} y-{{ rack.visualization_row }}" ng-class="{'active': rack.active, 'move': is_move, 'allowed': can_drop, 'not_allowed': !can_drop}">
     <a href="#/rack/{{ rack.id }}" class="wrapper" ng-show="mode == 'preview'">
-        <div class="name"> 
+        <div class="name">
             {{ rack.name | clean }}
         </div>
         <div class="used">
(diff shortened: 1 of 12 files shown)


src/ralph/ui/static/partials/data_center/rack.html
```diff
@@ -1,6 +1,6 @@
-<div class="rack rotate-{{ rack.orientation }} x-{{ rack.visualization_col }} y-{{ rack.visualization_row }}" ng-class="{'move': is_move, 'allowed': can_drop, 'not_allowed': !can_drop}">
+<div class="rack rotate-{{ rack.orientation }} x-{{ rack.visualization_col }} y-{{ rack.visualization_row }}" ng-class="{'active': rack.active, 'move': is_move, 'allowed': can_drop, 'not_allowed': !can_drop}">
     <a href="#/rack/{{ rack.id }}" class="wrapper" ng-show="mode == 'preview'">
-        <div class="name">
+        <div class="name">
```
kula1922: why blank char?

src/ralph/ui/static/ui/js/app/data_center/controllers.js
```diff
@@ -28,6 +28,10 @@ angular
         };
 
         $scope.$on('edit_rack', function (event, rack) {
+            angular.forEach($scope.data_center.rack_set, function(rack) {
```
kula1922: It is ok but for future you can use shortly way

$scope.data_center.rack_set.forEach(function(){...})

src/ralph/ui/static/ui/js/app/data_center/controllers.js
```diff
@@ -28,6 +28,10 @@ angular
...
+                rack.active = false;
+            });
+            rack.active = true;
```
kula1922: there is no way to active will be false. If previous loop set it like a false, here you back to true.

src/ralph/ui/static/partials/data_center/rack.html
zefciu: idk my bff jill. Will correct

vi4m:
LGTM

ar4s:
:ok:

PR #1334, opened 2015-03-24
Deleted test assets' settings

ar4s:


mkurek:
Maybe you could delete pricing-test-settings at the same time (in the same PR)? Scrooge has it's own test settings (https://github.com/allegro/ralph_pricing/blob/develop/src/ralph_scrooge/settings-test-scrooge.py)

mkurek:
:ok:

xliiv:
:ok:

PR #1337, opened 2015-03-25
Juniper devices ignored in Cisco plugin scan

ojagodzinski:
Because of 'while True' in CiscoSSHHandler, on Juniper devices plugin sometimes sends command and waits for Cisco specific prompt forever. Checking if device is Juniper switch avoids this behavior.

Searching for 'cisco' in SNMP name can be another way to solve this problem but then we 'lose' switches with proper SSH configuration and some SNMP issues.

### DIFF
--- a/src/ralph/scan/plugins/ssh_cisco_catalyst.py
+++ b/src/ralph/scan/plugins/ssh_cisco_catalyst.py
@@ -99,6 +99,8 @@ def get_subswitches(switch_version, hostname, ip_address):
 
 
 def scan_address(ip_address, **kwargs):
+    if 'juniper' in (kwargs.get('snmp_name') or '').lower():
+        raise NoMatchError('Incompatible Juniper found.')
     if 'nx-os' in (kwargs.get('snmp_name', '') or '').lower():
         raise NoMatchError('Incompatible Nexus found.')
     if kwargs.get('http_family') not in ('Unspecified', 'Cisco'):


src/ralph/scan/plugins/ssh_cisco_catalyst.py
```diff
@@ -99,6 +99,8 @@ def get_subswitches(switch_version, hostname, ip_address):
 
 
 def scan_address(ip_address, **kwargs):
+    if 'juniper' in (kwargs.get('snmp_name', '') or '').lower():
```
kula1922: if 'juniper' in kwargs.get('snmp_name', '').lower():

mkurek: maybe `if 'juniper' in (kwargs.get('snmp_name') or '').lower():` - what if kwargs is dict with `snmp_name=None`? :)

zefciu:
:+1:

vi4m:
LGTM

src/ralph/scan/plugins/ssh_cisco_catalyst.py
vi4m: @ojagodzinski maybe put this extra safety check and we will merge you ASAP

ojagodzinski:
check changed as mkurek proposed

mkurek:
:ok:

vi4m:
so 1337 !

PR #1339, opened 2015-03-26
Fixed 'Management address' in form of Addresses tab.

xliiv:
Removed redundant button in the form + added submit name to the form so that
view can save the form (view checks for specific name in request to save it)

Depends on:
https://github.com/allegro/django-bob/pull/57


ar4s:
:ok:

quamilek:
:ok:

PR #1341, opened 2015-03-26
Remove workaround for installation python-graph-core

quamilek:
Now python-graph-core package is hosted on https://pypi.python.org/


ar4s:
:ok:

xliiv:
:ok:

PR #1345, opened 2015-03-30
Truncate too long surnames when synchronising users with ldap.

xliiv:
Before, ldap users with surname longer then 30 (Django's default) chars break
synching. Now such surnames are truncated to 30 chars.

### DIFF
--- a/src/ralph/account/management/commands/ldap_sync.py
+++ b/src/ralph/account/management/commands/ldap_sync.py
@@ -4,6 +4,7 @@ import sys
 import logging
 
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
 from ldap.controls import SimplePagedResultsControl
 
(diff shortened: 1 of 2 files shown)


src/ralph/account/management/commands/ldap_sync.py
```diff
@@ -128,6 +143,8 @@ def populate_users(self):
         """Load users from ldap and populate them. Returns number of users."""
         synced = 0
         for user_dn, ldap_dict in self._get_users():
-            self._create_or_update_user(user_dn, ldap_dict)
             synced += 1
+            print(synced, user_dn, ldap_dict['sAMAccountName'])
```
ar4s: ?

src/ralph/account/management/commands/ldap_sync.py
```diff
@@ -22,6 +23,20 @@
...
 
+def _truncate_surname(ldap_dict):
+    """
+    Truncate user's surname when it's longer then default django value, which is
+    30 chars.
+    """
+    if 'sn' in ldap_dict:
+        valid_sns = []
+        for surname in ldap_dict['sn']:
+            max_length = User._meta.get_field('last_name').max_length
```
zefciu: Why are you retrieving this in every iteration?

Maybe you could turn the whole loop into a comprehension?

src/ralph/account/management/commands/ldap_sync.py
zefciu: ¿

zefciu: ؟

quamilek: :hamburger:

zefciu:
:+1:

ar4s:
:ok:

PR #1346, opened 2015-03-31
Fixed highlighting elements on hover in visualization.

xliiv:
Removed side borders, left only top and bottom borders - highlight is
still visible but no flaw.

@ar4s

### DIFF
--- a/src/ralph/ui/static/ui/css/rack.less
+++ b/src/ralph/ui/static/ui/css/rack.less
@@ -220,11 +220,9 @@
                     border-bottom: 1px solid darken(@device-background, 10%);
                     border-top: 1px solid lighten(@device-background, 10%);
                     background: @device-background;
-                    &:hover {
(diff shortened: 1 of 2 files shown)


ar4s:
:ok:, but you must change in less file and compile or change manual both files :)

src/ralph/ui/static/ui/css/rack.less
```diff
@@ -221,10 +221,12 @@
                     border-top: 1px solid lighten(@device-background, 10%);
                     background: @device-background;
                     &:hover {
-                        .active-border();
+                        border-bottom: 1px solid white;
```
zefciu: It was my mistake that active-border is a mixin. It should be a constant with value '1px solid white'. Then it could be used on top or bottom or around whole object appropriately.

zefciu:
:+1:

PR #1350, opened 2015-04-07
Corrected the check

zefciu:
There is no need to convert this to set, as [None] == [None]. On the
other hand lists are not hashable, so this crashes with new plugin


ar4s:
:ok:

mkurek:
:ok:

PR #1351, opened 2015-04-09
Fixed unicode error and empty country by default

ar4s:
Prevent error like ``UnicodeEncodeError: 'ascii' codec can't encode
character``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
   - make install
   - pip install coveralls
   - pip install flake8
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -16,6 +16,7 @@ install:
   - make install
   - pip install coveralls
   - pip install flake8
+  - pip install django-auth-ldap==1.1.3
```
xliiv: :ok:

xliiv:
:ok:

mkurek:
:ok:

PR #1353, opened 2015-04-10
Removed lck.activitylog from Ralph

ar4s:
Basically nobody analyzes traffic in Ralph by lck.activitylog. This change
reduce number of DB queries about 12 per each request.


kula1922:
:ok:

vi4m:
GR8 !

PR #1354, opened 2015-04-13
Modifications to demo data command

ar4s:
* It will scan for entry points
* The default action is to load all fixtures. You can use -i to run
  interactively


vi4m:
LGTM

PR #1357, opened 2015-04-13
Moved visualization to new section (DC view) 

ar4s:
Previous version of vis gets racks from core and assets now the primary
source of information about data centers and racks is assets module.


zefciu:
:+1:

vi4m:
LGTM

PR #1375, opened 2015-04-15
Changed name for DeviceEnvironmentFactory and ServiceCatalogFactory

ar4s:
Now factory return object with human readable name.


xliiv:
:ok:

szok:
:ok:

PR #1383, opened 2015-04-16
Demo data: VIP, Database, Virtual and Tenant

mkurek:
* added VIP, Database, Virtual and Tenant to demo data
* minor tweaks in other demo data classes


ar4s:
:ok:

szok:
:ok:

PR #1391, opened 2015-04-29
Added docker-ralph references when installation from sources

vi4m:
Documentation should point you to the ralph-docker in case of installation from sources.


vi4m:
Will close it now, and move some other documentation toward this branch.

PR #1401, opened 2015-05-07
LDAP implementation specifics

damian1976:
Just a small update reflecting possible difference between various LDAP server implementations. In case of OpenDJ in case of objects filtering there's objectclass name change: user -> person.

### DIFF
--- a/src/ralph/settings.py
+++ b/src/ralph/settings.py
@@ -312,6 +312,7 @@ FISHEYE_URL = ""
 FISHEYE_PROJECT_NAME = ""
 VMWARE_USER = None
 VMWARE_PASSWORD = None
+LDAP_SERVER_OBJECT_USER_CLASS = "user" # possible values: "user, person"
 
 ISSUETRACKERS = {
     'default': {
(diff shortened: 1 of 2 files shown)


mkurek:
@damian1976 squash it into one commit and we'll merge it

src/ralph/settings.py
```diff
@@ -312,6 +312,7 @@
 FISHEYE_PROJECT_NAME = ""
 VMWARE_USER = None
 VMWARE_PASSWORD = None
+LDAP_SERVER_OBJECT_USER_CLASS = "person" # possible values: "person, user"
```
mkurek: you shouldn't change default value which was 'user' - please set it to 'user' and overwrite it in your own settings if you need 'person'

mkurek:
:ok:

vi4m:
Wait, we should move it away from develop -> ng

vi4m:
It's primarly usable for @damian1976 , so should be moved to Ng

mkurek:
@szok remember about this when you'll move ldap stuff to ng

mkurek:
This will be added to ng: https://github.com/allegro/ralph/pull/1501

PR #1406, opened 2015-05-11
Removed exception when device have different processors.

ghost:
Removed exception rasing when device has processors with different types.


mkurek:
:ok:

ar4s:
:ok:

PR #1413, opened 2015-05-13
Added new commands ({dev, test}_ralph)

ar4s:
This is some shortcut. `ralph_dev` - run ralph with develop settings (`ralph.settings.dev`).
Now you can type for example `ralph_dev runserver_plus` instead of `ralph
runserver_plus --settings=ralph.settings.dev`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,8 @@ setup(
     entry_points={
         'console_scripts': [
             'ralph = ralph.__main__:main',
+            'dev_ralph = ralph.__main__:dev',
+            'test_ralph = ralph.__main__:test',
         ],
     },
     classifiers=[
--- a/src/ralph/__main__.py
+++ b/src/ralph/__main__.py
@@ -3,13 +3,21 @@ import os
 import sys
 
 
-def main():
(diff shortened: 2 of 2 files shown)


mkurek:
"discussable" :smirk:

We're going to use vagrant for development so maybe we could add to some `.bashrc` file etc some environment variables settings:

```
export DJANGO_SETTINGS_MODULE="ralph.settings.dev"
```

ar4s:
ok, but.. what if you want to run server in dev mode outside vagrant env (e.g. staging or prod :joy: )? IMHO it's small change and we know explicit (by `_dev` postfix) which settings is set.

mkurek:
I know - but this will make it inconvenient to autocomplete ralph command (ex. `ral<tab>`) if there will be more than one command with `ralph` prefix (normally, if there is only one command autocompletion fill it and insert space at the end). So maybe `dev_ralph` instead of `ralph_dev` (this won't confuse not-dev users as well - probably normal users will not notice this command, but will notice `ralph_dev` command)?

mkurek:
:ok:

PR #1414, opened 2015-05-13
POC Extends permission by access to each field in model

ar4s:
Take a look at this piece of code :boom:


vi4m:
Awesome as f4ck :-) Have to try this out!

vi4m:
I wonder what is the total number of fields for all models, though.

szok:
:ok:

ghost:
:ok:

PR #1416, opened 2015-05-14
Add blacklist to permissions model field.

szok:


src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,46 @@ def __new__(cls, name, bases, attrs):
...
-    # TODO
-    @property
-    def allowed_fields(self, user=None, group=None):
-        pass
 
     class Meta:
         abstract = True
+
+    class Permissions:
+        blacklist = set(['created', 'modified'])
```
mkurek: created and modified should be blacklisted in theirs model

src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,46 @@ def __new__(cls, name, bases, attrs):
...
+            blacklist = new_class.Permissions.blacklist
             name = field.name
-            new_class._meta.permissions.append((
-                'can_edit_{}_field'.format(name),
-                _('Can edit {} field').format(name)
-            ))
+            if not field.primary_key and name not in blacklist:
+                new_class._meta.permissions.append((
+                    'can_edit_{}_{}_field'.format(
+                        new_class.__name__.lower(),
```
ar4s: is not necessary.. `Permission` models contains `content_type` (https://github.com/django/django/blob/1.8.1/django/contrib/auth/models.py#L65)

mkurek: use `new_class._meta.model_name` and add `new_class._meta.app_label` here as well (in case if in two packages there will be models with the same name and the same field.

ar4s:
:ok:

src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,50 @@ def __new__(cls, name, bases, attrs):
...
+            blacklist = new_class.Permissions.blacklist
             name = field.name
-            new_class._meta.permissions.append((
-                'can_edit_{}_field'.format(name),
-                _('Can edit {} field').format(name)
-            ))
+            if not field.primary_key and name not in blacklist:
+                new_class._meta.permissions.append((
+                    'change_{}_{}_field'.format(
+                        new_class.__name__.lower(),
```
mkurek: `new_class._meta.model_name`

src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,50 @@ def __new__(cls, name, bases, attrs):
...
-    # TODO
-    def have_access_to_field(self, field, user=None, group=None):
-        pass
+    """Perm by field mixin."""
+
+    def get_perm_key(self, action, field_name):
+        return '{}.{}_{}_{}_field'.format(
+            self._meta.app_label,
+            action,
+            self.__class__.__name__.lower(),
```
mkurek: ditto

src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,50 @@ def __new__(cls, name, bases, attrs):
...
 
 
 class PermByFieldMixin(with_metaclass(PermissionByFieldBase, models.Model)):
 
-    # TODO
-    def have_access_to_field(self, field, user=None, group=None):
-        pass
+    """Perm by field mixin."""
+
+    def get_perm_key(self, action, field_name):
```
mkurek: extract it from class (just regular function) and use it in metaclass and here

src/ralph/lib/permissions/models.py
```diff
@@ -19,24 +19,50 @@ def __new__(cls, name, bases, attrs):
...
+        return '{}.{}_{}_{}_field'.format(
+            self._meta.app_label,
+            action,
+            self.__class__.__name__.lower(),
+            field_name
+        )
+
+    def have_access_to_field(self, field_name, user, action='change'):
```
mkurek: has_access_to_field

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,62 @@
...
 
-    # TODO
-    def have_access_to_field(self, field, user=None, group=None):
-        pass
+    """Perm by field mixin."""
+
+    def has_access_to_field(self, field_name, user, action='change'):
+        perm_key = get_perm_key(
+            action,
+            self.__class__.__name__.lower(),
```
mkurek: self._meta.model_name

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,62 @@
...
+        return user.has_perm(
+            "{}.{}".format(self._meta.app_label, perm_key)
+        )
+
+    def allowed_fields(self, user, action='change'):
+        result = []
+        for field in self._meta.fields:
+            perm_key = get_perm_key(
+                action,
+                self.__class__.__name__.lower(),
```
mkurek: ditto

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,62 @@
 from django.utils.translation import ugettext_lazy as _
 
 
+def get_perm_key(action, class_name, field_name):
+    return '{}_{}_{}_field'.format(action, class_name, field_name)
```
ghost, line 34: Maybe add example string in comment.

ghost:
I don't see many docstring. ;)

ar4s:
I see dead people...

szok:
?

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
 from django.utils.translation import ugettext_lazy as _
 
 
+def get_perm_key(action, class_name, field_name):
+    """
+    Generated django permission code name.
```
mkurek: Generate

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
+
+    :Example:
+
+        >> perm = get_perm_key('change', 'baseobject', 'remarks')
```
mkurek, line 21:
```
>>> get_perm_key('change', 'baseobject', 'remarks')
change_baseobject_remarks_field
```

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
+    :rtype: str
+    """
+    return '{}_{}_{}_field'.format(action, class_name, field_name)
+
+
 class PermissionByFieldBase(ModelBase):
-    """Extends permissions by edit per field of model."""
+
+    """
+    Metaclass added django permissions based on all fields in the model.
```
mkurek: adding

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
+
+    :Example:
+
+        class Test(with_metaclass(PermissionByFieldBase, models.Model)):
+
+            ...
+
+            class Permissions:
+                # Fields to exclude generated permissions
+                blacklist = set(['service_env'])
```
mkurek: don't incorporate servie_env here - rename it to `sample_field` or sth like that

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
         )
+        class_name = new_class._meta.model_name
         for field in new_class._meta.fields:
+            blacklist = new_class.Permissions.blacklist
             name = field.name
-            new_class._meta.permissions.append((
-                'can_edit_{}_field'.format(name),
-                _('Can edit {} field').format(name)
-            ))
+            if not field.primary_key and name not in blacklist:
```
mkurek, line 61: make sure this is working with multi-column primary key

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
+        Checks the user has the permission to the field
 
-    # TODO
-    @property
-    def allowed_fields(self, user=None, group=None):
-        pass
+        :Example:
+
+            >> user = User.objects.get(username='root')
+            >> print(model.has_access_to_field('remarks', user, action='change'))
```
mkurek: ditto - don't use print here - just show returned value:

```
>>> func()
True
```

src/ralph/lib/permissions/models.py
```diff
@@ -12,31 +12,133 @@
...
+            >> True
+
+        :param field_name: django model field name
+        :param user: django User object
+        :param action: permission action (change/view)
+
+        :type field_name: str
```
mkurek, line 89: to discuss if we want first definition of all fields and then all types or mixed (field, type, field, type, ...). @ar4s @vi4m @xliiv what do you think?

src/ralph/lib/tests/test_models.py
```diff
@@ -0,0 +1,106 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.test import TestCase
+
+from ralph.assets.models.choices import ObjectModelType
```
mkurek: dont use ralph models here - it should be "abstract"

mkurek: we've done sth similar in scrooge:
https://github.com/allegro/ralph_pricing/blob/develop/src/ralph_scrooge/tests/models.py
We've created models only for tests (notice that it was in Django 1.4 so there might be differences)

src/ralph/lib/permissions/models.py
szok, line 61: Yes is working properly:
In [3]: [i.primary_key for i in Test._meta.fields]
Out[3]: [True, True]

src/ralph/lib/permissions/models.py
ar4s, line 89: +1 for mixed

ar4s:
:ok:

mkurek:
:ok:

Issue #1411, opened 2015-05-13
Permissions per field


ar4s:
General resarch about permission per model's field. It is possible to integrate with builtin django's permissions?

ar4s:
Some POC https://github.com/allegro/ralph/pull/1414

mkurek:
Correct me if i'm wrong but afaik we're looking for complex solution which will allow/disallow group of users access to:
1. whole model (ex. some backoffice users shouldn't change datacenter objects)
2. model instances (similar to regions from ralph 2.0 - group of users has access to object based on some rules defined for example on model)
3. instance fields (ex. accountants shouldn't change admins fields and vice versa)

If we're making complex solution we should consider also readonly versions in types 2 and 3 (user could see instance or some fields but couldn't change them).

Permissions should be highly customisable (defined for groups of users in admin panel). 

This solution should be applied to every way of action user could make: GUI (both admin and custom views in future), API, importer etc. One way of doing this is processing every change through model and validate there user permissions (and for example don't use django orm bulk operations which skip some model methods).

ar4s:
I agree with you. #1414 it's only first step to complex solution.

PR #1417, opened 2015-05-14
WIP Form respect permission by field

ar4s:
Only for preview
connected to #1411


src/ralph/lib/permissions/widgets.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.forms import Field, HiddenInput
+
+
+class ReadOnlyWidget(HiddenInput):
+    is_hidden = False
+
+    def render(self, name, value, attrs=None):
+        html = super(ReadOnlyWidget, self).render(name, value, attrs)
+        wrapped_value = '<div class="empty">{}</div>'.format(value or 'empty')
```
mkurek: div or span, that is the question :broken_heart:

PR #1429, opened 2015-05-18
Docker and docker-compose configuration

mkurek:
- added Dockerfile for ralph
- added docker-compose (previously fig) configuration with mysql and nginx

connected to #1432

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,32 @@
+db:
+  image: mysql:5.6
+  environment:
+    MYSQL_ALLOW_EMPTY_PASSWORD: yes
+    MYSQL_DATABASE: ralph_ng
+    MYSQL_ROOT_PASSWORD:
+    MYSQL_USER: ralph_ng
+    MYSQL_PASSWORD: ralph_ng
+
+web:
+  build: .
+  restart: always
+  environment:
+    - DJANGO_SETTINGS_MODULE=ralph.settings.settings_local
+  ports:
+    - "8000:8000"
+  volumes:
+    - docker/settings_local.py:/opt/ralph/src/ralph/settings/settings_local.py
+  links:
+    - db:db
+
+nginx:
+  image: nginx
+  restart: always
+  ports:
+    - "80:80"
+  links:
+    - web:web
+  volumes:
+    - docker/ralph.conf:/etc/nginx/conf.d/default.conf
+  volumes_from:
+    - web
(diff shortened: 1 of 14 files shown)


docker/ralph.conf
```diff
@@ -0,0 +1,18 @@
+server {
+    listen 80;
+    server_name ralph;
+
+    location /docs {
+        alias /opt/ralph/docs/_build/html;
+    }
+
+#    location /static {
```
vi4m, line 9: Why is this commented out?

docker/upgrade.sh
```diff
@@ -0,0 +1,8 @@
+##!/bin/bash
```
vi4m, line 1: to many ##

docker/upgrade.sh
```diff
@@ -0,0 +1,8 @@
...
+set -e
+
+cd $RALPH_DIR
```
vi4m, line 4: what will happen if RALPH_DIR is not set?

vi4m:
Docker compose is not nearly production ready. We should point it out clearly in documentation.

docker/init.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+# wait for db init
+sleep 15
+
+ralph migrate
```
vi4m, line 7: always provide precise path for binary

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+
+web:
+  build: .
+  restart: always
+  environment:
+    - DJANGO_SETTINGS_MODULE=ralph.settings.settings_local
+  ports:
+    - "8000:8000"
+  volumes:
+    - docker/settings_local.py:/opt/ralph/src/ralph/settings/settings_local.py
```
vi4m, line 18: docker configuration shoule be the same as in use in production. Why we decided to drop ~/.ralph/settings ?

docker/upgrade.sh
mkurek, line 4: RALPH_DIR will always be set (it's in Dockerfile).

docker-compose.yml
mkurek, line 18: I've added a comment that it's only temporary until ralph will handle local settings in other (better) way.

mkurek:
@vi4m I've added info to documentation that it's not for production use right now (notice that docker-compose is one reason and second is using develop settings and packages - it's only for testing and fast deployment of test instances).

vi4m:
Awesome.

szok:
:ok:

ar4s:
:ok:

mkurek:
wait with merging it for now

PR #1435, opened 2015-05-19
List and add form back office models.

szok:
Added:
- Back Office list configuration
- Back Office add form configuration

Related to #1436

### DIFF
--- a/src/ralph/back_office/admin.py
+++ b/src/ralph/back_office/admin.py
@@ -2,6 +2,7 @@
 import reversion
 
 from django.contrib import admin
+from django.utils.translation import ugettext_lazy as _
 
 from ralph.back_office.models import (
     BackOfficeAsset,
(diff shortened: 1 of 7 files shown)


vi4m:
why `search_fields` covers only `hostname` ?

src/ralph/back_office/admin.py
```diff
@@ -11,7 +12,41 @@
...
 class BackOfficeAssetAdmin(reversion.VersionAdmin):
-    pass
+
+    """Back Office Asset admin class."""
+
+    list_display = [
+        'status', 'barcode', 'purchase_order', 'model', 'user', 'warehouse',
+        'sn', 'hostname', 'invoice_date', 'invoice_no'
+    ]
+    search_fields = ['hostname', '']
```
mkurek: empty?

src/ralph/back_office/admin.py
```diff
@@ -11,7 +12,41 @@
...
+    list_filter = ['owner', 'user']
```
mkurek: this will probably kill this view :) maybe some other fields which have lower cardinality?

src/ralph/back_office/admin.py
```diff
@@ -11,7 +12,41 @@
...
+    list_select_related = ['model', 'user', 'warehouse']
```
mkurek: if user why not owner too?

src/ralph/back_office/admin.py
szok: My mistake

src/ralph/back_office/admin.py
szok: Yes, we have to discuss

ar4s:
:ok:

ghost:
:ok:

PR #1441, opened 2015-05-20
License and supports in django admin

szok:
Added:
- License admin fieldsets
- Support admin fieldsets
- Fix Calendar icon
- AssetModel and service env to django admin

Changed:
- New mechanism to generate permissions

Connected to #1445


src/ralph/licences/models.py
```diff
@@ -36,8 +37,11 @@ class LicenceType(NamedMixin, models.Model):
     def create_from_string(cls, string_name, *args, **kwargs):
         return cls(name=string_name)
 
+    class Permissions:
+        blacklist = set()
```
mkurek: do we really need to add this to every class which inherits from PermByFieldMixin?

szok: Ok, I will add exception as there is no defined class.

ar4s: @mkurek why not?

mkurek: do you add empty `Meta` or 

```
class Meta:
    unique_together = []
```

to every model? Empty `blacklist` could be default option - only when you want to blacklist sth you could add `Permissions` class to model definition.

ar4s: Now I see :smile: I thought it concern `lib/permissions/models.py` 
Maybe that way https://github.com/django/django/blob/master/django/db/models/base.py#L83-L84

src/ralph/lib/permissions/models.py
```diff
@@ -56,7 +56,10 @@ def __new__(cls, name, bases, attrs):
         )
         class_name = new_class._meta.model_name
         for field in new_class._meta.fields:
-            blacklist = new_class.Permissions.blacklist
+            try:
```
mkurek, line 70: you could move this before for loop to check for blacklist only once

mkurek:
:ok:

ar4s:
:ok:

Issue #1410, opened 2015-05-13
Exporter from old ralph


ar4s:
## Story

Exporter must save data from old ralph and simultaneously convert data schema to new models design.
## Criteria
- convert old schema models to new models concept
- config as dict
- bulk export and import

``` python
{
    models.BackOfficeAsset: {list, dict} 
}
```

ar4s:
Export demo data to CSV from old ralph we will use it in ralph.mkurek.com for testing purposes.

PR #1456, opened 2015-05-28
NG exporter

ghost:
Added exporter to RalphNG.

connected to #1410

### DIFF
--- a/src/ralph/export_to_ng/management/commands/export_model.py
+++ b/src/ralph/export_to_ng/management/commands/export_model.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os
+import tempfile
+from optparse import make_option
+from zipfile import ZipFile
+
+from django.core.management.base import BaseCommand
+from django.db.models import get_models
+from import_export import resources
+
+import ralph_assets
+from ralph.discovery import models_device
+from ralph.export_to_ng import resources as ralph_resources
+
+ZIP_EXT = '.zip'
+NAMESPACE_SIZE = 10
+SIMPLE_MODELS = [
+    'AssetLastHostname',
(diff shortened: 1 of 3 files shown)


mkurek:
what's the difference between `export_model` and `export_models`?

ghost:
`export_models` will export all models required by NG, now it's just ordinary copy-paste with small change.

src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -0,0 +1,124 @@
...
+class Command(BaseCommand):
+    help = os.linesep.join([
+        'Export data in Ralph-NG format.',
+        ALLOWED_MODELS_MSG,
+    ])
+
+    # TODO:: zipped as arg
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--dst-dir',
```
mkurek: afaik if param has only `--` version it is not required - shouldn't it be required?

src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def handle(self, *args, **options):
+        models = args
+        if not models:
+            models = SIMPLE_MODELS + DEPENDENT_MODELS + MANY_TO_MANY
+        for model in models:
+            self.stdout.write('Processing model {}\n'.format(model))
+
+            model_resource = get_resource(model)
+            queryset = model_resource._meta.model.objects.all()[:1]
```
mkurek: is `[:1]` only for tests?

src/ralph/export_to_ng/management/commands/export_model.py
ghost: In current version should be I think. I'll fix it, thanks.

src/ralph/export_to_ng/management/commands/export_model.py
ghost: True

ar4s:
:ok:

mkurek:
:ok:

Issue #1426, opened 2015-05-18
Add data center visualisation


mkurek:
## Story

Data center visualisation in ng (based on ralph visualisation).
## Subtasks
- copy visualisation one-to-one from old ralph
- adjust it to new models and API

vi4m:
We can use sub-issues and reference it from here :)

mkurek:
Yes, but it'll be a little bit overhead here - everything has to be done in one step either.

Issue #1451, opened 2015-05-26
Gulp or another task runner and bower


ar4s:
## Criteria
- compile SASS
- bower as package manager (foundation, jquery, angular, datetimepicker)
- update vagrant provisioning

PR #1458, opened 2015-05-28
Add Data Center Visualization

szok:
Connected to #1426


mkurek:
what about https://github.com/allegro/ralph/issues/1451 ? Are we going to merge it as it is (just a copy from old ralph) and then adjust everything to support gulp etc or could we handle both topics in one PR? Or maybe should we first introduce gulp etc into NG and then move visualisation to NG? @ar4s @szok @xliiv ?

vi4m:
Rule of thumb: Keep issue/stories _thin and lean_, but make architecture great.

Integrating gulp as we've seen is not the smallest task, so maybe start with something actually working and move fast in next interations?

src/ralph/assets/tests/factories.py
```diff
@@ -0,0 +1,46 @@
...
+    AssetModel,
+    Environment,
+    Service,
+    ServiceEnvironment
+)
+
+
+class AssetModelFactory(DjangoModelFactory):
+
+    name = factory.Sequence(lambda n: 'Environment #{}'.format(n))
```
mkurek: Environment?

src/ralph/assets/tests/factories.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
mkurek, line 1: BTW to discuss if we want to reuse these factories in demo data generation process. If yes, then we should probably give them meaningful names etc. @ar4s @xliiv ?

src/ralph/data_center/models.py
```diff
@@ -160,6 +160,7 @@ class Meta:
         verbose_name_plural = _('VIPs')
 
 
+# TODO!!!
```
mkurek: TODO what?

src/ralph/data_center/tests/factories.py
```diff
@@ -0,0 +1,62 @@
...
+    DataCenterAsset,
+    Rack,
+    RackAccessory,
+    ServerRoom,
+)
+
+
+class DataCenterFactory(DjangoModelFactory):
+
+    name = factory.Sequence(lambda n: 'Data Center #{}'.format(n))
```
ar4s: use cycle to generate name https://github.com/allegro/ralph_assets/blob/94effeee0195b76ed25f1b1abfeed83765ebfb18/src/ralph_assets/tests/utils/assets.py 
https://github.com/allegro/ralph_assets/blob/94effeee0195b76ed25f1b1abfeed83765ebfb18/src/ralph_assets/tests/utils/assets.py#L62

src/ralph/dc_view/urls/api.py
```diff
@@ -0,0 +1,29 @@
...
+
+from django.conf.urls import patterns, url
+
+from ralph.dc_view.views.api import (
+    DCAssetsView,
+    DCRacksAPIView
+)
+
+
+urlpatterns = patterns(
```
mkurek: See
https://docs.djangoproject.com/en/1.8/topics/http/urls/#syntax-of-the-urlpatterns-variable and https://docs.djangoproject.com/en/1.8/releases/1.8/#django-conf-urls-patterns - from Django 1.8 `patterns` is deprecated.

src/ralph/dc_view/urls/ui.py
```diff
@@ -0,0 +1,19 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.conf.urls import patterns, url
+
+from ralph.dc_view.views.ui import DataCenterView
+
+
+urlpatterns = patterns(
```
mkurek: ditto

src/ralph/assets/tests/factories.py
```diff
@@ -0,0 +1,54 @@
...
+
+    name = factory.Iterator(['3310', 'XD300S', 'Hero 3', 'Computer set'])
+
+    class Meta:
+        model = AssetModel
+
+
+class DataCenterAssetModelFactory(DjangoModelFactory):
+
+    name = factory.Iterator(['DL360', 'DL380p', 'DL380', 'ML10', 'ML10 v21'])
```
mkurek, line 30: what will happen if I'll create (in the same test) 6 data center assets (which should reference `DataCenterAssetModelFactory` using `SubFactory`)?

src/ralph/assets/tests/factories.py
```diff
@@ -0,0 +1,57 @@
...
+        model = AssetModel
+        django_get_or_create = ['name']
+
+
+class EnvironmentFactory(DjangoModelFactory):
+
+    name = factory.Iterator(['prod', 'dev', 'test'])
+
+    class Meta:
+        model = Environment
```
mkurek, line 42: imo should have `django_get_or_create` as well

szok, line 42: Done

mkurek:
:ok:

src/ralph/data_center/models/physical.py
```diff
@@ -129,17 +219,38 @@ class DataCenterAsset(Asset):
...
-        symmetrical=False,
-    )
+    def get_related_assets(self):
+        """Returns the children of a blade chassis"""
+        orientations = [Orientation.front, Orientation.back]
+        assets_by_orientation = []
+        for orientation in orientations:
+            assets_by_orientation.append(list(
+                DataCenterAsset.objects.select_related('model').filter(
+                    position=self.position,
```
ar4s: try replace `position=self.position, rack=self.rack` by `parent=self`

ar4s:
:ok:

PR #1463, opened 2015-05-29
Changed device's widgets in admin.

ghost:
Old widgets was unefficient because they loaded a lot of items.
New widgets autocomplete values so all items are not loaded.


ar4s:
:ok:

szok:
:ok:

Issue #1430, opened 2015-05-18
Moving to the mkdocs engine


vi4m:
This is general discussion issue about making a switch to the mkdocs architecture. 

Benefits:
- Markdown!
- Simple build-free instant preview
- Read the docs already support it so it won't be much an effort to move out
- RST is so verbose and error-prone, just one notable example:

```
Header
------------- oh shit.

# Header no shit
```
- One standard for github and docs
- We're rewriting documentation anyway
- It's cool :)

PR #1470, opened 2015-06-02
Added mkdocs support

ar4s:
Connected to #1430


requirements/docs.txt
```diff
@@ -0,0 +1 @@
+mkdocs==0.12.2
```
mkurek, line 1: could you require (include) it in test.txt/dev.txt?

ar4s, line 1: I added `-r docs.txt` in `test.txt`

ghost:
:ok:

mkurek:
:ok:

PR #1474, opened 2015-06-03
WIP Ralph exporter

szok:


src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -0,0 +1,125 @@
...
+        'Possible models to export:',
+        POSSIBLE_MODELS,
+    ])
+
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--model_name',
+            help=(
+                'The model name up to export '
+                '(type "ralph export_to_ng -h" for possible models)'
```
vi4m, line 103: It is just ralph export, not export_to_ng, we've built generic mechanism

mkurek, line 103: on the one hand yes, but on the other side export format is perfectly tailored to ng (export format does not suit current ralph models).

vi4m, line 103: You want to emphasize the fact it's ng driven?

PR #1487, opened 2015-06-15
Add access by user login to sitetrees menu

szok:


ghost:
:ok:

src/ralph/admin/sitetrees.py
```diff
@@ -39,63 +39,75 @@
                 'data_center.change_virtualserver'
             ],
             perms_mode_all=False,
+            access_loggedin=True,
```
mkurek: what do you think about creating some `ralph_item` which will be partial to item with `access_loggedin=True` set?

szok: good idea

src/ralph/admin/sitetrees.py
```diff
@@ -21,9 +21,15 @@
 
 activate(settings.LANGUAGE_CODE)
 
+
+def ralph_item(*args, **kwargs):
+    kwargs['access_loggedin'] = True
```
mkurek: maybe use `setdefault` here (is `access_loggedin` is defined, we don't want to overwrite it)

ar4s:
:ok:

PR #1491, opened 2015-06-16
Add import from directory and zip file.

szok:


src/ralph/data_importer/management/commands/importer.py
```diff
@@ -79,8 +83,49 @@ def add_arguments(self, parser):
...
+        )
+
+    def from_zip(self, options):
+        with open(options.get('zipfile'), 'rb') as f:
+            out_path = tempfile.mkdtemp()
+            z = zipfile.ZipFile(f)
+            for name in z.namelist():
+                z.extract(name, out_path)
+            options['dir'] = out_path
+            options['skipid'] = True
```
mkurek: why?

szok: Because, the default is not skipid, and when importing from a ZIP we want create old id to ImportedObject

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -79,8 +83,49 @@ def add_arguments(self, parser):
...
+        file_list = []
+        for path in glob.glob(os.path.join(options.get('dir'), '*.csv')):
+            base_name = os.path.basename(path)
+            file_name = os.path.splitext(base_name)[0].split('_')
+            file_list.append({
+                'model': file_name[1],
+                'path': path,
+                'sort': file_name[0]
+            })
+        file_list = sorted(file_list, key=lambda x: x['sort'])
```
mkurek, line 108: (related with https://github.com/allegro/ralph/pull/1456, but it's already merged and i think about the problem post factum)

https://github.com/allegro/ralph/pull/1456/files#diff-78029b48156ee44655b1cf8e84dce40fR127 - add filling file number with zeros - otherwise this sorting will not work. Example:

```
>>> sorted(['1_sth.csv', '2_sth.csv', '10_sth.csv'])
['10_sth.csv', '1_sth.csv', '2_sth.csv']
```

File names should looks like that:

```
>>> sorted(['0001_sth.csv', '0002_sth.csv', '0010_sth.csv'])
['0001_sth.csv', '0002_sth.csv', '0010_sth.csv']
```

szok, line 108:
```
>>> sort_list = [{'sort': 10}, {'sort': 1}, {'sort': 2}]
>>> sorted(sort_list, key=lambda x: x['sort'])
[{'sort': 1}, {'sort': 2}, {'sort': 10}]
>>> 
```

mkurek, line 108: ah now i see. but you'll have actually strings - not ints.

```
>>> sort_list = [{'sort': '10'}, {'sort': '1'}, {'sort': '2'}]
>>> sorted(sort_list, key=lambda x: x['sort'])
[{'sort': '1'}, {'sort': '10'}, {'sort': '2'}]
```

src/ralph/data_importer/management/commands/importer.py
mkurek: When importing from old ralph - yes. But what when someone want to import everything without skipping? IMO this flag should stay as it is and when importing from NG we should use it explicitly (`--skipid`)

szok: Ok, fact

ghost:
:ok:

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -111,21 +155,21 @@ def handle(self, *args, **options):
...
+            if not options.get('noinput'):
+                self.stdout.write('Done')
+
+    def handle(self, *args, **options):
+        if not any([
+            options.get('dir'),
+            options.get('data_file'),
+            options.get('zipfile')
+        ]):
+            raise CommandError(
```
mkurek: since we only want to get one option, maybe there should be only one param with value (and 3 available choices), ex:

```
parser.add_argument(
    '-t', '--type',
    choices=['file', 'dir', 'zip'],
    required=True
)
```

?

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -79,8 +89,52 @@ def add_arguments(self, parser):
             default='utf-8',
             help="Output encoding",
         )
+        parser.add_argument(
```
mkurek: dir and zipfile are no longer necessary. Use single, positional param for file source path.

mkurek:
:ok:

PR #1497, opened 2015-06-18
Change 'f5app' strings to '.f5.' for discovering f5 devices.

jjagodzinski:


jjagodzinski:
F5 has changed kernel name syntax. Earlier there was '.f5app' in kernel name, right now there is only .f5. Change is needed to discover new f5.

mkurek:
Could we check for both (`.f5app` and `.f5.`) to keep backward compatibility with F5s with old kernel?

jjagodzinski:
Ok, changed.

mkurek:
ok - so please squash it into one commit and it's :ok: for me

andrzej-jankowski:
:ok:

PR #1507, opened 2015-06-23
NG-exporter changes for better data porting to RalphNG.

ghost:
Mainly resources update.

### DIFF
--- a/src/ralph/export_to_ng/management/commands/export_model.py
+++ b/src/ralph/export_to_ng/management/commands/export_model.py
@@ -27,7 +27,7 @@ SIMPLE_MODELS = [
     'Service',
     'Manufacturer',
     'Category',
-    'ComponentModel',
+    # mainly trash data: 'ComponentModel',
     'Warehouse',
     'DataCenter',
     'Accessory',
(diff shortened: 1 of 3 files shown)


src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -111,7 +111,7 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         models = args
         if not models:
-            models = SIMPLE_MODELS + DEPENDENT_MODELS + MANY_TO_MANY
+            models = SIMPLE_MODELS + DEPENDENT_MODELS
```
mkurek: what about many-to-many?

src/ralph/export_to_ng/resources.py
```diff
@@ -64,14 +74,20 @@ def dehydrate_rack(self, asset):
...
+            rack = ''
         return rack
 
     def dehydrate_service_env(self, asset):
-        device_environment = (
-            asset.device_environment.name if asset.device_environment else ''
+        service = getattr(
+            asset.service,
+            'name',
+            settings.NG_EXPORTER['default_service_name'],
```
mkurek: service could be null now so we shouldn't assign default values here

src/ralph/export_to_ng/resources.py
```diff
@@ -64,14 +74,20 @@ def dehydrate_rack(self, asset):
...
+        )
+        device_environment = getattr(
+            asset.device_environment,
+            'name',
+            settings.NG_EXPORTER['default_environment_name'],
```
mkurek: ditto

mkurek:
:ok:

szok:
:ok:

PR #1520, opened 2015-06-23
Fixed frozen request when content-length equals 0

ar4s:
Prevent freezing request in `get_http_info`.

### DIFF
--- a/src/ralph/discovery/http.py
+++ b/src/ralph/discovery/http.py
@@ -20,29 +20,45 @@ FAMILIES = {
     'IBM_HTTP_Server': 'IBM',
     '': 'Unspecified',
 }
-REQUEST_TIMEOUT = 5
+REQUEST_KWARGS = {
+    'timeout': 5,
+    'verify': False,
+    'allow_redirects': False,
+}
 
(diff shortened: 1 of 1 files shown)


ghost:
:ok:

src/ralph/discovery/http.py
```diff
@@ -38,11 +54,12 @@ def get_http_info(ip):
         logger.error("Exception: {} for url: {}".format(e, url))
         url = "https://{}".format(':'.join((ip, '8006')))  # for Proxmox
         try:
-            response = requests.get(url, timeout=REQUEST_TIMEOUT, verify=False)
+            response = http_get_method(session, url)
             response.raise_for_status()
             headers, content = response.headers, response.content
         except requests.exceptions.RequestException as e:
             logger.error("Exception: {} for url: {}".format(e, url))
+    session.close()
```
mkurek: shouldn't this be in some finally block?

ar4s: maybe better than finally block will be with statement around whole `try..execpt` blocks?

mkurek:
:ok:

PR #1521, opened 2015-06-24
Fixed error during scans Dell

ar4s:
In some cases scaned disks have empty serial number, this changes prevent
error like `'NoneType' object has no attribute 'strip'`.


mkurek:
what about unit tests? :)

ar4s:
@mkurek I changed samples data - I think this is enough to test my case :)

mkurek:
:ok:

ghost:
:ok:

Issue #1517, opened 2015-06-23
Sample tabs


mkurek:
Sample tabs (with static content) for:
- [x] DC Asset components
- [x] BO Asset components
- [x] DC Asset Software
- [x] BO Asset Software
- [x] DC Asset Security Info

PR #1522, opened 2015-06-24
Sample tabs

szok:
Connected to #1517

### DIFF
--- a/src/ralph/back_office/templates/backofficeasset/components.html
+++ b/src/ralph/back_office/templates/backofficeasset/components.html
@@ -1,0 +1,32 @@
+{% extends BASE_TEMPLATE %}
+{% load i18n %}
+{% block view_content %}
+<div class="large-12 columns">
+  <table class="full">
+   <thead>
+      <tr>
+        <th>{% trans "Label" %}</th>
+        <th>{% trans "Model" %}</th>
+        <th>{% trans "Serial number" %}</th>
+        <th>{% trans "Speed" %}</th>
+        <th>{% trans "Size" %}</th>
+        <th>{% trans "Count" %}</th>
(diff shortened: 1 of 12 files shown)


src/ralph/back_office/templates/backofficeasset/components.html
```diff
@@ -0,0 +1,32 @@
...
+    <tbody>
+      {% for item in components %}
+        <tr>
+          <td>{{ item.label }}</td>
+          <td>{{ item.model }}</td>
+          <td>{{ item.serial_number }}</td>
+          <td>{{ item.speed }}</td>
+          <td>{{ item.size }}</td>
+          <td>{{ item.count }}</td>
+          <td>{{ item.action }}</td>
```
ghost, line 26: Could it be a a template tag which takes `cols` and `rows`?

szok, line 26: These are just samples, no logic, later it will look completely different.

ghost:
:ok:

ar4s:
:ok:

src/ralph/data_center/views.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+        context['data_classification'] = [
+            {'label': 'Availability', 'value': 'A3'},
+            {'label': 'Confidentiality', 'value': 'C2'},
+            {'label': 'Integrity', 'value': 'I4'},
+        ]
+        context['logs'] = [
+            '/var/log',
+        ]
+        context['critical_pathes'] = [
```
vi4m, line 97: paths!

mkurek, line 97: should be patches here

PR #1528, opened 2015-06-26
Hardware information in newer ASAs is different. Small fix for this

jjagodzinski:
Different models have different results:
 Hardware:   ASA5540, 2048 MB RAM, CPU Pentium 4 2000 MHz
 Hardware:   ASA5545, 12288 MB RAM, CPU Lynnfield 2660 MHz, 1 CPU (8 cores)
Change variables: model, ram, cpu to table where more variables can
be stored

### DIFF
--- a/src/ralph/scan/plugins/ssh_cisco_asa.py
+++ b/src/ralph/scan/plugins/ssh_cisco_asa.py
@@ -110,7 +110,7 @@ def scan_address(ip_address, **kwargs):
         ssh.close()
     pairs = parse.pairs(lines=[line.strip() for line in lines])
     sn = pairs.get('Serial Number', None)
-    model, ram, cpu = pairs['Hardware'].split(',')
+    model, ram, cpu = pairs['Hardware'].split(',')[:3]
     boot_firmware = pairs['Boot microcode']
     macs = []
     for i in xrange(99):


vi4m:
@xliiv

src/ralph/scan/plugins/ssh_cisco_asa.py
```diff
@@ -110,7 +110,8 @@ def scan_address(ip_address, **kwargs):
         ssh.close()
     pairs = parse.pairs(lines=[line.strip() for line in lines])
     sn = pairs.get('Serial Number', None)
-    model, ram, cpu = pairs['Hardware'].split(',')
+    # hardware = model, ram, cpu, corenum*
+    hardware = pairs['Hardware'].split(',')
```
ghost: You can write it as:

```
model, ram, cpu, corenums = pairs['Hardware'].split(',')
```

then the comment is useless ( + update rest of the code to reflect the change).

jjagodzinski: It is not possible. Some devices have 3 variables (model, ram, cpu) and some have 4 variables (model, ram, cpu, corenum). I can remove comment.

ghost: So it's "ok" to me. :)

ghost:
:ok:

src/ralph/scan/plugins/ssh_cisco_asa.py
mkurek: so maybe sth like this:

```
model, ram, cpu = pairs['Hardware'].split(',')[:3]
```

this will work even if in some new devices they'll add another variable and we'll keep nice variables naming instead of `hardware[0]` etc.

jjagodzinski: ok, changed

mkurek:
:ok:

Issue #1509, opened 2015-06-23
BO models importer/exporter


mkurek:
Working importer/exporter for BO models:
- [x] Asset (BO)
- [x] OfficeInfo (should be merged in BackOfficeAsset?)
- [x] AssetManufacturer
- [x] AssetModel
- [x] AssetCategory
- [x] Warehouse

and others see #1526

PR #1534, opened 2015-06-29
Add Back Office Asset export resource

szok:
Connected to #1509

### DIFF
--- a/src/ralph/export_to_ng/management/commands/export_model.py
+++ b/src/ralph/export_to_ng/management/commands/export_model.py
@@ -42,7 +42,7 @@ SIMPLE_MODELS = [
 DEPENDENT_MODELS = [
     'AssetModel',
     # new scan handles that: 'GenericComponent',
-    # 'BackOfficeAsset',
+    'BackOfficeAsset',
     'ServerRoom',
     'Rack',
     'DataCenterAsset',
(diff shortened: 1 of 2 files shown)


src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -112,7 +113,7 @@ def handle(self, *args, **options):
         models = args
         if not models:
             models = SIMPLE_MODELS + DEPENDENT_MODELS + MANY_TO_MANY
-
+        # models = ['BackOfficeAsset']  # TODO usun
```
ar4s: :)

ar4s:
:ok:

src/ralph/export_to_ng/resources.py
```diff
@@ -42,20 +45,67 @@ class Meta:
...
+        return username
+
+    def dehydrate_hostname(self, asset):
+        try:
+            hostname = asset.device_info.ralph_device.name or ''
+        except AttributeError:
+            hostname = ''
+        return hostname
+
+    def dehydrate_service_env(self, asset):
```
mkurek, line 97: try to export (and then import) service-env in format `"<service_id>|<env_id>"`

ghost:
:ok:

src/ralph/export_to_ng/resources.py
mkurek, line 97: as discussed, if we'll find it necessary, this will be handled in separated issue/PR for DC and BO at the same time.

mkurek:
:ok:

Issue #1508, opened 2015-06-23
DC models importer/exporter


mkurek:
working exporter & importer for DC models:
- [x] Environment
- [x] Service
- [x] Category
- [x] DataCenter
- [x] ServerRoom
- [x] Rack
- [x] AssetModel
- [x] Asset (only DC, without parts)
- [x] IPAddress
- [x] Network
- [x] Accessory
- [x] Manufacturer
- [x] RackAccessory
  ...

In addition:
DataCenterAssetResource spits service_env as names, like "google.com|prod". It should be replaced by id. This involves updating DataCenterAsset (in NG) resource.

PR #1535, opened 2015-06-29
Added new resources NetworkResource, IPAddressResource.

ghost:
Two new resources for exporter.

Connected to #1508.

### DIFF
--- a/src/ralph/export_to_ng/resources.py
+++ b/src/ralph/export_to_ng/resources.py
@@ -7,8 +7,9 @@ from __future__ import unicode_literals
 from import_export import fields
 from import_export import resources
 from ralph.cmdb import models_ci
-from ralph.discovery import models_device
 from ralph.discovery import models_component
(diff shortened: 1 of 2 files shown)


src/ralph/export_to_ng/resources.py
```diff
@@ -216,6 +217,74 @@ def get_queryset(self):
...
+            'kind',
+            # 'network_environment',
+        )
+        model = models_network.Network
+
+
+class IPAddressResource(resources.ModelResource):
+    asset = fields.Field('asset', column_name='asset')
+    def dehydrate_asset(self, ip_address):
+        return ip_address.device_id or ''
```
mkurek: are you sure that by exporting device_id we'll be able to match it to asset in ng (which is mapped probably to asset_id in mapping table)?

mkurek:
:ok:

szok:
:ok:

PR #1543, opened 2015-07-01
Add import Licence and LicenceAsset models.

szok:


src/ralph/back_office/templates/backofficeasset/bo_asset_licence.html
```diff
@@ -0,0 +1,39 @@
+{% extends BASE_TEMPLATE %}
```
mkurek, line 1: Can't we use Django admin somehow? For example show list based on model and fields chosen in class definition instead of template ? Basically this list should (?) be similar to list od all licences (filtered by asset)

szok, line 1: Yes, we can and we will. But it is not yet ready (Arek working on it).
We wanted to add licenses for demo :)

mkurek, line 1: ok

src/ralph/data_importer/widgets.py
```diff
@@ -37,6 +66,32 @@ def clean(self, value):
...
 
+class AssetWidget(widgets.ForeignKeyWidget):
+
+    """Widget for BO/DC Asset."""
+
+    def clean(self, value):
+        if not value:
+            return None
+        result = None
+        value = value.split("|")  # dc/bo, pk
```
ghost: `asset_type, asset_id = value.split("|")`?

ghost:
:ok:

src/ralph/data_importer/widgets.py
szok: Ok :)

ar4s:
:ok:

PR #1545, opened 2015-07-01
Added parent exporting to DataCenterAssetResource.

ghost:
DataCenterAssetResource now exports parent as well + added a few minor
updates (return empty string, instead of None)


mkurek:
:ok:

ar4s:
:ok:

PR #1548, opened 2015-07-02
iDRAC autoscan fix

jjagodzinski:
iDRAC 1.57.57 returns header server as Embedthis-http. Added this
to discovery.http.guess_family(). Fixed bug with decoding response
from server (UnicodeDecodeError). discovery.http.get_http_info() is
able to discover http family for iDRAC.

### DIFF
--- a/src/ralph/discovery/http.py
+++ b/src/ralph/discovery/http.py
@@ -59,11 +59,18 @@ def get_http_info(ip):
                 headers, content = response.headers, response.content
             except requests.exceptions.RequestException as e:
                 logger.error("Exception: {} for url: {}".format(e, url))
+                url = "https://{}/login.html".format(ip)
+                try:
+                    response = http_get_method(session, url)
+                    response.raise_for_status()
+                    headers, content = response.headers, response.content
+                except requests.exceptions.RequestException as e:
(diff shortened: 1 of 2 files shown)


vi4m:
LGTM but other person required

vi4m:
@jjagodzinski you can check your source pep8 compat  with `make flake`

src/ralph/discovery/http.py
```diff
@@ -105,15 +112,15 @@ def guess_family(headers, document):
...
             family = 'VTL'
-    elif family in ('Mbedthis-Appweb', 'Embedthis-Appweb'):
-        if '/sclogin.html?console' in document:
+    elif family in ('Mbedthis-Appweb', 'Embedthis-Appweb', 'Embedthis-http'):
+        document = document.decode("utf8")
+        if 'sclogin.html' in document:
             family = 'Dell'
         elif 'Juniper' in document:
             family = 'Juniper'
+        family = 'Dell'
```
ghost: It won't get value `juniper`, becuase it will be always `dell`.

jjagodzinski: OK, changed.

ghost:
:ok:

PR #1550, opened 2015-07-02
Add Support models to import resource

szok:
Added Support and SupportType model resources.


vi4m:
Where is the description?

ar4s:
:ok:

mkurek:
:ok:

PR #1559, opened 2015-07-06
Fixed export order for service-env model.

ghost:
ServiceEnvironment model should be imported before Assets so it must also be
exported before Assets.


szok:
:ok:

mkurek:
:ok:

PR #1561, opened 2015-07-06
Beautiful validation styling

vi4m:
# After:
## Option 3

zrzut ekranu 2015-07-06 o 11 30 20
# Before

zrzut ekranu 2015-07-06 o 10 53 58


quamilek:
:two:

szok:
:ok: :three:

mkurek:
:ok: :three:

PR #1564, opened 2015-07-06
Fixed ldap sync + minor MySQL tweaks

mkurek:
- fixed conversion from bytes to str in python 3 (see https://docs.python.org/3/library/stdtypes.html#str)
- set at least STRICT_TRANS_TABLES MySQL mode (common for Vagrant (custom mysql config) and docker (by default in mysql-5.6 image))
- fixed creating superuser (move get_user_model import after django setup)
- fixed saving user with None gender or country - default value is set now
- moved LDAP_SERVER_OBJECT_USER_CLASS to base settings
- set DJANGO_SETTINGS_MODULE to ralph.settings.local in Vagrant by default
- mysql version in vagrant bumped to 5.6 (the same minor version)

See http://blog.ionelmc.ro/2014/12/28/terrible-choices-mysql/ for details about mysql config

### DIFF
--- a/src/ralph/accounts/models.py
+++ b/src/ralph/accounts/models.py
@@ -69,3 +69,12 @@ class RalphUser(AbstractUser):
 
     class Meta(AbstractUser.Meta):
         swappable = 'AUTH_USER_MODEL'
+
+    def save(self, *args, **kwargs):
+        # set default values if None provided
+        for field in ('gender', 'country'):
+            val = getattr(self, field)
+            if val is None:
+                val = self._meta.get_field_by_name(field)[0].default
+                setattr(self, field, val)
+        return super().save(*args, **kwargs)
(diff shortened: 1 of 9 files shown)


src/ralph/accounts/models.py
```diff
@@ -69,3 +69,12 @@ class RalphUser(AbstractUser):
 
     class Meta(AbstractUser.Meta):
         swappable = 'AUTH_USER_MODEL'
+
+    def save(self, *a, **kw):
```
ar4s: `*args, **kwargs`

src/ralph/settings/base.py
```diff
@@ -115,6 +114,8 @@
 CHECK_IP_HOSTNAME_ON_SAVE = True
 ASSET_HOSTNAME_TEMPLATE = 'test'
 
+LDAP_SERVER_OBJECT_USER_CLASS = 'user'  # possible values: "user, person
```
ar4s: `# possible values: user, person`

ar4s:
:ok:

szok:
:ok:

PR #1567, opened 2015-07-07
Fixed searching in raw_id-popup.

ghost:
Before this change, user after search in raw_id popup couldn't pick item,
because `change` view was shown instead of picking the item.
This change fixes that.


mkurek:
:ok:

szok:
:ok:

PR #1569, opened 2015-07-07
Add another "model name" for Dells R630 with 10 disks

jjagodzinski:
We need to know which servers have 10 or 8 disks. This fix will change
model_name to "Dell r630 - 10 disks slot" if there is 10 disk in chassis.
Added also some fixes for discovery/http.py which repair some problems
with discovery

### DIFF
--- a/src/ralph/scan/plugins/idrac.py
+++ b/src/ralph/scan/plugins/idrac.py
@@ -243,6 +243,24 @@ def _get_memory(idrac_manager):
     ]
 
 
+def _get_enclosures(idrac_manager):
+    tree = idrac_manager.run_command('DCIM_EnclosureView')
+    xmlns_n1 = XMLNS_N1_BASE % "DCIM_EnclosureView"
+    q = "{}Body/{}EnumerateResponse/{}Items/{}DCIM_EnclosureView".format(
+        XMLNS_S,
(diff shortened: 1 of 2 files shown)


src/ralph/scan/plugins/idrac.py
```diff
@@ -335,6 +353,11 @@ def idrac_device_info(idrac_manager):
     fibrechannel_cards = _get_fibrechannel_cards(idrac_manager)
     if fibrechannel_cards:
         device_info['fibrechannel_cards'] = fibrechannel_cards
+    slotnumber = _get_enclosures(idrac_manager)
+    if slotnumber:
+        if (slotnumber[0] == '10'
+                and device_info['model_name'] == 'Dell PowerEdge R630'):
+            device_info.update({'model_name': 'Dell r630 - 10 disks slot'})
```
mkurek, line 360: we cannot accept it in current form - this is very custom case, probably used internally. Besides that you shouldn't rely on device model and use asset model instead.

Why do you want to extract this case as separated model?

ghost, line 360: I agree with @mkurek

mkurek:
@jjagodzinski do you still want to merge it (after our discussion, that now models should be provided by user in assets module)?

PR #1573, opened 2015-07-07
Add additional columns to IPAddress model admin

szok:
Added:
- Additional columns to IPAddress
- Raw ID Fields to IPAddress and Rack

### DIFF
--- a/src/ralph/settings/dev.py
+++ b/src/ralph/settings/dev.py
@@ -21,3 +21,7 @@ DEBUG_TOOLBAR_CONFIG = {
 }
 
 ROOT_URLCONF = 'ralph.urls.dev'
+
+# Disable exception for the missing element in SiteTree
+# https://github.com/idlesign/django-sitetree/pull/157/files
+RAISE_ITEMS_ERRORS_ON_DEBUG = False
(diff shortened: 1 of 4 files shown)


src/ralph/settings/dev.py
```diff
@@ -21,3 +21,5 @@ def only_true(request):
 }
 
 ROOT_URLCONF = 'ralph.urls.dev'
+
+RAISE_ITEMS_ERRORS_ON_DEBUG = not DEBUG
```
vi4m: what is this? If it's really needed, comment it clearly

src/ralph/settings/dev.py
```diff
@@ -21,3 +21,7 @@ def only_true(request):
 }
 
 ROOT_URLCONF = 'ralph.urls.dev'
+
+# Disable exception for the missing element in SiteTree
+# https://github.com/idlesign/django-sitetree/pull/157/files
+RAISE_ITEMS_ERRORS_ON_DEBUG = not DEBUG
```
mkurek: why not explicitly False?

mkurek:
:ok:

ghost:
:ok:

PR #1574, opened 2015-07-07
Removed environments from ServiceAdmin

ar4s:
I removed `enivronments` from admin's page because we manage relation between Service-Environment in `ServiceEnvironment` admin.

### DIFF
--- a/src/ralph/assets/admin.py
+++ b/src/ralph/assets/admin.py
@@ -28,6 +28,7 @@ class ServiceEnvironmentAdmin(RalphAdmin):
 @register(Service)
 class ServiceAdmin(RalphAdmin):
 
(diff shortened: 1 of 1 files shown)


ghost:
:ok: add desc. please ;)

mkurek:
the truth is that we want to select available environments on service page :)

mkurek:
check this out:

```
from django.contrib.admin import TabularInline

...

class ServiceEnvironmentInline(TabularInline):
    model = ServiceEnvironment


@register(Service)
class ServiceAdmin(RalphAdmin):
    exclude = ('environments',)
    inlines = (ServiceEnvironmentInline,)
    search_fields = ['name']
```

To discuss if there is a way to add only service (with selected environments) and then go back to service-envs list.

PR #1575, opened 2015-07-07
Added class `required` if field is required

ar4s:
Added required class to label by removed custom template for label


mkurek:
if you'll add description then :ok: :)

ghost:
@ar4s add desc. and i'll put `ok`

PR #1578, opened 2015-07-07
Fixed missing form fields

mkurek:
When someone without permission to some field open the form for the first time,
`fieldsets` fields were overwritted by fields without that particular field and
this change was later propagated for all users. After this fix, admin class
fieldsets are no longer overwritten, but operations are made on deep copy of them.


szok:
:ok:

ghost:
:ok: + I like the description which justify change. :)

PR #1579, opened 2015-07-08
Add str method to Rack model

szok:
Added:
- **str** method to Rack
- Filter racks by DataCenter
- Display name and server room in rack list

### DIFF
--- a/src/ralph/data_center/models/physical.py
+++ b/src/ralph/data_center/models/physical.py
@@ -164,6 +164,13 @@ class Rack(NamedMixin.NonUnique, models.Model):
     class Meta:
         unique_together = ('name', 'server_room')
 
+    def __str__(self):
+        return "{} ({}/{})".format(
+            self.name,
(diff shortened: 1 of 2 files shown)


src/ralph/data_center/models/physical.py
```diff
@@ -164,6 +164,13 @@ class Rack(NamedMixin.NonUnique, models.Model):
     class Meta:
         unique_together = ('name', 'server_room')
 
+    def __str__(self):
+        return "{} ({}/{})".format(
+            self.name,
+            self.server_room.data_center,
```
ghost, line 170: shouldn't  this
`self.server_room.data_center,`
be
`self.server_room.data_center.name,`
?

szok, line 170: Beacuse DataCenter **str** return self.name.

ghost:
:ok:

ar4s:
:ok:

PR #1580, opened 2015-07-08
Fixed adding service and rack

mkurek:
In both model exclude many to many with through table from basic fields
(environments and accessories) and use inlines instead.

Improved version of https://github.com/allegro/ralph/pull/1574


ar4s:
:ok:

szok:
:ok:

PR #1582, opened 2015-07-08
Add permission to display list

szok:
Added permission by field to django admin display list


src/ralph/lib/permissions/admin.py
```diff
@@ -42,3 +42,14 @@ def get_form(self, request, obj=None, **kwargs):
         return super(PermissionAdminMixin, self).get_form(
             request, obj, **kwargs
         )
+
+    def get_list_display(self, request):
+        """Return only field respects user permissions."""
```
mkurek: Return fields with respect to user permissions.

mkurek:
:ok:

ghost:
:ok:

PR #1589, opened 2015-07-08
Fixed collectstatic in docker

mkurek:
Use copying files instead of linking in docker collectstatic (nginx could not resolve links on mounted volume)


szok:
:ok:

ar4s:
:ok:

PR #1597, opened 2015-07-10
Better sorting in admin

ar4s:
For better sorting in admin.


xliiv:
:ok:

mkurek:
models should be ordered first by manufacturer name, then by model name - is it working like that now?

ar4s:
TODO: tests

src/ralph/admin/change_list.py
```diff
@@ -0,0 +1,42 @@
...
+                try:
+                    none, pfx, idx = p.rpartition('-')
+                    field_name = self.list_display[int(idx)]
+                    fields = self.get_ordering_from_related_model_admin(
+                        pfx, field_name
+                    )
+                    if fields:
+                        # replace one field by list of ordering fields
+                        ordering.remove('{}{}'.format(pfx, field_name))
+                        ordering = fields + ordering
```
mkurek: what if before this loop ordering was `['a', 'b', 'c']` and here we want to replace `'b'` to `'b.1', 'b.2'`? As far as i understand this code you're changing sort order here. Shouldn't it be sth like 

```
index = ordering.index('{}{}'.format(pfx, field_name))
ordering[index:index+1] = fields
```

?

src/ralph/tests/models.py
```diff
@@ -12,3 +12,14 @@
...
 class Foo(AdminAbsoluteUrlMixin, models.Model):
     bar = models.CharField('bar', max_length=50)
+
+
+class Manufacturer(models.Model):
+    name = models.CharField(max_length=50)
+    country = models.CharField(max_length=50)
+
+
+class Car(models.Model):
```
xliiv: ?

src/ralph/admin/tests/tests_functional.py
```diff
@@ -200,6 +203,100 @@ class FooAdmin(RalphAdmin):
...
+            request=request,
+            list_display=list_display
+        )
+        return cl.get_ordering(request, model.objects.all())
+
+    def test_sort_related(self):
+        ordering = self._get_ordering_list(
+            Car, CarAdmin, [1, 2], ['pk', 'name', 'manufacturer']
+        )
+        test_data = [
```
mkurek: consider using [ddt](http://ddt.readthedocs.org/en/latest/index.html) for such a test

src/ralph/admin/tests/tests_functional.py
```diff
@@ -200,6 +203,100 @@ class FooAdmin(RalphAdmin):
...
+        cl = self._change_list_factory(
+            model=Car,
+            model_admin=CarAdmin,
+            request=request,
+            list_display=list_display
+        )
+        return cl.get_ordering(request, model.objects.all())
+
+    def test_sort_related(self):
+        ordering = self._get_ordering_list(
```
mkurek, line 307: this isn't probably needed anymore

src/ralph/admin/views/main.py
```diff
@@ -0,0 +1,45 @@
...
+                        pfx, field_name
+                    )
+                    # replace one field by list of ordering fields
+                    # or move field at the begining
+                    old_field = '{}{}'.format(pfx, field_name)
+                    ordering.remove(old_field)
+                    ordering = (fields or [old_field]) + ordering
+                except (IndexError, ValueError):
+                    raise
+                    continue
```
mkurek: so raise or continue? :)

ghost:
:ok:

mkurek:
:ok:

PR #1599, opened 2015-07-10
Fixed titles in views.

ghost:
Get titles in views from model.Meta.
For listing-view as plural.
For edit-view as singular.

### DIFF
--- a/src/ralph/admin/helpers.py
+++ b/src/ralph/admin/helpers.py
@@ -1,0 +1,3 @@
+from django.apps import apps
+
+APP_MODELS = {model._meta.model_name: model for model in apps.get_models()}
--- a/src/ralph/admin/sitetrees.py
+++ b/src/ralph/admin/sitetrees.py
@@ -9,6 +9,7 @@ from sitetree.utils import (
     item,
 )
 
+from ralph.admin.helpers import APP_MODELS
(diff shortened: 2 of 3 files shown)


src/ralph/admin/sitetrees.py
```diff
@@ -10,6 +10,7 @@
 )
 
 from ralph.admin.sites import ralph_site
+from ralph.data_importer.management.commands.importer import APP_MODELS
```
ar4s: Move `APP_MODELS` to new module like `ralph.admin.helpers`

mkurek: it's only one line so we can redefine it here. besides that - it's definition (`APP_MODELS = {model._meta.model_name: model for model in apps.get_models()}`) breaks when there will be model with the same name in more than one app

src/ralph/admin/helpers.py
```diff
@@ -0,0 +1,3 @@
+from django.apps import apps
+
+APP_MODELS = {model._meta.model_name: model for model in apps.get_models()}
```
mkurek, line 3: it breaks when there will be model with the same name in more than one app

ghost:
Change page titles come from model instead of sitetree.

PR #1600, opened 2015-07-10
Remove Network foreign key from IPAddress model 

szok:
Remove Network foreign key from IPAddress model


ar4s:
:ok:

mkurek:
:ok:

mkurek:
add unit test for that

mkurek:
http://www.nooooooooooooooo.com/ 

Don't run reset.sh - create 0002 migration with removing network field - we'll squash all migrations into one file someday. 

Btw could you clean-up network tab in data-center asset? 
- [ ] Ip should be added (new ip) or selected from list of existing ip addresses (or at least saving with existing ip should not raise an error - it should assign asset to this ip). 
- [ ] There could be only one management address
- [ ] Cannot add new rows (jquery error)

ar4s:
_Cannot add new rows (jquery error)_ - @szok you can omit this bug (unless you know how fix this) - I will take care of it in next week.

szok:
@mkurek

PR #1602, opened 2015-07-13
Changed page titles comes from model name instaed of sitetree.

ghost:
Some of models shouldn't be added to menu so you can't rely on that data, so
sitetree data is replaced with model's data.


ar4s:
:ok:

mkurek:
:ok:

PR #1605, opened 2015-07-15
Add managment IP validation

szok:
Added validation to managment asset IP

### DIFF
--- a/src/ralph/data_center/forms/network.py
+++ b/src/ralph/data_center/forms/network.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from django.forms import ValidationError
+from django.forms.models import BaseInlineFormSet
+
+
+def validate_is_management(forms):
+    """
+    Validate is_management field in IpAddress formset
+    """
+    is_management = []
+    for form in forms:
+        cleaned_data = form.cleaned_data
+        if (
+            cleaned_data and
+            not cleaned_data.get('DELETE', False)
+        ):
+            is_management.append(cleaned_data.get('is_management'))
+
+    count_management_ip = is_management.count(True)
+    if is_management and count_management_ip == 0:
+        raise ValidationError(
+            'One IP address must be management'
+        )
+    elif is_management and count_management_ip != 1:
+        raise ValidationError((
+            'Only one managment IP address can be assigned '
+            'to this asset'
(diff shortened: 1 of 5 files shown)


src/ralph/data_center/models/networks.py
```diff
@@ -529,15 +514,7 @@ def save(self, allow_device_change=True, *args, **kwargs):
         if settings.CHECK_IP_HOSTNAME_ON_SAVE:
             if not self.address:
                 self.address = network.hostname(self.hostname, reverse=True)
-            if not self.hostname:
-                self.hostname = network.hostname(self.address)
         self.number = int(ipaddress.ip_address(self.address))
-        try:
-            self.network = Network.from_ip(self.address)
```
mkurek, line 536: remove this method from Network model too - if we'll need this in the future we'll copy it from Ralph 2.0 - for now it's useless and unused code.

src/ralph/data_center/forms/network.py
```diff
@@ -0,0 +1,21 @@
...
+        is_management = []
+        for form in self.forms:
+            cleaned_data = form.cleaned_data
+            if (
+                cleaned_data and
+                not cleaned_data.get('DELETE', False)
+            ):
+                is_management.append(cleaned_data.get('is_management'))
+
+        if is_management.count(True) > 1:
```
mkurek: `if is_management and is_management.count(True) != 1` - if there is any ip, one should be marked as management

ghost:
:ok:

src/ralph/data_center/forms/network.py
ghost: `count` on empty list also works so:
`if  is_management.count(True) != 1`
is enough, what do you think?

src/ralph/data_center/tests/test_forms.py
```diff
@@ -0,0 +1,45 @@
...
+        with self.assertRaisesRegexp(
+            ValidationError,
+            (
+                'Only one managment IP address can be assigned '
+                'to this asset'
+            )
+        ):
+            validate_is_management([form_1, form_2])
+
+    def test_vlaidate_not_management(self):
```
mkurek: validate

mkurek:
:ok:

Issue #1611, opened 2015-07-15
Investigate licences source code


mkurek:
https://github.com/allegro/ralph/blob/ng/src/ralph/licences/models.py#L196

Check if this code is still necessary.

PR #1616, opened 2015-07-15
Replaced old code by fildsets in admin

ar4s:
Old code in licence module is no more useful in new Ralph - is replaced by
inlines in licence's admin.

Connected to #1611

### DIFF
--- a/src/ralph/licences/admin.py
+++ b/src/ralph/licences/admin.py
@@ -1,17 +1,56 @@
 # -*- coding: utf-8 -*-
+from django.contrib.admin import TabularInline
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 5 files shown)


src/ralph/licences/admin.py
```diff
@@ -22,6 +41,7 @@ class LicenceAdmin(PermissionAdminMixin, RalphAdmin):
     list_select_related = ['licence_type', 'software_category']
     raw_id_fields = ['software_category', 'manufacturer']
     resource_class = resources.LicenceResource
+    inlines = [BaseObjectLicenceInline, LicenceUserInline]
```
mkurek: add these inlines to tabs

ghost:
:ok:

szok:
:ok:

PR #1617, opened 2015-07-15
Fixed CMDB `Jira-Changes` filtering.

ghost:
There were 3 bugs:
- one was the result of conflicting datepickers (bootstrap vs.
  jquery-ui) fixed by removing `jquery-ui` datepicker from source
  (Ralph should use boostrap version)
- the other was getting wrong form values to filtering (util.py)
- another was wrong range in filtering


vi4m:
xliiv -if you already fix it, could you limit number of affected ci's in this view? 
We should display something like: 

Affected Ci's: [one main ci] + 100 additional cis (show) instead of boring as fuck list.

ghost:
@vi4m I'll fix it in next pull-request.
Let's merge this one.

mkurek:
:ok:

ar4s:
:ok:

PR #1618, opened 2015-07-15
Added new filter (created in jira) to CMDB `Jira Changes`

ghost:
Added fields to form + handled them in add_filter fn.


ar4s:
:ok:

szok:
:ok:

Issue #1615, opened 2015-07-15
admin & ralph_admin cleanup


mkurek:
https://github.com/allegro/ralph/tree/ng/src/ralph/admin/templates ralph_admin & admin should be merged into one directory

ghost:
connected to #1621

PR #1621, opened 2015-07-16
Dir `admin/` templates reorganized.

ghost:
Dir admin/templates had two dirs:
    - `admin/templates/admin`
    - `admin/templates/ralph_admin`
Now they are merged into: `admin/templates/admin`

Connected to #1615


ar4s:
:ok:

szok:
:ok:

Issue #1612, opened 2015-07-15
Empty readonly fields


mkurek:
When field is readonly for current user and it has empty value it should be represented as `-` or something similar.

ghost:
connected to: #1623

PR #1623, opened 2015-07-16
Default value for read-only field.

ghost:
Show default ('-') when read-only field has no value set.

connected to: #1612


mkurek:
:ok:

ar4s:
:ok:

PR #1626, opened 2015-07-21
Better representation for CIs in `Jira Changes` (CMDB)

ghost:
Put CIs into div so that CSS overflow works and limited height to 300px.


ar4s:
:ok:

mkurek:
:ok:

vi4m:
thank you @xliiv

ghost:
No problem :)
That long list is only enclosed within a fixed-width div now, but in future we could changed it to something more relevant.

PR #1628, opened 2015-07-21
Fixed version of keystoneclient dep. packages.

ghost:
Keystoneclient package requires a few packages as '>=' which breaks
Ralph. This change fixes versions of those packages.


ar4s:
:ok:

szok:
:ok:

PR #1630, opened 2015-07-22
Fixed require.js with hashing static handler.

ghost:
Static handler exchange (to hashing one) didn't update require.js configs.
This change fixes that.


ar4s:
:ok:

szok:
:ok:

PR #1631, opened 2015-07-23
Fixed CMDB's datepickers in firefox.

ghost:
Fixed require.js configuration by adding missing depenendencies.
Removed conflicting jquery import from base.html in cmdb.


ar4s:
:ok:

szok:
:ok:

PR #1636, opened 2015-07-23
Changed issue-type filter in CMDB's views.

ghost:
Field `issue-filter` has now values relevant to view which renders it, eg:
- incident view has only incidents types
- problem view has only problems views


szok:
:ok:

ar4s:
:ok:

Issue #1610, opened 2015-07-15
Add search on details/edit view


mkurek:
- Currently search is visible only on list. It should also be visible on details/edit view.

PR #1639, opened 2015-07-24
Added search to `edit-view`.

ghost:
Search now is added to edit views of objects (e.g: supports).

connected to: #1610

### DIFF
--- a/src/ralph/admin/templates/admin/search_form.html
+++ b/src/ralph/admin/templates/admin/search_form.html
@@ -1,10 +1,10 @@
-{% load i18n admin_static %}
+{% load ralph_tags i18n admin_static %}
 
-{% if cl.search_fields %}
-  <form id="changelist-search" action="" method="get">
+{% if search_fields %}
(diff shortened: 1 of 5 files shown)


src/ralph/admin/templates/admin/search_form.html
```diff
@@ -1,44 +1,30 @@
...
       </div>
-      <div class="small-3 columns">
+      <div class="small-2 columns">
         <button type="submit" class="button expand search">
           <i class="fa fa-search"></i> {% trans 'Search' %}
         </button>
       </div>
     </div>
   </form>
-{#   <div id="toolbar"> #}
```
mkurek, line 19: if it's not necessary remove it

vi4m:
Is this really so important to implement this feature right now?  We still want to have left-sidebar search like in old version.

ghost:
@vi4m removed context-switcher, only added search to edit views.

mkurek:
:ok:

ar4s:
:ok:

PR #1648, opened 2015-07-27
Enable isort diff parameter for better description errors.

quamilek:
Now isort show diff when file have wrong order in imports


mkurek:
:ok:

szok:
:ok:

PR #1656, opened 2015-07-28
Fixed `deprecation` typo in ng exporter.

ghost:
Since RalphNG has correctly named depreciation fields, exporter should
returns correct column names, and this change fixes that.


mkurek:
:ok:

ar4s:
:ok:

PR #1657, opened 2015-07-29
[DON'T MERGE] Transitions on django-fsm in ng

ar4s:
Added transition mechanism by django-fsm library and adapt to be more flexible - configuration in database instead of in code.


mkurek:
Final PR (#1681) doesn't use fsm, so this one could be closed?

PR #1661, opened 2015-07-31
Fixed static serving.

ghost:
Django static serving was updated to use iter(..) on content instead of fully
loaded str. LCK middlerware is not ready for this, because it processes
content like it was a string. This makes static returned as empty string.


ar4s:
:ok:

mkurek:
:ok:

PR #1663, opened 2015-07-31
Allow blank default pool for LoadBalancerVirtualServer

mkurek:
LoadBalancerVirtualServer could be null - this commit enable it to be blank (ex. in admin) as well.


ghost:
:ok:

szok:
:ok:

PR #1664, opened 2015-08-03
Fixed Firefox error

ar4s:
In Firefox event.keyCode is equal 0 in keyPress event - solution: get key code
from event.which.


ghost:
:ok:

mkurek:
:ok:

Issue #1666, opened 2015-08-03
Display asset's status on device view


ghost:
Add `status` field below `model` labelled as `asset status`.

PR #1669, opened 2015-08-04
Added asset status field to device-info form.

ghost:
The field show read only status (if asset is linked to displayed device) or
empty string.

connected to: #1666

### DIFF
--- a/src/ralph/ui/forms/devices.py
+++ b/src/ralph/ui/forms/devices.py
@@ -34,6 +34,7 @@ from ralph.ui.widgets import (
 )
 from ralph.ui.forms.util import all_ventures, all_roles
 from ralph_assets.models import Asset, Orientation
+from ralph_assets.models_assets import AssetStatus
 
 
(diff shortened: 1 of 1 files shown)


src/ralph/ui/forms/devices.py
```diff
@@ -397,6 +404,16 @@ class Meta(DeviceForm.Meta):
             'deleted',
         )
 
+    def set_asset_status(self):
+        asset_status = ''
+        linked_asset = self.instance.get_asset()
+        if linked_asset:
+            asset_status = (
+            AssetStatus.NameFromID(linked_asset.status) if
```
mkurek: fix indentation here (weird - flake8 should handle it :expressionless:)

mkurek: afaik there is underscore version of this method (`name_from_id`), which is more pythonic - could you use it?

mkurek:
:ok:

ar4s:
:ok:

PR #1670, opened 2015-08-04
Vagrant local config

mkurek:
Added possibility to specify local Vagrant config (which will override default Vagrant settings).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,8 @@ src/ralph/urls/*local.py
 
 # Vagrant
 vagrant/.vagrant
+vagrant/Vagrantfile.local
+vagrant/provision.local.sh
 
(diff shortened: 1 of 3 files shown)


ar4s:
:ok:

.gitignore
```diff
@@ -60,6 +60,8 @@ src/ralph/urls/*local.py
 
 # Vagrant
 vagrant/.vagrant
+vagrant/Vagrantfile.local
+vagrant/provision.sh.local
```
ghost: Maybe `vagrant/provision.local.sh`, so that editors would be aware of syntax.

ghost:
:ok:

szok:
:ok:

Issue #1668, opened 2015-08-04
Data center asset form validation & improvements


mkurek:
Add additional data center asset form validation (from Ralph 2.0):
- [x] slot number validation (only valid, when model type is blade)
- [x] orientation validation

Fields clean-up:
- [x] fields order: move hostname, barcode, sn, service-env to top
- [x] check out, if all financial info date fields are necessary (Request date, Provider order date, Delivery date)
- [x] are required support, Production year, Production use date, Loan end date fields necessary?
- [x] should slots field be required? Shouldn't it be defined on model? What's the difference between slots and height of device?
- [x] add service uid field

General improvements:
- [x] display help-text near to field
- [x] backspace does not trigger autocomplete event

PR #1671, opened 2015-08-05
Models cleanup

mkurek:
connected to #1668 

Models & admin cleanup & improvements (performance and visual)
- added ability to override model pointed by raw_id field (usefull when fields are inherited and we want to specify descending class)
- fixed edit related and add related (openinig in popup now)
- auto-complete widget fixes: hide by default, trigger on backspace, hide when characters count < 2
- fetch related objects by default in auto-complete widgets (performance improvement to not call every table in sequence but use join and fetch them at once)
- changed order of AssetModel fields
- introduced `NullableCharField`, which saves null instead of '' to database
- changed some fields from `CharField` to `NullableCharField`: `barcode`, `niw`, `sn`, `wwn`
- new `Service` field - `uid`
- removed some `Asset` fields:  `provider_order_date`, `request_date`
- removed `slots` field from `DataCenterAsset`
- moved `loan_end_date` and `purchase_order` from `Asset` to `BackOfficeAsset`
- moved `delivery_date`, `production_use_date`, `production_year` to `DataCenterAsset`, `source`
- changed fields order (and fieldsets) in `DataCenterAssetAdmin`
- all data center models are now available to import from `ralph.data_center.models`
- help_text is now displayed as tooltip
- added orientation, slot_no and position validation in `DataCenterAsset`

screencapture-127-0-0-1-8000-data_center-datacenterasset-add-1438949372776

### DIFF
--- a/src/ralph/__main__.py
+++ b/src/ralph/__main__.py
@@ -3,8 +3,11 @@ import os
 import sys
 
 
-def main(settings_module='ralph.settings'):
-    os.environ.setdefault('DJANGO_SETTINGS_MODULE', settings_module)
+def main(settings_module='ralph.settings', force=False):
+    if force:
+        os.environ['DJANGO_SETTINGS_MODULE'] = settings_module
+    else:
+        os.environ.setdefault('DJANGO_SETTINGS_MODULE', settings_module)
 
     from django.core.management import execute_from_command_line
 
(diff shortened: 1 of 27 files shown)


ar4s:
:ok:

szok:
:ok:

PR #1672, opened 2015-08-05
Fixed internal server error during autocomplete AJAX request

ar4s:
After close popup (by select item or add) autocomplete's detail view raised 500.


ghost:
:ok:

mkurek:
:ok:

PR #1675, opened 2015-08-06
Templates improvements

mkurek:
- added new template loader, which allow to specify app label, ex. when extending template
- added additional buttons block to submit_line template
- added helper to simplify getting old object pk from new object

### DIFF
--- a/src/ralph/lib/template/loaders.py
+++ b/src/ralph/lib/template/loaders.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+import io
+from os.path import abspath, dirname, join
+
+from django.apps import apps
+from django.template import TemplateDoesNotExist
+from django.template.loaders.base import Loader as BaseLoader
+
+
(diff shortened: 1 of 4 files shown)


ar4s:
:ok:

src/ralph/lib/template/loaders.py
```diff
@@ -0,0 +1,68 @@
...
+        return join(template_dir, template_name)
+
+    def load_template_source(self, template_name, template_dirs=None):
+        filepath = self.get_template_path(template_name, template_dirs)
+        try:
+            with io.open(filepath, encoding=self.engine.file_charset) as fp:
+                return fp.read(), filepath
+        except IOError:
+            pass
+        raise TemplateDoesNotExist(template_name)
```
ghost: this could be instead of `pass`?

mkurek: done

szok:
:ok:

Issue #1641, opened 2015-07-24
Fix the layout, and add the description for the report


vi4m:
Current reports don't provide any description. Could be just 2-3 lines what's the purpose and how to use it. 

Styling is a shitty too - too wide.

PR #1683, opened 2015-08-10
Add report description.

szok:
Adding a simple description for reports

Connected to #1641


mkurek:
I would replace descriptions to `Number of assets in ....` (ex. `Number of assets in each model category`), eventually `.. with each ..`, @vi4m what do you think?

vi4m:
:ok:

ar4s:
:ok:

PR #1684, opened 2015-08-10
Fixed autocomplete widget when admin does not have list_select_related defined

mkurek:
Fixed autocomplete widget when admin does not have list_select_related defined (default value for `list_select_related` is `False` :expressionless:)


ar4s:
:ok:

ghost:
:ok:

Issue #1688, opened 2015-08-11
Autocomplete result list is not hidden


mkurek:
After second search (and selecting anything) in the same field result list is not hidden.

PR #1698, opened 2015-08-11
Fixed behaviour suggest list when user click on item

ar4s:
After click on item suggest list will be close.

Connected to #1688


mkurek:
:ok:

szok:
:ok:

Issue #1606, opened 2015-07-15
Mass adding asset


mkurek:
- should be similar to mass adding in Ralph 2.0
- overwrite barcode and SN widgets to support multiline input
- save should validate uniqueness of barcodes/SNs, equal number of barcodes and SNs (equal number of lines - could be empty if asset doesn't have barcode or sn, but at least one should be specified) and if it's valid save multiple assets
- after save show message that X assets were created with following barcodes/SNs

At the end (not necessarily in one PR/issue/sprint):
- improve widget: add single/multiple mode button which will change number of rows in textarea

damian1976:
IMHO: when validating multiline SNs/barcodes please remove leading/trailing newline(s) (if any) using lstrip()/rstrip(). Currently in 2.x the case when message saying "New line exists. Please remove it" appears is unnecessary

ghost:
@damian1976 I don't see such text in ralph_assets sources. Are you sure it's still present?

damian1976:
@xliiv if possible, please try do add an asset with SN and/or Barcode ending with newline character

PR #1704, opened 2015-08-12
[wip] Ng bulk adding

ghost:
working version

Connected to #1606

### DIFF
--- a/src/ralph/admin/fields.py
+++ b/src/ralph/admin/fields.py
@@ -57,14 +57,34 @@ class MultivalueFormMixin(object):
     one_of_mulitvalue_required = []
 
     @transaction.atomic
-    def save_model(self, request, obj, form, change):
-        super().save_model(request, obj, form, change)
-        obj = self.model(**form.cleaned_data)
-        for idx, form_values in enumerate(self.multivalues_rows()):
-            for value, field_name in zip(form_values, self.multivalue_fields):
-                setattr(obj, field_name, value)
-        msg = _('Successfully added {} assets'.format(idx + 1))
-        self.message_user(request, msg)
+    def save(self, commit=True):
+        import copy
+        objs = []
+        for row_as_dict in self.multivalues_rows_as_dict():
+            obj_data = copy.deepcopy(self.cleaned_data)
+            #import ipdb;ipdb.set_trace()
+            obj_data.update(row_as_dict)
+            obj = self._meta.model(**obj_data)
+            #import ipdb;ipdb.set_trace()
+            obj.save()
+            objs.append(obj)
+        if len(objs) > 1:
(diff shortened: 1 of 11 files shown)


damian1976:
IMHO when bulk adding assets I would change the old method relying on 2 textareas(SN + Barcode)  into triples of 'SN' + 'Barcode' + 'Position' next to each other. In order to add next triple '+' Button would exist underneath. By clicking on we would activate the next triple below. In this way, in one step we can create new assets and place them correctly inside the rack. Additionally when clicking on '+', 'Position' 
field in **Location Info** would have to be ignored/disabled to let the user know that positions are now read from these located next to SN/Barcode textfields.
Does it sound reasonably?
Anyhow, the idea is to introduce 'Position' field to every asset added in 'Bulk add' mode.

src/ralph/admin/tests/tests_functional.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- cmding: utf-8 -*-
```
vi4m, line 1: ?

damian1976:
If you disagree with the above could you at least add position to sn/Barcode. It would really help in bulk add....

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,140 @@
...
+from django.forms.models import save_instance
+from django import forms
+from django.db import transaction
+from django.forms.util import ErrorList
+from django.utils.translation import ugettext as _
+
+
+class MultilineField(forms.CharField):
+    """
+    This field is a textarea which treats its content as many values seperated
```
vi4m, line 15: Well done descriptions across entire pull request. Very good!

vi4m:
@damian1976 we have to make sure we need this functionality too - we will ask our operations team.

damian1976:
@vi4m Did you by chance ask them about this functionality?
maybe I'm wron but how do you bulk add e.g 20 identical servers?You ad1 asset with 20 SNs, then after creating them you search them by service/env or similar, and finally go one by one setting valid position? My proposition would make it easier and would shortened the whole process to a single step...

ghost:
@damian1976 Your case could be solved by new concept (https://github.com/allegro/ralph/pull/1721). However nothing was settled and concept is still work in progress. Nonetheless the case you described should be resolved in comfortable way for the user. Editing tons of assets indyvidually for `position` or `slot number` (similar case IMHO) is unacceptable.

damian1976:
@xliiv OK, thanks. I'll keep the fingers crossed for the proposed solution:) Hope, this will facilitate the bulk adding process and setting assets in correct position inside rack without editing them continuously and setting location info once again one by one. Is it possible to prepare (testing version of course) this concept for DC asset as well ASAP?

Issue #1457, opened 2015-05-28
New attachments 


ghost:
I'd like to add attachment to asset, let's say an image.
The image (and other attachments) should be visible on asset's page as a thumbnail.

@vi4m what do you think?

damian1976:
Sometimes for the persons sitting behind the desk it's nice to be able (without going to DC) to see how given device looks like and read its specification . I had this problem and curiosity when adding devices (sometimes layout of blades in a chassis is unclear, etc.).
Mapping some kinds of layouts in ralph in 100 % is not possible. It this case the real device picture helps. As much as web url to device specification. Please look at the the following example:

![](http://e.huawei.com/~/media/EBG/Images/product/server/eh/e9000/E9000-c2.jpg)

The above information can be added in separate tab in DC Asset

mkurek:
Attachments should be generic and there should be possibility to add attachment to every object in ralph (at lest at BaseObject level, but it'd great to add them for every object).

Resources: https://www.djangopackages.com/grids/g/attachments/

PR #1706, opened 2015-08-12
Added attachments app

ar4s:
Connected to #1457

### DIFF
--- a/src/ralph/admin/mixins.py
+++ b/src/ralph/admin/mixins.py
@@ -2,7 +2,6 @@
 import os
 import urllib
 from copy import copy
-from functools import wraps
 
 from django import forms
 from django.conf import settings
@@ -12,7 +11,6 @@ from django.core import urlresolvers
 from django.core.urlresolvers import reverse
 from django.db import models
 from django.http import HttpResponseRedirect
-from django.utils.functional import curry
 from django.views.generic import TemplateView
 from import_export.admin import ImportExportModelAdmin
 from reversion import VersionAdmin
@@ -21,6 +19,7 @@ from ralph.admin import widgets
 from ralph.admin.autocomplete import AjaxAutocompleteMixin
 from ralph.admin.helpers import get_field_by_relation_path
 from ralph.admin.views.main import BULK_EDIT_VAR, BULK_EDIT_VAR_IDS
+from ralph.helpers import add_request_to_form
 from ralph.lib.mixins.forms import RequestFormMixin
 from ralph.lib.permissions.admin import PermissionsPerObjectFormMixin
 
@@ -35,6 +34,7 @@ def get_common_media():
     Shared across extra views and admin class
     """
     js = map(lambda x: os.path.join(*x), [
+        ('admin', 'js', 'core.js'),
         ('admin', 'js', 'jquery.js'),
         ('admin', 'js', 'jquery.init.js'),
         ('admin', 'js', 'actions.js'),
@@ -131,8 +131,8 @@ class RalphAdminMixin(RalphAutocompleteMixin):
         """
         Return form with request param passed by default.
         """
-        Form = super().get_form(request, obj, **kwargs)
-        return wraps(Form)(curry(Form, _request=request))
+        Form = super().get_form(request, obj, **kwargs)  # noqa
+        return add_request_to_form(Form, request=request)
 
     def get_changelist(self, request, **kwargs):
         from ralph.admin.views.main import RalphChangeList
--- a/src/ralph/attachments/admin.py
+++ b/src/ralph/attachments/admin.py
@@ -1,0 +1,43 @@
+import time
+
+from django.conf.urls import url
+from django.http import FileResponse
+from django.shortcuts import get_object_or_404
+from django.utils.http import http_date
+
+from ralph.attachments.models import Attachment
+from ralph.attachments.views import AttachmentsView
+
+
+class AttachmentsMixin(object):
+    """
+    Mixin add new URL to admin for download file.
+    """
+    def __init__(self, *args, **kwargs):
+        self.change_views.append(AttachmentsView)
+        super().__init__(*args, **kwargs)
+
+    def get_urls(self):
+        urlpatterns = super().get_urls()
+        urlpatterns += [
+            url(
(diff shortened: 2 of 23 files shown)


src/ralph/attachments/admin.py
```diff
@@ -0,0 +1,35 @@
...
+        urlpatterns += [
+            url(
+                r'^attachment/(?P<id>\d+)-(?P<filename>\S+)',
+                self.serve_attachment,
+                name='get_attachment'
+            ),
+        ]
+        return urlpatterns
+
+    def serve_attachment(self, request, id, filename):
```
mkurek, line 31: please add some TODO that we'll need permissions validation here

src/ralph/lib/foundation/templatetags/foundation_form.py
```diff
@@ -10,9 +10,12 @@ def label(field):
 
 
 @register.inclusion_tag('foundation_form/form_field.html')
-def field(field):
+def field(field, is_admin=True, show_label=True):
     """Render foundation field."""
-    return {'field': field, 'show_label': True}
+    return {
+        'field': is_admin and field.field or field,
```
ghost: why this?

src/ralph/attachments/templates/attachments/attachments.html
```diff
@@ -0,0 +1,42 @@
...
+{% block view_content %}
+  <form method="post" action="" enctype="multipart/form-data">
+    {% csrf_token %}
+    {{ formset.management_form }}
+    <table class="full" id="attachemnts">
+      <thead>
+        <tr>
+          <th>{% trans "File" %}</th>
+          <th>{% trans "Description" %}</th>
+          <th>{% trans "Delete" %}</th>
```
ghost, line 14: can we make it from loop instead of hardcoding?

src/ralph/attachments/views.py
```diff
@@ -0,0 +1,50 @@
...
+class AttachmentsView(RalphDetailView):
+    icon = 'paperclip'
+    name = 'attachment'
+    label = 'Attachments'
+    url_name = 'attachment'
+    template_name = 'attachments/attachments.html'
+
+    def get_formset(self, data=None, files=None):
+        qs = Attachment.objects.get_attachments_for_object(self.object)
+        AttachmentForm._request = self.request
```
mkurek: isn't there better way to handle this? assigning to class it's not the best idea. Maybe you could use `RalphAdminForm` (https://github.com/allegro/ralph/blob/ng/src/ralph/admin/mixins.py#L82)?

src/ralph/lib/foundation/templatetags/foundation_form.py
mkurek: as we discussed please replace it with `if .. else ..`

src/ralph/settings/base.py
```diff
@@ -106,9 +107,12 @@
 STATIC_URL = '/static/'
 
 STATICFILES_DIRS = (
-    os.path.join(BASE_DIR, 'static'),
+    os.path.join(BASE_DIR, 'var', 'static'),
```
mkurek: do we have any file in `var/static`? :speak_no_evil:

src/ralph/attachments/models.py
```diff
@@ -0,0 +1,91 @@
...
+                object_id=pk,
+                attachment=attachment,
+            ))
+        if new_items:
+            self.bulk_create(new_items)
+
+    def dettach(self, pk, content_type, attachments):
+        self.filter(attachment__in=attachments).delete()
+
+    def refresh(self, obj, new_objects=None, deleted_objects=None):
```
mkurek, line 82: shouldn't this be wrapped in transaction?

src/ralph/attachments/models.py
```diff
@@ -0,0 +1,91 @@
...
+        unique=False,
+    )
+    file = models.FileField(upload_to=get_file_path, max_length=255)
+    mime_type = models.CharField(
+        max_length=100,
+        unique=False,
+        default='application/octet-stream',
+    )
+    description = models.TextField(blank=True)
+    uploaded_by = models.ForeignKey(settings.AUTH_USER_MODEL)
```
mkurek, line 114: shouldn't it be nullable?

src/ralph/attachments/models.py
```diff
@@ -0,0 +1,91 @@
...
+
+    objects = AttachmentManager()
+
+    @staticmethod
+    def _safe_filename(filename):
+        allowed = '-_.() ' + string.ascii_letters + string.digits
```
mkurek: what if there will be some non-ascii letter in attachment name (ex. polish letter)? we're storing this file under our own name, so we could allow every (escaped of course) filename IMO.

src/ralph/admin/mixins.py
```diff
@@ -31,6 +31,10 @@ def get_common_media():
     Shared by across extra views and admin class
```
vi4m: by across? Shared across ...

vi4m:
@mkurek i don't see any blueprint describing this functionality. Where is this?

What's the purpose of separate attachments app? I could imagine, but what about Arek's son and daughter? Will they live in the world of secrets?

src/ralph/attachments/models.py
ar4s, line 114: why? I don't see any pros if this fileld will be nullable.

src/ralph/attachments/templates/attachments/attachments.html
ar4s, line 14: The ratio of benefits and time-to-implement it's to low :(

ar4s:
@vi4m I don't understand - a reusable app is a bad thing? My life is now collapsed..

vi4m:
Reusable app is great thing and i really like this code. But on the other hand I tried to say the we are lacking modules docstrings if we want to be fully open source. Standing on the "opposite" site during my vacation, without talking at the office, only shows me how much it's needed.  

Even more - I would propose, that introducing any new module should come with extensive module docstring describing its roots. This is the way Python standard library is being developed. We really want developers across entire world to be obsessed about our code readability, right?

mkurek:
:ok:

szok:
:ok:

Issue #1695, opened 2015-08-11
Advanced search


mkurek:
Add possibility to search by concrete field (like search-bar on the left in Ralph 2.0). 
- Search fields should be easily modified (ex. `search_by = ['barcode', 'sn', 'model__category', 'service__name']` in admin definition).
- Search fields should be displayed in one box with filter fields (which should be used for simple dict-fields (choices) etc).

PR #1708, opened 2015-08-13
Advanced search

szok:
Added search filter to:
- Data Center Asset
- Back Office Asset

Connected to #1695

### DIFF
--- a/src/ralph/admin/filters.py
+++ b/src/ralph/admin/filters.py
@@ -1,0 +1,163 @@
+# -*- coding: utf-8 -*-
+import re
+from datetime import datetime
+from functools import lru_cache
+
+from django.contrib.admin import SimpleListFilter
+from django.db.models import Q
+from django.utils.encoding import smart_text
+from django.utils.formats import get_format
+from django.utils.translation import ugettext_lazy as _
+
+
+@lru_cache()
+def date_format_to_human(value):
+    """
+    Convert Python format date to human.
+
+    Example:
+        >>> date_format_to_human('%Y-%m-%d')
+        YYYY-MM-DD
+
+    :param value: Date format example: %Y-%m
+    """
+    maps = {
+        '%y': 'YY',
+        '%Y': 'YYYY',
+        '%m': 'MM',
+        '%d': 'DD',
+    }
+    for k, v in maps.items():
+        value = value.replace(k, v)
+    return value
+
+
+class BaseCustomFilter(SimpleListFilter):
+
+    def has_output(self):
+        return True
+
+    def lookups(self, request, model_admin):
+        return None
+
+
+class ChoicesFilter(BaseCustomFilter):
+
+    """Renders filter form with choices field."""
+
+    template = 'admin/filters/choices_filter.html'
+    choices_list = []
+
+    def choices(self, cl):
+        yield {
+            'selected': False,
+            'value': '',
+            'display': _('All'),
+            'parameter_name': self.parameter_name
+        }
+        for lookup, title in self.choices_list:
+            yield {
+                'selected': smart_text(lookup) == self.value(),
+                'value': lookup,
+                'display': title,
+                'parameter_name': self.parameter_name,
+            }
+
+    def queryset(self, request, queryset):
+        if self.value():
+            return queryset.filter(**{self.parameter_name: self.value()})
+
+
+class BooleanFilter(ChoicesFilter):
+
+    """Renders filter form with boolean field."""
+
+    choices_list = [
+        ('1', _('Yes')),
+        ('0', _('No')),
+    ]
+
+
+class DateFilter(BaseCustomFilter):
+
+    """Renders filter form with date field."""
+
+    template = 'admin/filters/date_filter.html'
+
+    def __init__(self, request, params, model, model_admin):
+        used_parameters = {}
+        for param in self.expected_parameters():
+            try:
+                used_parameters[param] = params.pop(param, '')
+            except KeyError:
+                pass
+        super().__init__(request, params, model, model_admin)
+        self.used_parameters = used_parameters
+
+    def expected_parameters(self):
+        return [self.parameter_name_start, self.parameter_name_end]
+
+    def value(self):
+        return [
+            self.used_parameters.get(param)
+            for param in self.expected_parameters()
+        ]
+
+    def queryset(self, request, queryset):
+        if any(self.value()):
+            date = self.value()
+            try:
+                date_start = datetime.strptime(
+                    date[0], get_format('DATE_INPUT_FORMATS')[0]
+                )
+                queryset = queryset.filter(**{
+                    '{}__gte'.format(self.parameter_name): date_start,
+                })
+            except ValueError:
+                pass
+
+            try:
+                date_end = datetime.strptime(
+                    date[1], get_format('DATE_INPUT_FORMATS')[0]
+                )
+                queryset = queryset.filter(**{
+                    '{}__lte'.format(self.parameter_name): date_end,
+                })
+            except ValueError:
+                pass
+
+            return queryset
+
+    def choices(self, cl):
+        value = self.value()
+        yield {
+            'parameter_name_start': self.parameter_name_start,
+            'parameter_name_end': self.parameter_name_end,
+            'date_start': value[0],
+            'date_end': value[1],
+            'date_format': date_format_to_human(
+                get_format('DATE_INPUT_FORMATS')[0]
+            )
+        }
+
+
+class TextFilter(BaseCustomFilter):
+
+    """Renders filter form with char field."""
+
+    template = "admin/filters/text_filter.html"
+
+    def queryset(self, request, queryset):
+        if self.value():
+            query = Q()
+            for value in re.split(r'[;|]', self.value()):
+                query |= Q(
+                    **{'{}__icontains'.format(self.parameter_name): value}
+                )
+            return queryset.filter(query)
+
+    def choices(self, cl):
+        return ({
+            'current_value': self.value(),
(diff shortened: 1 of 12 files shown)


src/ralph/admin/filters.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def choices(self, cl):
+        for lookup, title in (
+                ('', _('All')),
+                ('1', _('Yes')),
+                ('0', _('No'))):
+            yield {
+                'selected': self.value() == lookup,
+                'display': title,
+                'value': lookup if not None else '',
```
mkurek: is this working? :) don't think so

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,143 @@
...
+            'value': '',
+            'display': _('All'),
+            'parameter_name': self.parameter_name
+        }
+        for lookup, title in self.choices_list:
+            yield {
+                'selected': smart_text(lookup) == self.value(),
+                'value': lookup,
+                'display': title,
+            }
```
mkurek, line 64: shouldn't parameter_name be specified here too?

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,143 @@
...
+        return [
+            self.used_parameters.get(param)
+            for param in self.expected_parameters()
+        ]
+
+    def queryset(self, request, queryset):
+        if any(self.value()):
+            date = self.value()
+            try:
+                date_start = datetime.strptime(date[0], '%Y-%m-%d')
```
mkurek: date format should depend on l10n

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,143 @@
...
+class TextFilter(BaseCustomFilter):
+
+    """Renders filter form with char field."""
+
+    template = "admin/filters/text_filter.html"
+
+    def queryset(self, request, queryset):
+        if self.value():
+            query = Q()
+            for value in re.findall(r"[\w']+", self.value()):
```
mkurek: why `'` is in regex?

src/ralph/admin/filters.py
szok, line 64: No, because we take only name of the first.

src/ralph/admin/templates/admin/includes/javascripts.html
```diff
@@ -3,27 +3,43 @@
...
 <script>
-  $(document).ready(function(){
-    $('select.filter').on('change', function(data) {
-        window.location.href = $('option:selected', this).val();
-    });
+$(document).ready(function() {
+      $('#filter-button').bind('click', function() {
+          /**
+           * Clone data filters to filtering form (id: filter-form).
+           * Because we can't use the Django form of change_list
```
mkurek: add additional note why we cannot use django form

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,148 @@
...
+class TextFilter(BaseCustomFilter):
+
+    """Renders filter form with char field."""
+
+    template = "admin/filters/text_filter.html"
+
+    def queryset(self, request, queryset):
+        if self.value():
+            query = Q()
+            for value in re.findall(r"[\w]+", self.value()):
```
ar4s: `re.split(';|')` instead of `re.findall`

ar4s:
:ok:

src/ralph/admin/templates/admin/includes/javascripts.html
```diff
@@ -3,27 +3,44 @@
...
-  $(document).ready(function(){
-    $('select.filter').on('change', function(data) {
-        window.location.href = $('option:selected', this).val();
-    });
+$(document).ready(function() {
+      $('#filter-button').bind('click', function() {
+          /**
+           * Clone data filters to filtering form (id: filter-form).
+           * We not used the Django form of change_list because it supports
+           * only the POST request, we need a GET request.
```
vi4m: Explain here _why_ we decided to use get request - avoiding auto refreshing? multiple parameters handling? 
This is horrible hack, and needs good explanation for future generations.

vi4m:
What's the difference between admin and ralph_admin directory? Did we clean this up?

szok:
ralph_admin has been changed to 'admin'

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,148 @@
...
+class BaseCustomFilter(SimpleListFilter):
+
+    def has_output(self):
+        return True
+
+    def lookups(self, request, model_admin):
+        return None
+
+
+class BooleanFilter(BaseCustomFilter):
```
mkurek: couldn't `BooleanFilter` inherit from `ChoicesFilter` and just overwrite `self.choices_list`?

src/ralph/admin/templates/admin/filters/date_filter.html
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+<h6>
+  {% blocktrans with filter_title=title %}
+    {{ filter_title }}
+  {% endblocktrans %}
+</h6>
+{% with choices.0 as i %}
+  <input type="text" name="{{ i.parameter_name_start }}" placeholder="Start YYYY-MM-DD" class="datepicker" value="{{ i.date_start }}" />
```
mkurek: replace `YYYY-MM-DD` with `get_format('DATE_INPUT_FORMATS')[0]` (through some context)

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,148 @@
+# -*- coding: utf-8 -*-
```
mkurek, line 1: I'd love to see some unit tests for those filters :)

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,165 @@
...
+    Example:
+        >>> date_format_to_human('%Y-%m-%d')
+        YYYY-MM-DD
+
+    :param value: Date format example: %Y-%m
+    """
+    maps = {
+        '%Y': 'YYYY',
+        '%m': 'MM',
+        '%d': 'DD',
```
mkurek, line 28: add `'%y': 'YY'` as well

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,165 @@
...
+from datetime import datetime
+
+from django.contrib.admin import SimpleListFilter
+from django.db.models import Q
+from django.utils.encoding import smart_text
+from django.utils.formats import get_format
+from django.utils.translation import ugettext_lazy as _
+
+
+def date_format_to_human(value):
```
mkurek: add lru_cache on this method

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,168 @@
...
+
+class BooleanFilter(ChoicesFilter):
+
+    """Renders filter form with boolean field."""
+
+    def choices(self, cl):
+        for lookup, title in (
+                ('', _('All')),
+                ('1', _('Yes')),
+                ('0', _('No'))):
```
mkurek: `):` should be in new line

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,168 @@
...
+                'value': lookup,
+                'display': title,
+            }
+
+    def queryset(self, request, queryset):
+        if self.value():
+            return queryset.filter(**{self.parameter_name: self.value()})
+
+
+class BooleanFilter(ChoicesFilter):
```
mkurek, line 71: why cannot you just assign 
`
(
  ('', _('All')),
  ('1', _('Yes')),
  ('0', _('No')),
)
`
to `choices_list` as class attribute?

src/ralph/admin/filters.py
```diff
@@ -0,0 +1,172 @@
...
+
+    """Renders filter form with boolean field."""
+
+    choices_list = [
+        ('', _('All')),
+        ('1', _('Yes')),
+        ('0', _('No')),
+    ]
+
+    def choices(self, cl):
```
mkurek: this method isn't necessary anymore, am i right?

src/ralph/admin/tests/tests_filters.py
```diff
@@ -0,0 +1,115 @@
...
+            request=None,
+            params={'status': AssetStatus.in_progress},
+            model=DataCenterAsset,
+            model_admin=DataCenterAssetAdmin,
+        )
+        queryset = choices_filter.queryset(None, DataCenterAsset.objects.all())
+
+        self.assertEqual(0, queryset.count())
+
+    def test_text_filter(self):
```
mkurek, line 95: please add one more test to search by value with separators (which will also test 'containing' - not full match)

src/ralph/admin/tests/tests_filters.py
```diff
@@ -0,0 +1,115 @@
...
+            request=None,
+            params={'barcode': 'barcode_test'},
+            model=DataCenterAsset,
+            model_admin=DataCenterAssetAdmin,
+        )
+        queryset = text_filter.queryset(None, DataCenterAsset.objects.all())
+
+        self.assertEqual(1, queryset.count())
+
+    def test_date_filter(self):
```
mkurek: add two more tests with only `invoice_date_start` and only `invoice_date_end`

src/ralph/admin/filters.py
szok: is needed because BooleanFilter and ChoiceFilter returned are completely different values.

src/ralph/admin/filters.py
szok, line 71: Django create filter objects, not us. 
I can not insert any attribute.

mkurek:
:ok:

ar4s:
:ok:

Issue #1699, opened 2015-08-11
Support assignments


mkurek:
On support page there should be tab with objects for which this support is assigned (with edit option - to assign this support to some objects). It should work analogously to tab in licences (but without quantity).

There should be also tab in `DataCenterAsset` and `BackOfficeAsset` (but in general in every object inherited from `BaseObject`) which will allow to preview and assign support for given object.

PR #1710, opened 2015-08-17
Support assignments

szok:
- Added support and licence to back office asset
- Added support and licence to data center asset

Connected to #1699


src/ralph/back_office/admin.py
```diff
@@ -1,15 +1,48 @@
...
+
+
+class BackOfficeAssetSupport(RalphDetailViewAdmin):
+    icon = 'bookmark'
+    name = 'bo_asset_support'
+    label = _('Supports')
+    url_name = 'back_office_asset_support'
+
+    class BackOfficeAssetSupportInline(RalphTabularInline):
+        model = Support.base_objects.through
```
mkurek: maybe it'll be better to use `BackOfficeAsset.supports.related.through` here? It's the same table - the only change it that in your example we're going to it from support (touching base object) and in mine there is straightforward way from BO asset to supports.

src/ralph/back_office/admin.py
```diff
@@ -1,15 +1,48 @@
...
+    icon = 'bookmark'
+    name = 'bo_asset_licence'
+    label = _('Licence')
+    url_name = 'back_office_asset_licence'
+
+    class BackOfficeAssetLicenceInline(RalphTabularInline):
+        model = BaseObjectLicence
+        raw_id_fields = ('licence',)
+        extra = 1
+        verbose_name = _('Support')
```
mkurek: licence

src/ralph/back_office/admin.py
```diff
@@ -1,15 +1,48 @@
...
+        raw_id_fields = ('support',)
+        extra = 1
+        verbose_name = _('Support')
+
+    inlines = [BackOfficeAssetSupportInline]
+
+
+class BackOfficeAssetLicence(RalphDetailViewAdmin):
+
+    icon = 'bookmark'
```
mkurek: use different icon for supports and licences.

mkurek:
Apply comments from BO to DC too.

ar4s:
:ok:

src/ralph/data_center/admin.py
```diff
@@ -58,6 +58,35 @@ class NetworkView(RalphDetailViewAdmin):
     inlines = [NetworkInline]
 
 
+class DataCenterAssetLicence(RalphDetailViewAdmin):
```
mkurek: use the same order in DC and BO admin modules (here is `DataCenterAssetLicence` then `DataCenterAssetSupport`, in BO it's reversed).

mkurek:
@szok fix my small notice, make it mergable and it's :ok: for me

PR #1716, opened 2015-08-20
Dell servers fix in discovery/http module

jjagodzinski:
Fix discover Dell servers. Not every Dell server has in header server variable.


mkurek:
:ok:

ghost:
:ok:

Issue #1452, opened 2015-05-27
REP: Permissions per object


mkurek:
There should be possibility to specify permissions on object level (a'la Regions in old Ralph).

Requirements:
- permissions should be easily defined (similar form to django-rules)
- permissions should be applied on single object and on list of objects too
- performance - when permissions are checked on list of objects, it should be done as filter on ORM side
- permissions should be checked when using:
  - django admin
  - API
  - importer / exporter
- permissions for related objects should be validated too (ex. if I have rights to save object with selected region)

Example usage of this permissions will be regions mechanism, applied on BO Asset, Licence and Support models (and, of course, user). Only users with selected region could view and change asset/licence/region with the same region. User could only assign region which is assigned to him.

Resources:
https://www.djangopackages.com/grids/g/perms/

mkurek:
There will be 5 PRs for this issue:
- [x] mechanism + regions implementation at django-admin level: #1709 
- [x] implementation of permissions per object at API level (+ simple usage with regions): #1717  
- [ ] implementation of permissions per object at importer/exporter level (+ simple usage with regions): `<work in progress>`
- [x] data importer for regions: #1741
- [x] data exporter for regions (+ minor changes in supports and licences, ex. additional_notes -> remarks): #1738

romcheg:
Closing a stale proposal since we don't plan doing this in the observable future.

PR #1717, opened 2015-08-20
Permissions in API

mkurek:
Connected to #1452 

API base classes/mixins:
- permissions per object (on list, on single objects, for related objects in object details)
- permissions per field (both readonly and change fields)
- custom router (hide models for which user doesn't have permissions)
- `ReversedChoiceField` - better handling `dj.choices.Choices` fields in API
- improved API urls handling

Permissions mechanism fixes:
- allow to specify permissions for M2M fields too
- fixed permissions per object with inheritance (the same check in Q was repeated multiple times, with many tautologies (empty Q())

TODO: 
- [x] move permissions parts from `ralph.api` to `ralph.permissions.api`
- [x] add unit tests
- [x] add docstrings everywhere

### DIFF
--- a/src/ralph/api/tests/_base.py
+++ b/src/ralph/api/tests/_base.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+
+from ralph.tests.models import Foo
+
+
+class APIPermissionsTestMixin(object):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.foo = Foo.objects.create(bar='rab')
+
+        def create_user(name, **kwargs):
+            params = dict(
+                username=name,
+                is_staff=True,
+                is_active=True,
+            )
+            params.update(kwargs)
+            return get_user_model().objects.create(**params)
+        cls.user1 = create_user('user1')
+        cls.user2 = create_user('user2')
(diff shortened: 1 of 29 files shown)


src/ralph/api/tests/_base.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+
+from ralph.tests.models import Foo
+
+
+class APIPermissionsTestMixin(object):
+    def setUp(self):
```
ar4s: convert to `setUpClass(cls)` if it's possible

src/ralph/data_center/urls/api.py
```diff
@@ -1,12 +1,6 @@
...
+from ralph.api import router
 from ralph.data_center.views.api import DataCenterAssetViewSet
 
-router = routers.DefaultRouter()
 router.register(r'data-center-assets', DataCenterAssetViewSet)
-
-urlpatterns = [
-    url(r'^', include(router.urls)),
-]
+urlpatterns = []
```
ar4s, line 6: it is necessary?

mkurek, line 6: yes - see my comment here: https://github.com/allegro/ralph/pull/1717/files#diff-67e5e6c21dea1f8b8f26c4d0d7720774R7

ar4s:
:ok:

szok:
:ok:

Issue #1691, opened 2015-08-11
BaseObject reference in autocomplete widget


mkurek:
When there is reference to BaseObject in autocomplete widget, all values are now displayed as `BaseObject` (each `DataCenterAsset`, `BackOfficeAsset`, `VirtualServer` etc). We should display proper name of object here and allow to search by fields that are enabled to search for given object type (ex. name, barcode, sn, ... for DC Asset; hostname, barcode, sn, imei, ... for BO Asset; name, hypervisor, ... for Virtual etc)

Maybe we should se how it's done in django polymorphic: https://github.com/chrisglass/django_polymorphic

ar4s:
TIP: ``[rel.related_model for rel BaseObject._meta.get_all_related_objects()]``

``https://github.com/hersonls/django-queryset-join/blob/master/src/querysetjoin/__init__.py``

PR #1725, opened 2015-08-25
BaseObject reference in autocomplete widget

szok:
Added:
- Polymorphic library
- **str** method to models
- Polymorphic to BaseObject model

Connected to #1691

### DIFF
--- a/src/ralph/admin/autocomplete.py
+++ b/src/ralph/admin/autocomplete.py
@@ -35,7 +35,11 @@ class SuggestView(JsonViewMixin, View):
         """
         Returns serialized dict as JSON object.
         """
-        can_edit = ralph_site._registry[self.model].has_change_permission(
+        # TODO
+        # Add in the future redirected to the model that is being edited
+        can_edit = not getattr(
+            self.model, 'is_polymorphic', False
+        ) and ralph_site._registry[self.model].has_change_permission(
             self.request
         )
         results = [
@@ -68,16 +72,46 @@ class AjaxAutocompleteMixin(object):
                     return HttpResponseBadRequest()
                 return super().dispatch(request, *args, **kwargs)
 
+            def get_base_ids(self, model, value):
+                """
+                Return IDs for related models.
+                """
+                search_fields = ralph_site._registry[model].search_fields
+                query_filters = [
+                    Q(**{'{}__icontains'.format(field): value})
+                    for field in search_fields
+                ]
+                if issubclass(model, PermissionsForObjectMixin):
+                    queryset = model._get_objects_for_user(
+                        self.request.user, model.objects
+                    )
+                else:
+                    queryset = model.objects
+                return queryset.filter(
+                    *query_filters
+                )[:self.limit].values_list('pk', flat=True)
+
             def get_queryset(self, user):
                 queryset = self.model._default_manager.all()
-                if self.query:
-                    qs = [
-                        Q(**{'{}__icontains'.format(field): self.query})
-                        for field in search_fields
-                    ]
-                    queryset = queryset.filter(reduce(operator.or_, qs))
-                if issubclass(self.model, PermissionsForObjectMixin):
-                    queryset = self.model._get_objects_for_user(user, queryset)
+                if getattr(self.model, 'is_polymorphic', False):
+                    id_list = []
+                    for related_model in self.model._polymorphic_descendants:
+                        id_list.extend(self.get_base_ids(
+                            related_model,
+                            self.query,
+                        ))
+                    queryset = queryset.filter(pk__in=id_list)
+                else:
+                    if self.query:
+                        qs = [
+                            Q(**{'{}__icontains'.format(field): self.query})
+                            for field in search_fields
+                        ]
+                        queryset = queryset.filter(reduce(operator.or_, qs))
+                    if issubclass(self.model, PermissionsForObjectMixin):
+                        queryset = self.model._get_objects_for_user(
+                            user, queryset
+                        )
                 return queryset.order_by(*ordering)[:self.limit]
 
         class Detail(SuggestView):
@@ -90,7 +124,7 @@ class AjaxAutocompleteMixin(object):
                 return super().dispatch(request, *args, **kwargs)
 
             def get_queryset(self, user):
-                queryset = self.model.objects.filter(pk=int(self.pk))
+                queryset = self.model._default_manager.filter(pk=int(self.pk))
                 if issubclass(self.model, PermissionsForObjectMixin):
                     queryset = self.model._get_objects_for_user(user, queryset)
                 if not queryset.exists():
--- a/src/ralph/admin/widgets.py
+++ b/src/ralph/admin/widgets.py
@@ -191,7 +191,7 @@ class AutocompleteWidget(forms.TextInput):
                     pass
         current_object = None
         if value:
-            current_object = self.rel_to.objects.select_related(
+            current_object = self.rel_to._default_manager.select_related(
                 # https://docs.djangoproject.com/en/1.8/ref/models/querysets/#select-related
                 # we cannot pass empty list - this would select all related
                 # model - instead we pass None, which means that none of
@@ -212,14 +212,15 @@ class AutocompleteWidget(forms.TextInput):
         can_edit = self.admin_site._registry[self.rel_to].has_change_permission(
             self.request
         )
-        if value and can_edit:
+        is_polymorphic = getattr(self.rel_to, 'is_polymorphic', False)
+        if value and can_edit and not is_polymorphic:
             context['change_related_template_url'] = self.get_related_url(
                 info, 'change', value,
             )
         can_add = self.admin_site._registry[self.rel_to].has_add_permission(
             self.request
         )
-        if can_add:
+        if not is_polymorphic and can_add:
             context['add_related_url'] = self.get_related_url(info, 'add')
         template = loader.get_template('admin/widgets/autocomplete.html')
         return template.render(context)
--- a/src/ralph/assets/models/base.py
+++ b/src/ralph/assets/models/base.py
@@ -4,14 +4,20 @@ from django.db import models
 from ralph.lib.mixins.models import TimeStampMixin
 from ralph.lib.permissions import PermByFieldMixin
 from ralph.lib.permissions.models import PermissionsBase
+from ralph.lib.polymorphic.models import Polymorphic, PolymorphicBase
 from ralph.lib.transitions import TransitionWorkflowBase
 
 BaseObjectMeta = type(
-    'BaseObjectMeta', (PermissionsBase, TransitionWorkflowBase), {}
+    'BaseObjectMeta', (
+        PolymorphicBase,
+        PermissionsBase,
+        TransitionWorkflowBase
+    ), {}
 )
 
 
 class BaseObject(
+    Polymorphic,
     PermByFieldMixin,
     TimeStampMixin,
     models.Model,
--- a/src/ralph/data_center/models/virtual.py
+++ b/src/ralph/data_center/models/virtual.py
@@ -9,11 +9,17 @@ class Database(BaseObject):
         verbose_name = _('database')
         verbose_name_plural = _('databases')
 
+    def __str__(self):
+        return 'Database: {}'.format(self.service_env)
+
 
 class VIP(BaseObject):
     class Meta:
         verbose_name = _('VIP')
         verbose_name_plural = _('VIPs')
+
+    def __str__(self):
+        return 'VIP: {}'.format(self.service_env)
 
 
 class VirtualServer(BaseObject):
@@ -21,6 +27,11 @@ class VirtualServer(BaseObject):
         verbose_name = _('Virtual server (VM)')
         verbose_name_plural = _('Virtual servers (VM)')
 
+    def __str__(self):
+        return 'VirtualServer: {}'.format(self.service_env)
+
 
 class CloudProject(BaseObject):
-    pass
+
+    def __str__(self):
+        return 'CloudProject: {}'.format(self.service_env)
--- a/src/ralph/lib/polymorphic/models.py
+++ b/src/ralph/lib/polymorphic/models.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+"""
+Polymorphic using inherited models in Django.
+Displays information in queryset list about the model that inherits
+the polymorphic model
+
+Example:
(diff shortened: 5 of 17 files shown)


src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,101 @@
...
+            result.append((
+                obj.content_type_id, obj.pk)
+            )
+        result = sorted(result, key=lambda x: x[0])
+        result = groupby(result, lambda x: x[0])
+
+        result_query = []
+        for k, v in result:
+            model = ct_map[k]
+            if model.__name__ not in 'BaseObject':
```
mkurek: since it's in lib it shouldn't reference to ralph objects, especially hardcoded

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,101 @@
...
+    def get_base_ids(model, value):
+        """
+        Return BaseObject ID for related models.
+        """
+        search_fields = ralph_site._registry[model].search_fields
+        qs = [
+            models.Q(**{'{}__icontains'.format(field): value})
+            for field in search_fields
+        ]
+        return model.objects.filter(*qs)[:10].values_list('pk', flat=True)
```
mkurek: count should be parametrized

src/ralph/assets/models/base.py
```diff
@@ -2,9 +2,9 @@
 from django.db import models
 
 from ralph.lib.mixins.models import TimeStampMixin
-from ralph.lib.permissions import PermByFieldMixin
+from ralph.lib.permissions import PermByFieldMixin, TransitionWorkflowBase
```
ar4s: why `TransitionWorkflowBase` is here?

szok: My mistake in rebase

src/ralph/data_center/models/virtual.py
```diff
@@ -9,18 +9,29 @@ class Meta:
...
+        return 'Database: {}'.format(self.service_env)
+
 
 class VIP(BaseObject):
     class Meta:
         verbose_name = _('VIP')
         verbose_name_plural = _('VIPs')
 
+    def __str__(self):
+        return 'Database: {}'.format(self.service_env)
```
ar4s: :see_no_evil:

src/ralph/data_center/models/virtual.py
```diff
@@ -9,18 +9,29 @@ class Meta:
...
+
 
 class VirtualServer(BaseObject):
     class Meta:
         verbose_name = _('Virtual server (VM)')
         verbose_name_plural = _('Virtual servers (VM)')
 
+    def __str__(self):
+        return 'Database: {}'.format(self.service_env)
```
ar4s: :see_no_evil:

src/ralph/data_center/models/virtual.py
```diff
@@ -9,18 +9,29 @@ class Meta:
...
+
 
 class CloudProject(BaseObject):
-    pass
+
+    def __str__(self):
+        return 'Database: {}'.format(self.service_env)
```
ar4s: :see_no_evil:

src/ralph/admin/autocomplete.py
```diff
@@ -35,7 +36,9 @@ def get(self, request, *args, **kwargs):
         """
         Returns serialized dict as JSON object.
         """
-        can_edit = ralph_site._registry[self.model].has_change_permission(
+        can_edit = not issubclass(
```
mkurek: it's not good condition IMO - sometimes we'll point in autocomplete to concrete subclass of Polimorphic (BaseObject), ex. Database will have reference to physical server (DatacenterAsset) on which it's stored (it's example, of course), and with your condition we won't be able to edit DatacenterAsset from Database view.

This condition should eliminate only `BaseObject` and `Asset` in our case - so maybe add some attribute to those classes and use getattr here?

src/ralph/admin/widgets.py
```diff
@@ -212,14 +213,14 @@ def render(self, name, value, attrs=None):
...
-        if value and can_edit:
+        if value and can_edit and not issubclass(self.rel_to, Polymorphic):
             context['change_related_template_url'] = self.get_related_url(
                 info, 'change', value,
             )
         can_add = self.admin_site._registry[self.rel_to].has_add_permission(
             self.request
         )
-        if can_add:
+        if not issubclass(self.rel_to, Polymorphic) and can_add:
```
mkurek: check my previous comment above - we should somehow distinguish final classes (`DataCenterAsset`, `VIP`) etc from base classes (`BaseObject`, `Asset`)

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,103 @@
...
+    def save(self, *args, **kwargs):
+        """
+        Save object content type
+        """
+        if not self.content_type:
+            self.content_type = ContentType.objects.get_for_model(self)
+        super().save(*args, **kwargs)
+
+    @classmethod
+    def get_related_models(cls):
```
mkurek: `get_descendants_models` suits better here

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,103 @@
...
+            ]
+
+        result = []
+        for related in get_models(cls):
+            result.extend(get_models(related))
+            result.append(related)
+        return result
+
+    @staticmethod
+    def get_base_ids(model, value, limit=10):
```
mkurek: this is used only in autocomplete and it should be moved there (in lib there should be only generic code, easily reusable in other applications).

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,103 @@
...
+            result.append((
+                obj.content_type_id, obj.pk)
+            )
+        result = sorted(result, key=lambda x: x[0])
+        result = groupby(result, lambda x: x[0])
+
+        result_query = []
+        for k, v in result:
+            model = ct_map[k]
+            polymorphic_model = getattr(model, 'polymorphic_model')
```
mkurek: the same case as above - we should somehow (maybe from `__mro__`?) deduce what are base classes (classes, which are inherited) and what are final classes (classes, which are not inherited anymore).

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def iterator(self):
+        """
+        Override iterator:
+            - Iterate for all objects and collected ID
+            - For each ContentType generates additional queryset
+            - Returns iterator with different models
+        """
+        result = []
+        ct_map = {}
```
mkurek: rename it to `content_type_model_map`

src/ralph/admin/autocomplete.py
```diff
@@ -70,12 +73,21 @@ def dispatch(self, request, *args, **kwargs):
...
-                    ]
-                    queryset = queryset.filter(reduce(operator.or_, qs))
+                if issubclass(self.model, Polymorphic):
+                    bs_ids = []
+                    for related_model in self.model.get_related_models():
+                        bs_ids.extend(self.model.get_base_ids(
+                            related_model,
+                            self.query
+                        ))
+                    queryset = queryset.filter(pk__in=bs_ids)
```
mkurek: this conflicts a little bit with permissions per object.The simplest way IMO will be to add some logic to `get_base_ids` method (since i suggested to move it here, it will not conflict anyway with Polymorphic), pass there user and check there if related model is subclass of PermissionsForObjectMixin - if yes, then `_get_objects_for_user` should be called

src/ralph/admin/autocomplete.py
```diff
@@ -35,7 +35,9 @@ def get(self, request, *args, **kwargs):
         """
         Returns serialized dict as JSON object.
         """
-        can_edit = ralph_site._registry[self.model].has_change_permission(
+        can_edit = not getattr(
```
mkurek, line 40: use 3-argument getattr (this will raise `AttributeError` when model has no attribute)

src/ralph/admin/autocomplete.py
```diff
@@ -68,16 +70,46 @@ def dispatch(self, request, *args, **kwargs):
                     return HttpResponseBadRequest()
                 return super().dispatch(request, *args, **kwargs)
 
+            def get_base_ids(self, model, value):
+                """
+                Return BaseObject ID for related models.
```
mkurek: don't use BaseObject here - it should be generic

src/ralph/admin/autocomplete.py
```diff
@@ -68,16 +70,46 @@ def dispatch(self, request, *args, **kwargs):
...
-                if self.query:
-                    qs = [
-                        Q(**{'{}__icontains'.format(field): self.query})
-                        for field in search_fields
-                    ]
-                    queryset = queryset.filter(reduce(operator.or_, qs))
-                if issubclass(self.model, PermissionsForObjectMixin):
-                    queryset = self.model._get_objects_for_user(user, queryset)
+                if getattr(self.model, '_polymorphic_models', None):
+                    bs_ids = []
```
mkurek: use more descriptive name

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
```
mkurek, line 1: add some docstring for this module with in-depth description what is going one here, why we need this mixin etc

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,102 @@
...
+
+class PolymorphicBase(models.base.ModelBase):
+    def __new__(cls, name, bases, attrs):
+        full_mro = set(
+            tuple([mro for b in bases for mro in b.__mro__]) + bases
+        )
+        base_polymorphic = set(
+            [b for b in full_mro if issubclass(b, Polymorphic)]
+        )
+        attrs['_polymorphic_models'] = base_polymorphic
```
mkurek, line 73: i'd mark here every class from `base_polymorphic` that polymorphic query should be used for it - for now it should only apply to `BaseObject` and `Asset`. Then you could use this param instead of checking `_polymorphic_models` in autocomplete.

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,102 @@
...
+        Save object content type
+        """
+        if not self.content_type:
+            self.content_type = ContentType.objects.get_for_model(self)
+        super().save(*args, **kwargs)
+
+    @classmethod
+    def get_descendants_models(cls):
+        """
+        Return related models.
```
mkurek: Elaborate on this. What related models? Only polymorphic?

src/ralph/assets/models/base.py
```diff
@@ -23,3 +33,6 @@ class BaseObject(
     parent = models.ForeignKey('self', null=True, blank=True)
     remarks = models.TextField(blank=True)
     service_env = models.ForeignKey('ServiceEnvironment', null=True)
+
+    polymorphic_objects = PolymorphicQuerySet.as_manager()
```
mkurek: this could be defined in `Polymorphic`

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,102 @@
...
+        result = []
+        content_type_model_map = {}
+        for obj in super().iterator():
+            content_type_model_map[
+                obj.content_type_id
+            ] = obj.content_type.model_class()
+            result.append((
+                obj.content_type_id, obj.pk)
+            )
+        result = sorted(result, key=lambda x: x[0])
```
mkurek: Could you try to maintain original order?

src/ralph/tests/models.py
```diff
@@ -45,3 +50,22 @@ def pack(self, **kwargs):
     @transition_action
     def go_to_post_office(self, **kwargs):
         pass
+
+
+class PolymorphicModelBaseTest(
```
mkurek: this should be in `ralph.lib.polymorphic.tests.models` - we don't want to use ralph (test) models (or more general anything from ralph) in lib

src/ralph/admin/autocomplete.py
mkurek, line 40: please add also TODO comment that for polymorphic models edit should be enabled, but should redirect to editing concrete (final) object.

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,118 @@
...
+    def save(self, *args, **kwargs):
+        """
+        Save object content type
+        """
+        if not self.content_type:
+            self.content_type = ContentType.objects.get_for_model(self)
+        super().save(*args, **kwargs)
+
+    @classmethod
+    def get_descendants_models(cls):
```
mkurek: in my opinion we should do this only once per model and it should be done in `PolymorphicBase` (when you're iterating over `base_polymorphic` you could store additional variable in each of this classes list, to which you will append currently created class), ex:

```
attrs['_polymorphic_descendats'] = []
new_class = super().__new__(cls, name, bases, attrs)  # notice calling super before loop
for polymorphic_class in base_polymorphic:
    polymorphic_class._is_polymorphic = True  # notice _ here at the beginning
    polymorphic_class._polymorphic_descendats.append(new_class)
return new_class
```

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,102 @@
...
+            if polymorphic_models and model not in polymorphic_models:
+                result_query.extend(
+                    model.objects.filter(pk__in=[i[1] for i in v])
+                )
+
+        for obj in result_query:
+            yield obj
+
+
+class PolymorphicBase(models.base.ModelBase):
```
mkurek, line 54: please add some comment what is going on here (how newly created class is affected and base classes)

mkurek:
:ok:

src/ralph/lib/polymorphic/models.py
```diff
@@ -0,0 +1,112 @@
...
+        for k, v in result:
+            model = content_type_model_map[k]
+            polymorphic_models = getattr(model, '_polymorphic_models', [])
+            if polymorphic_models and model not in polymorphic_models:
+                result_query.extend(
+                    model.objects.filter(pk__in=[i[1] for i in v])
+                )
+
+        for obj in result_query:
+            yield obj
```
ghost, line 51: Name is `iterator` and `yield` makes it generator. Shouldn't be simple return?

ar4s:
:ok:

PR #1726, opened 2015-08-25
Fixed qunit test with PhantomJS

ar4s:
Now it works!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 sudo: false
 language: python
 
+cache:
+  directories:
+    - node_modules
+    - bower_components
+
 python:
     - "3.4"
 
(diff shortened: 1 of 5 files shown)


szok:
:ok:

.travis.yml
```diff
@@ -9,6 +14,7 @@ before_install:
     - make flake
 
 install:
+    - npm config set python /usr/bin/python2.7
```
mkurek, line 17: why can't we use python3?

ar4s, line 17: because some dependency (`node-gyp`) work only with Python 2.x., JS's tests and Python's tests are separeted from each other

mkurek:
:ok:

Issue #1516, opened 2015-06-23
BO models REST API


mkurek:
Working REST API for BO models (create, get, list, update):
- [ ] Warehouse
- [ ] BackOfficeAsset

...

PR #1727, opened 2015-08-25
BO REST API

mkurek:
Connected to #1516 

Added REST API resources for:
- accounts app (`RalphUser`, `Region`)
- assets app (`Environment`, `Service`, `ServiceEnvironment`, `Manufacturer`, `Category`, `AssetModel`)
- back office (`BackOfficeAsset`, `Warehouse`)
- supports (`Support`, `SupportType`)

Additional API improvements:
- new Permission `IsSuperuserOrReadonly` - only superuser could save this resource
- fixed permission per fields in API when fields were defined excplicitly

Added additional models/fields:
- `BusinessSegment` model
- `ProfitCenter` model
- `business_owners` and `technical_owners` for service
- changed service `profit_center` from char field to foreign key to `ProfitCenter` table

### DIFF
--- a/src/ralph/accounts/api.py
+++ b/src/ralph/accounts/api.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+from django.contrib.auth import get_user_model
+
+from ralph.accounts.models import Region
+from ralph.api import RalphAPISerializer, RalphAPIViewSet, router
+from ralph.api.permissions import IsSuperuserOrReadonly
+
+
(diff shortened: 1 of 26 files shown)


src/ralph/accounts/api.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from ralph.accounts.models import RalphUser, Region
```
szok: maybe use get_user_model

szok:
:ok:

ghost:
:ok:

PR #1732, opened 2015-08-26
API improvements

mkurek:
Various Ralph API improvements:
- improved serialization for save actions (POST, PUT, PATCH) - from now, there will be default serializer created for save actions, in which every relation will be serialized using `PrimaryKeyRelatedField` (so user should specify object PK in relation field). This serializer could be overwritten using `save_serializer_class` attribute in `RalphAPIViewSet`
- serializer now returns both object PK and its url (see `get_default_field_names` for details)
- improved passing context (request and user) to subserializers - from now every serializer (even instantiated directly as a field in other serializer) will have current request and user set (see `assign_request_and_user` and it's usages for details).
- introduced `QuerysetRelatedMixin` which can easily handle `select_related` and `prefetch_related` on queryset. By default `select_related` is fill with related admin site `list_select_related` value
- new `AdminSearchFieldsMixin` thanks to which search fields will be by default taken from related admin site.

TODO:
- [x] Ralph API guideline (both for user and developer)
- [x] unit tests for new features

### DIFF
--- a/docs/development/api.md
+++ b/docs/development/api.md
@@ -1,0 +1,45 @@
+# Ralph API resources
+
+## How to create new API resource based on Django model
+
+```django
+# model
+from django.db import models
+
+class Foo(model.Model):
+    bar = models.CharField(max_length=10)
+
+
+# API
+from ralph.api import RalphAPISerializer, RalphAPIViewSet, router
+
+
+class FooSerializer(RalphAPISerializer):
+    class Meta:
+        model = Foo
+
+
+class FooViewSet(RalphAPIViewSet):
+    queryset = Foo.objects.all()
+    serializer_class = FooSerializer
+
+
+# use Ralph router instance to register new viewset
+router.register(r'foos', FooViewSet)
+```
+
+## What are built-in features of Ralph API classes?
+
+### `RalphAPISerializer`
+* include both object PK and its url (see `ralph.api.serializers.RalphAPISerializer.get_default_field_names` for details).
+* every sub-serializer (even instantiated directly as a field in other serializer) have current context set.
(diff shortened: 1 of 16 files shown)


src/ralph/api/viewsets.py
```diff
@@ -25,6 +67,37 @@ def __init__(self, *args, **kwargs):
...
+            return type(
+                '{}SaveSerializer'.format(Meta.model.__name__),
+                (serializers.ModelSerializer,),
+                {
+                    'Meta': Meta,
+                    'serializer_choice_field': ReversedChoiceField,
+                    'serializer_related_field': relations.PrimaryKeyRelatedField
+                }
+            )
+            return self.save_serializer_class
```
mkurek: remove this line

docs/development/api.md
```diff
@@ -0,0 +1 @@
+# TODO
```
mkurek: add this file to mkdocs.yml

src/ralph/api/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        queryset = super().get_queryset()
+        admin_site = ralph_site._registry.get(queryset.model)
+        if self.select_related:
+            queryset = queryset.select_related(*self.select_related)
+        if self.prefetch_related:
+            queryset = queryset.prefetch_related(*self.prefetch_related)
+        if (
+            admin_site and
+            not self._skip_admin_list_select_related
+            and admin_site.list_select_related
```
ar4s: move `and` one line up at the end

ar4s:
:ok:

andrzej-jankowski:
:ok:

PR #1738, opened 2015-08-28
merged migrations

mkurek:
The best option is to remove whole database and run this migration and run export/import from Ralph 2.0 again.

If you want to maintain data before this squash (not recommended), checkout to commit before this, run `ralph migrate`, then for each app in which there are migrations run `ralph migrate <app_name> 0001 --fake`, then checkout to this commit.


szok:
:ok:

ar4s:
:ok:

vi4m:
@mkurek you forgot about the beta testers - we should _at least_ put some warning how to cope with broken migrations.

Issue #1694, opened 2015-08-11
Transitions actions from Ralph 2.0


mkurek:
Copy (and adjust) all transitions actions from Ralph 2.0
- [x] assign user (with additional form field to select user)
- [x] assign owner (with additional form field to select owner)
- [x] unassign (current) user
- [x] unassign (current) owner
- [x] assign loan end date (with additional form field to type date)
- [x] unassign loan end date
- [x] assign warehouse (with additional form field to select warehouse)
- [x] unassign licences (clear all licenses assigned to objects)
- [x] change hostname (with additional field to select country (region?))

Ralph 2.0 transitions: https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/views/transition.py

PR #1748, opened 2015-08-31
Transitions actions from Ralph 2.0

szok:
Copy transitions from old Ralph 2.0
- assign user
- assign owner
- unassign user
- unassign owner
- assign loan end date
- assign warehouse
- unassign licences
- change hostname

Connected to #1694

### DIFF
--- a/src/ralph/back_office/models.py
+++ b/src/ralph/back_office/models.py
@@ -1,14 +1,20 @@
 # -*- coding: utf-8 -*-
 
+from dj.choices import Country
+from django import forms
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.accounts.models import Regionalizable
+from ralph.assets.country_utils import iso2_to_iso3
 from ralph.assets.models.assets import Asset
 from ralph.assets.models.choices import AssetStatus
 from ralph.lib.mixins.models import NamedMixin, TimeStampMixin
+from ralph.lib.transitions.decorators import transition_action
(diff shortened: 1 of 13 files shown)


ar4s:
:ok:

ghost:
:ok:

src/ralph/back_office/models.py
```diff
@@ -41,7 +47,92 @@ class Meta:
 
     @property
     def country_code(self):
-        return 'PL'
+        if self.owner:
+            iso2 = Country.name_from_id(int(self.owner.country)).upper()
+            return iso2_to_iso3(iso2)
+        return 'POL'
```
mkurek, line 53: i know it's already merged but:
- why country code couldn't be null/None? 
- what if owner doesn't have country specified? you'll probably get some ValueError here (IMO we should return None then)

src/ralph/back_office/models.py
```diff
@@ -41,7 +47,92 @@ class Meta:
...
 
     def __str__(self):
         return 'BackOfficeAsset: {}'.format(self.hostname)
+
+    @transition_action
+    def assign_user(self, **kwargs):
+        self.user = get_user_model().objects.get(pk=int(kwargs['user']))
```
mkurek, line 60: it should be enough to only assign user id here (we don't need to get user instance here and db will validate foreign keys on it's own)

src/ralph/back_office/models.py
szok, line 53: But country field is a required field, so there's always some country will be assigned.
Default country is: pl

PR #1750, opened 2015-09-01
Moving cmdgen.CommandGenerator() inside loop

jjagodzinski:
cmdgen.CommandGenerator() is moved inside 'for' loop in discovery/snmp.
Without this change loop is useless because priv_protocol is not switching
to next one. It uses first priv_protocol from the list 4 times.


ar4s:
:ok:

ghost:
:ok:

PR #1755, opened 2015-09-03
Modify catalyst and ssg plugins to prevent hang when scanning host

jjagodzinski:
ssh_ssg and ssh_cisco_catalyst hangs when scanned IP is not assigned to Cisco
device. This fix prevent running plugin when IP belongs to other device then
Cisco.

### DIFF
--- a/src/ralph/scan/plugins/ssh_ssg.py
+++ b/src/ralph/scan/plugins/ssh_ssg.py
@@ -100,8 +100,8 @@ def _ssh_ssg(ip_address, user, password):
 
 
 def scan_address(ip_address, **kwargs):
-    if kwargs.get('http_family') not in ('SSG', 'Unspecified'):
-        raise NoMatchError("It's not a Juniper SSG.")
+    if 'ssg' not in kwargs.get('http_family', '').lower():
+        raise NoMatchError("Incompatible device found.")
     if 'nx-os' in (kwargs.get('snmp_name', '') or '').lower():
         raise NoMatchError("Incompatible Nexus found.")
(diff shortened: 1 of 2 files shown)


src/ralph/scan/plugins/ssh_ssg.py
```diff
@@ -100,7 +100,7 @@ def _ssh_ssg(ip_address, user, password):
 
 
 def scan_address(ip_address, **kwargs):
-    if kwargs.get('http_family') not in ('SSG', 'Unspecified'):
+    if kwargs.get('http_family') not in ('SSG'):
```
mkurek: this will not work properly - it's not a tuple

src/ralph/scan/plugins/ssh_ssg.py
```diff
@@ -100,8 +100,8 @@ def _ssh_ssg(ip_address, user, password):
 
 
 def scan_address(ip_address, **kwargs):
-    if kwargs.get('http_family') not in ('SSG', 'Unspecified'):
```
mkurek, line 103: what will now happen with devices with http_family unspecified?

jjagodzinski, line 103: Plugin will not run. I believe that autoscan should detect if device is an SSG and fill http_family parameter. If 'Unspecified' stays and you scan device (with http_family='Unspecified') then plugin can freeze all scanning process.

mkurek:
:ok:

ar4s:
:ok:

PR #1679, opened 2015-08-06
[wip] Added bulk add to DC & BO Assets.

ghost:
Added new field MultilineField which handles seperated values.
Added MultivalueFormMixin which validates many MultilineField in one form.

connected to: #1606

PR #1761, opened 2015-09-03
Mass adding asset

szok:
Added multi adding models in admin page.

Connected to #1606

### DIFF
--- a/src/ralph/admin/fields.py
+++ b/src/ralph/admin/fields.py
@@ -1,0 +1,154 @@
+# -*- coding: utf-8 -*-
+import re
+
+from django import forms
+from django.forms import ValidationError
+from django.forms.util import ErrorList
+from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext as _
+
+
+class MultilineField(forms.CharField):
+    """
+    This field is a textarea which treats its content as many values seperated
+    by `separators`
+    Validation:
+        - separated values cannot duplicate each other,
+    """
+    widget = forms.Textarea
+    separators = ",|\n"
+
+    def __init__(self, allow_duplicates=True, *args, **kwargs):
+        self.allow_duplicates = allow_duplicates
+        super().__init__(*args, **kwargs)
+
+    def validate(self, values):
+        if not values and self.required:
+            error_msg = _(
+                "Field can't be empty. Please put the item OR items separated "
+                "by new line or comma."
+            )
+            raise forms.ValidationError(error_msg, code='required')
+        non_empty_values = [item for item in values if len(item.strip()) > 0]
+        if not self.allow_duplicates:
+            has_duplicates = len(set(non_empty_values)) != len(non_empty_values)
+            if has_duplicates:
+                raise forms.ValidationError(_("There are duplicates in field."))
+
+    def to_python(self, value):
+        items = []
+        if value:
+            for item in re.split(self.separators, value):
+                items.append(item.strip(' \t\n\r'))
+
+        return items
+
+
+class MultivalueFormMixin(object):
+    """A form that has several multiline fields that need to have the
+    same number of entries.
+
+    :param multivalue_fields: list of form fields which require the same item
+    count
+    :param one_of_mulitvalue_required: list of form fields names which must
+    be provided to form, e.g.
+        one_of_mulitvalue_required = ['sn', 'barcode']
+        means that if user inputs such data:
+        sn, barcode:
+        1,1, # valid
+        2,   # valid
+        ,2   # valid
+        ,    # invalid, because neither sn nor barcode was provided
+    """
+    multivalue_fields = []
+    one_of_mulitvalue_required = []
+    model = None
+
+    def equal_count_validator(self, cleaned_data):
+        """Adds a validation error if form's multivalues fields have
+        different count of items."""
+        items_count_per_multi = set()
+        for field in self.multivalue_fields:
+            if cleaned_data.get(field, []):
+                items_count_per_multi.add(len(cleaned_data.get(field, [])))
+        if len(items_count_per_multi) > 1:
+            for field in self.multivalue_fields:
+                if field in cleaned_data:
+                    msg = _((
+                        "Fields: %(fields)s "
+                        "- require the same number of items")
+                    ) % {
+                        'fields': ', '.join(self.multivalue_fields)
+                    }
+                    self.errors.setdefault(field, []).append(msg)
+
+    def any_in_multivalues_validator(self, data):
+        """
+        Checks if each row has filled at least one field specified
+        by one_of_mulitvalue_required.
+        """
+        def rows_of_required():
+            rows_of_required = [
+                self.cleaned_data[field_name] for field_name in
+                self.one_of_mulitvalue_required if field_name in
+                self.cleaned_data
+            ]
+            for multivalues_row in zip(*rows_of_required):
+                yield multivalues_row
+
+        if self.one_of_mulitvalue_required:
+            for row_of_required in rows_of_required():
+                if not any(row_of_required):
+                    for field_name in self.one_of_mulitvalue_required:
+                        errors = self._errors.setdefault(
+                            field_name, ErrorList()
+                        )
+                        msg = _('Fill at least on of %(v)s in each row') % {
+                            'v': ','.join(self.one_of_mulitvalue_required)
+                        }
+                        errors.append(_(msg))
+                    break
+
+    def check_field_uniqueness(self, model, field_name, values):
+        """
+        Check field (pointed by *self.db_field_path*) uniqueness.
+        If duplicated value is found then raise ValidationError
+
+        :param string Model: model field to be unique (as a string)
+        :param list values: list of field values
+        """
+        if not values:
+            return
+        conditions = {'{}__in'.format(field_name): values}
+        objs = model.objects.filter(**conditions)
+        if objs:
+            if hasattr(model, 'get_absolute_url'):
+                url = '<a href="{}">{}</a>'
+                comma_items = ', '.join([
+                    url.format(obj.get_absolute_url(), obj.id)
+                    for obj in objs
+                ])
+            else:
+                comma_items = ', '.join([str(obj) for obj in objs])
+            msg = _('Following items already exist: ') + comma_items
+            raise ValidationError(mark_safe(msg))
+
+    def check_uniqness(self, data):
+        for field_name in self.multivalue_fields:
+            field = self[field_name].field
+            if field.allow_duplicates:
+                continue
+            try:
+                self.check_field_uniqueness(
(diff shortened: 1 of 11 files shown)


src/ralph/admin/templates/admin/submit_line.html
```diff
@@ -10,6 +11,13 @@
     <div class="small-{% if show_delete_link %}11{% else %}12{% endif %} columns text_right right_section">
       {% block additional_buttons %}{% endblock %}
       <ul class="button-group radius right">
+        {% if multi_add_url %}
+          <li>
+              <a href="{{ multi_add_url }}" class="default button small info">
+                Bulk Add
```
ar4s: `trans`

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def validate(self, values):
+        if not values and self.required:
+            error_msg = _(
+                "Field can't be empty. Please put the item OR items separated "
+                "by new line or comma."
+            )
+            raise forms.ValidationError(error_msg, code='required')
+        if not self.allow_duplicates:
+            has_duplicates = len(set(values)) != len(values)
```
ghost: This won't work for case when values include blank lines. ;)

src/ralph/admin/views/multiadd.py
```diff
@@ -0,0 +1,126 @@
...
+
+            try:
+                self.obj.save()
+            except IntegrityError as e:
+                print(forms.forms.NON_FIELD_ERRORS)
+                form._errors[forms.forms.NON_FIELD_ERRORS].append(str(e))
+                return self.form_invalid(form)
+
+        messages.success(
+            self.request, _('Saved many assets successfully with no fuss')
```
ghost: "Saved {saved_assets} assets" will be better (I know I wrote it, but it was PoC ;))

ghost:
https://github.com/allegro/ralph/pull/1679 here you have tests which was not included because it was PoC

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+    count
+    :param one_of_mulitvalue_required: list of form fields names which must
+    be provided to form, e.g.
+        one_of_mulitvalue_required = ['sn', 'barcode']
+        means that if user inputs such data:
+        sn, barcode:
+        1,1, # valid
+        2,   # valid
+        ,2   # valid
+        ,    # invalid, because `sn` neither `barcode` was provided
```
mkurek: neither sn nor barcode

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+        """Adds a validation error if form's multivalues fields have
+        different count of items."""
+        items_count_per_multi = set()
+        for field in self.multivalue_fields:
+            if cleaned_data.get(field, []):
+                items_count_per_multi.add(len(cleaned_data.get(field, [])))
+        if len(items_count_per_multi) > 1:
+            for field in self.multivalue_fields:
+                if field in cleaned_data:
+                    msg = _("Fields: %(fields)s - require the same count") % {
```
mkurek: the same number of items

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+                        'fields': ', '.join(self.multivalue_fields)
+                    }
+                    self.errors.setdefault(field, []).append(msg)
+
+    def any_in_multivalues_validator(self, data):
```
mkurek, line 85: add some docstring here

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+            for row_of_required in rows_of_required():
+                if not any(row_of_required):
+                    for field_name in self.one_of_mulitvalue_required:
+                        errors = self._errors.setdefault(
+                            field_name, ErrorList()
+                        )
+                        msg = _('Fill at least on of %(v)s in each row') % {
+                            'v': ','.join(self.one_of_mulitvalue_required)
+                        }
+                        errors.append(_(msg))
```
mkurek, line 109: you could break here to not get the same error more than once

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+        If duplicated value is found then raise ValidationError
+
+        :param string Model: model field to be unique (as a string)
+        :param list values: list of field values
+        """
+        if not values:
+            return
+        conditions = {'{}__in'.format(field_name): values}
+        objs = model.objects.filter(**conditions)
+        if objs:
```
mkurek, line 124: use `exists` here (https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.exists)

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class MultilineField(forms.CharField):
+    """
+    This field is a textarea which treats its content as many values seperated
+    by `separators`
+    Validation:
+        - separated values cannot duplicate each other,
+    """
+    allow_duplicates = True
```
mkurek: this should be False by default IMO, @ar4s @xliiv what do you think?

src/ralph/admin/views/multiadd.py
```diff
@@ -0,0 +1,146 @@
...
+        params = self.model._meta.app_label, self.model._meta.model_name
+        url = 'admin:{}_{}_changelist'.format(*params)
+        return reverse(url)
+
+    @transaction.atomic
+    def form_valid(self, form):
+        saved_assets = []
+        args = [form.cleaned_data[field] for field in self.fields]
+        for data in zip(*args):
+            self.obj.baseobject_ptr_id = None
```
mkurek: as far as i understand, this view should be generic (easy to use for any model) - if yes, then we shouldn't rely here on asset/baseobject etc - find out if we can get inheritance hierarchy (and id's pointers) from meta or sth like that

src/ralph/admin/fields.py
ghost: +1 to `False`

src/ralph/admin/fields.py
szok, line 124: I can not because then for each object generates URLs

src/ralph/admin/fields.py
ghost, line 109: I think the current version is better, because error message is showed under each field which validation is applied to.

src/ralph/back_office/admin.py
```diff
@@ -79,6 +81,9 @@ class BackOfficeAssetAdmin(
         'status', 'barcode', 'purchase_order', 'model', 'user', 'warehouse',
         'sn', 'hostname', 'invoice_date', 'invoice_no', 'region',
     ]
+    multiadd_fields = ['sn', 'barcode']
```
mkurek: what about imei? as far as i remember there were imei field too when specific category was chosen for bo asset. Maybe you could call `get_multiadd_fields` method, to which new object will be passed and by default it'll return value of `multiadd_fields` attribute, but you could customize it for BO Asset and depending on asset category add (or not) imei to multiadd fields. @ar4s @xliiv what do you think?

src/ralph/admin/fields.py
mkurek, line 124: ok, never mind

src/ralph/admin/fields.py
mkurek, line 109: i mean the outer for loop

ghost, line 109: Ok, then +1

src/ralph/back_office/admin.py
ghost: +1 to imei flashback and solution :)

szok: We have not imei field, which model to add?

mkurek: true. So:
- leave it as it is right now
- in the next PR add additional BO Asset fields (https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/models_assets.py#L1062) and add imei to multiadd_fields only if condition is fulfilled (https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/forms.py#L750)

src/ralph/admin/views/multiadd.py
```diff
@@ -0,0 +1,148 @@
...
+        params = self.model._meta.app_label, self.model._meta.model_name
+        url = 'admin:{}_{}_changelist'.format(*params)
+        return reverse(url)
+
+    @transaction.atomic
+    def form_valid(self, form):
+        saved_assets = []
+        args = [form.cleaned_data[field] for field in self.fields]
+        for data in zip(*args):
+            for model in self.obj._meta.parents:
```
mkurek: this won't work - `_meta.parents` return only direct parent, not all ancestors. Use sth like this:

```
def _get_ancestors_pointers(model):
    result = []
    for parent_model, parent_field in model._meta.parents.items():
        result.append(parent_field.column)
        result.extend(_get_ancestors_pointers(parent_model))
    return result
```

then set all those fields to `None`

mkurek:
:ok:

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,154 @@
...
+    def validate(self, values):
+        if not values and self.required:
+            error_msg = _(
+                "Field can't be empty. Please put the item OR items separated "
+                "by new line or comma."
+            )
+            raise forms.ValidationError(error_msg, code='required')
+        non_empty_values = [item for item in values if len(item.strip()) > 0]
+        if not self.allow_duplicates:
+            has_duplicates = len(set(non_empty_values)) != len(non_empty_values)
```
ghost, line 34: :+1:

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,154 @@
...
+                        "Fields: %(fields)s "
+                        "- require the same number of items")
+                    ) % {
+                        'fields': ', '.join(self.multivalue_fields)
+                    }
+                    self.errors.setdefault(field, []).append(msg)
+
+    def any_in_multivalues_validator(self, data):
+        """
+        Check if every row has at least one required value passed/filled.
```
ghost: "Checks if each row has filled at least one field specified by `one_of_mulitvalue_required`"?

ghost:
:ok:

src/ralph/admin/fields.py
```diff
@@ -0,0 +1,154 @@
...
+    separators = ",|\n"
+
+    def __init__(self, allow_duplicates=True, *args, **kwargs):
+        self.allow_duplicates = allow_duplicates
+        super().__init__(*args, **kwargs)
+
+    def validate(self, values):
+        if not values and self.required:
+            error_msg = _(
+                "Field can't be empty. Please put the item OR items separated "
```
damian1976, line 28: Maybe this message will be a bit better?:) "Field cannot be empty. Please insert an item (multiple items can be newline or comma separated)."

PR #1762, opened 2015-09-04
Optional `status` field on Device.

ghost:
Device's `status` field should be excluded from form validation.
The field is only for informative purpose. So `required` changed to `False`.


ar4s:
:ok:

szok:
:ok:

Issue #1692, opened 2015-08-11
Changing location of parent object should change location of all descendants


mkurek:
Changing location (ex. rack) of parent object should change location of all it's descendants (ex. blade chassis - blade servers).

PR #1763, opened 2015-09-04
Changing location of parent object should change location of all descendants

szok:
Added:
- Saving rack for all descendants

Connected to #1692


ghost:
:ok:

mkurek:
:ok:

PR #1766, opened 2015-09-08
Added profit center and business segment to importer

mkurek:
- added business segment and profit center resources to importer
- fixed rack repr, when it doesn't have server room assigned


ghost:
:ok:

PR #1771, opened 2015-09-09
Allow primary key fields to be accessed in read-only mode.

msuszko:
This adds view permission to primary key fields.

### DIFF
--- a/src/ralph/lib/permissions/models.py
+++ b/src/ralph/lib/permissions/models.py
@@ -157,18 +157,18 @@ class PermissionsBase(ModelBase):
         )
         for field in model_fields:
             name = field.name
-            if (
-                not field.primary_key and
-                name not in new_class._permissions.blacklist
-            ):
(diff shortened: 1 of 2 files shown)


mkurek:
@msuszko what is your motivation for this? what is the possible use case of readonly id?

msuszko:
I'd like to present id field on admin list (.list_display). It works unless PermissionAdminMixin is used or user is an admin.
This patch allows to grant _view permission, and view id field for non-admin users.

msuszko:
Motivations is the id field doesn't change and can be used as additional reference (as serial number, barcode, inventory number). I'm developing buisiness application registry on top of Ralph 3 and it has primary key used as application identifier

mkurek:
so it's :ok: for me, but please squash it into one commit

msuszko:
done.

vi4m:
LGTM

ar4s:
:ok:

Issue #1765, opened 2015-09-08
Bulk add improvements


mkurek:
Continue of https://github.com/allegro/ralph/pull/1761
- bulk add button should be available on add views (not only edit views)
- on edit views this button should be called differently (ex. 'Copy', 'Duplicate', etc)
- bulk add for BO Asset should have imei field as well only if specified asset category was selected - see https://github.com/allegro/ralph/pull/1761#discussion-diff-38838736 for details

PR #1772, opened 2015-09-09
Bulk add improvements

szok:
Added:
- Add multi add button to add view
- Change name from Bulk add to Duplikace
- Add Imei field to Back Office Asset
- Add Imei field to Back Office multi add view

Connected to #1765

### DIFF
--- a/src/ralph/admin/fields.py
+++ b/src/ralph/admin/fields.py
@@ -29,7 +29,9 @@ class MultilineField(forms.CharField):
                 "by new line or comma."
             )
             raise forms.ValidationError(error_msg, code='required')
-        non_empty_values = [item for item in values if len(item.strip()) > 0]
+        non_empty_values = [
+            item for item in values if str(item).strip()
+        ]
         if not self.allow_duplicates:
             has_duplicates = len(set(non_empty_values)) != len(non_empty_values)
             if has_duplicates:
@@ -42,6 +44,16 @@ class MultilineField(forms.CharField):
                 items.append(item.strip(' \t\n\r'))
 
         return items
+
+
+class IntegerMultilineField(MultilineField):
+
+    def to_python(self, value):
+        result = super().to_python(value)
+        try:
+            return [int(i) for i in result]
+        except ValueError:
+            raise ValidationError(_('Enter a valid number.'))
 
 
 class MultivalueFormMixin(object):
--- a/src/ralph/admin/views/multiadd.py
+++ b/src/ralph/admin/views/multiadd.py
@@ -3,12 +3,17 @@ from django import forms
 from django.conf.urls import url
 from django.contrib import messages
 from django.core.urlresolvers import reverse
-from django.db import IntegrityError, transaction
+from django.db import IntegrityError, models, transaction
+from django.forms import ValidationError
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django.utils.translation import ugettext_lazy as _
 
-from ralph.admin.fields import MultilineField, MultivalueFormMixin
+from ralph.admin.fields import (
+    IntegerMultilineField,
+    MultilineField,
+    MultivalueFormMixin
+)
 from ralph.admin.mixins import RalphTemplateView
 from ralph.admin.sites import ralph_site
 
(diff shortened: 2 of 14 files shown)


src/ralph/admin/views/multiadd.py
```diff
@@ -153,3 +161,15 @@ def get_urls(self):
             ),
         ]
         return _urls + urls
+
+    def response_add(self, request, obj, post_url_continue=None):
+        """
+        Override respodne add from django model admin.
```
szok: response

src/ralph/back_office/models.py
```diff
@@ -40,6 +40,9 @@ class BackOfficeAsset(Regionalizable, Asset):
         default=AssetStatus.new.id,
         choices=AssetStatus(),
     )
+    imei = models.CharField(
```
mkurek: wasn't there any validator for imei in old ralph?

src/ralph/admin/fields.py
```diff
@@ -44,6 +46,17 @@ def to_python(self, value):
...
 
 
+class IntegerMultilineField(MultilineField):
+
+    def to_python(self, value):
+        result = super().to_python(value)
+        try:
+            result = [int(i) for i in result]
+        except ValueError:
+            result = []
```
mkurek: shouldn't any ValidationError be raised?

src/ralph/admin/views/multiadd.py
```diff
@@ -34,7 +39,15 @@ def get_form(self):
             'model': self.model
         }
         for field in self.fields:
-            multi_form_attrs[field] = MultilineField(allow_duplicates=False)
+            field_type = self.model._meta.get_field(field)
+            if isinstance(field_type, models.IntegerField):
+                multi_form_attrs[field] = IntegerMultilineField(
+                    allow_duplicates=False
```
mkurek: it should be possible to add multiple items with the same position (ex. blade servers), so allow_duplicates should be set to True here sometimes. Is it possible to specify it on field level?

szok: No, I can add allow_duplicates options

src/ralph/back_office/models.py
```diff
@@ -58,6 +63,18 @@ def __str__(self):
     def __repr__(self):
         return '<BackOfficeAsset: {}>'.format(self.id)
 
+    def validate_imei(self):
```
mkurek, line 70: this in one-field-only related so you could attach it to imei field (instead of calling it in clean): https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.validators

szok, line 70: validators works only in modelform, does not work in clean() on model

src/ralph/admin/fields.py
```diff
@@ -29,7 +29,9 @@ def validate(self, values):
                 "by new line or comma."
             )
             raise forms.ValidationError(error_msg, code='required')
-        non_empty_values = [item for item in values if len(item.strip()) > 0]
+        non_empty_values = [
+            item for item in values if len(str(item).strip()) > 0
```
vi4m: is it pythonic?

src/ralph/settings/base.py
```diff
@@ -167,3 +168,7 @@
     'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',  # noqa
     'PAGE_SIZE': 10,
 }
+
+
+IMEI_UNTIL_2003 = re.compile(r'^\d{6} *\d{2} *\d{6} *\d$')
```
vi4m: We should think about another way for doing this - it's specific for allegro, so not open source applicable, nor should sits in settings in the current form.

vi4m: IMEI_VALIDATORS + dependency injection + allegro custom settings(not settings.py) + later etcd should do the job

mkurek: it's not allegro specific - it's international standard (https://en.wikipedia.org/wiki/International_Mobile_Station_Equipment_Identity#Structure_of_the_IMEI_and_IMEISV_.28IMEI_Software_Version.29), so IMO this should be placed in source code, not settings (it was placed in source code in old assets too: https://github.com/allegro/ralph_assets/blob/33073a541c7484ce294f385e14fd167424193e50/src/ralph_assets/forms.py#L373)

src/ralph/back_office/models.py
```diff
@@ -58,6 +63,18 @@ def __str__(self):
...
+        is_imei = settings.IMEI_UNTIL_2003.match(self.imei) or \
+            settings.IMEI_SINCE_2003.match(self.imei)
```
mkurek: i'd revert this condition, to check first to match new style imei - probably all of devices stored in ralph will have new style imei, so there would be no need to check second condition

src/ralph/data_center/admin.py
```diff
@@ -169,8 +169,12 @@ class DataCenterAssetAdmin(
         }),
     )
 
-    def get_multiadd_fields(self):
-        return ['sn', 'barcode', 'position']
+    def get_multiadd_fields(self, obj=None):
+        return [
+            {'field': 'sn', 'allow_duplicates': False},
+            {'field': 'barcode', 'allow_duplicates': False},
+            {'field': 'position', 'allow_duplicates': False},
```
mkurek: position field could have duplicates

src/ralph/admin/fields.py
```diff
@@ -44,6 +46,16 @@ def to_python(self, value):
...
 
 
+class IntegerMultilineField(MultilineField):
+
+    def to_python(self, value):
+        result = super().to_python(value)
+        try:
+            return [int(i) for i in result]
+        except ValueError:
+            raise ValidationError(_('Enter a whole number.'))
```
mkurek: a valid number

src/ralph/back_office/tests/test_models.py
```diff
@@ -110,3 +111,18 @@ def test_convert_iso2_to_iso3(self):
...
+
+        with self.assertRaisesRegex(
+            ValidationError,
+            '{} is not IMEI format'.format(bo_asset_failed.imei)
+        ):
+            bo_asset_failed.validate_imei()
+
+        try:
+            bo_asset.validate_imei()
+        except Exception as e:
```
ar4s: why you catch `Exception`?

szok: It is expected that here no exception. How will exception I raise fail tests

ar4s:
:ok:

src/ralph/back_office/models.py
```diff
@@ -58,6 +64,18 @@ def __str__(self):
     def __repr__(self):
         return '<BackOfficeAsset: {}>'.format(self.id)
 
+    def validate_imei(self):
+        imei_until_2003 = re.compile(r'^\d{6} *\d{2} *\d{6} *\d$')
```
mkurek: make them constants at the module level and add some comments about this regexps (ex. link to wiki, example imeis etc)

mkurek:
:ok:

PR #1773, opened 2015-09-09
Custom search filters docs + date filter improvement

mkurek:
- added docs for creating custom search filter
- from now 'parameter_name_start' and 'parameter_name_end' params are generated automatically


szok:
:ok:

ar4s:
:ok:

Issue #1745, opened 2015-08-29
Category model status report error


vi4m:
FieldError showing, instead of the report.

Steps to reproduce: click Reports -> Category model status

Result:

FieldError at /category_model__status_report
Cannot resolve keyword 'status' into field. Choices are: backofficeasset, barcode, baseobject, baseobject_ptr, baseobject_ptr_id, content_type, content_type_id, created, datacenterasset, depreciation_end_date, depreciation_rate, diskshare, disksharemount, force_depreciation, genericcomponent, hostname, id, invoice_date, invoice_no, ipaddress, licences, model, model_id, modified, niw, order_no, parent, parent_id, price, provider, remarks, required_support, service_env, service_env_id, sn, supports, task_url

Package: ralph-3.0.0-26

PR #1774, opened 2015-09-09
Fixed status field in reports

mkurek:
In one of transitions commits, status field was moved from `Asset` separately to `DataCenterAsset` and `BackOfficeAsset`. This causes some of reports with status field (for Asset model) to crash. In this commit I turned off 'all' option in 'Category - model - status' and 'Status - model' reports - these reports are now explicit, either in DC mode or BO mode. Default mode is DC.

Conntected to #1745


ar4s:
:ok:

szok:
:ok:

Issue #1735, opened 2015-08-27
API token authentication


mkurek:
User should authenticate to API using generated token. We could use DRF TokenAuthentication mechanism: http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication

Requirements:
- token is generated for every already created user
- token is generated when saving user which doesn't have token assigned
- expose view where user could obtain token after passing valid username and password
- consider token expiration (ex. after 1 day, 1 week etc)

PR #1777, opened 2015-09-10
Added API Token authentication

mkurek:
- token api authentication on production
- moved `RalphPermission` class to `ralph.lib.permissions` (it couldn't be imported from `ralph.api` by DRF because of circular import dependency and thus it doesn't have any ralph-dependency I moved it to ralph.lib

Connected to #1735


szok:
:ok:

ar4s:
:ok:

Issue #1515, opened 2015-06-23
DC models REST API


mkurek:
Working REST API for DC models (create, get, list, update):
- [x] DataCenter
- [x] ServerRoom
- [x] RackAccessory
- [x] Rack
- [x] DataCenterAsset
- [ ] Database
- [ ] Network
- [ ] NetworkEnvironment ? 
- [ ] IPAddress
- [x] AssetModel
- [x] AssetManufacturer
- [x] Service
- [x] Environment
- [x] ServiceEnvironment
- [x] Category
- [ ] VIP
- [ ] VirtualServer
- [ ] CloudProject

Components could be now skipped.

PR #1784, opened 2015-09-11
DataCenter part REST API

mkurek:
Created REST endpoints for:
- DataCenterAsset
- Rack
- RackAccessory
- Accessory
- ServerRoom
- DataCenter

Other DC models are not usable now (VIP, Network, Virtual etc), so there is no need for now for API resources for them

I changed default behaviour of DRF form in browsable API - now it displays input field for foreign keys instead of choice field, which perform very poorly. Eventually autocomplete widget should work here.

Connected to #1515


ar4s:
:ok:

szok:
:ok:

PR #1787, opened 2015-09-11
CSS improvements

ar4s:
Changes:
- changed autocomplete widget height,
- full user's name,
- search in popup,
- changed form spacing,

pr-impro-1
pr-impro-2
pr-impro-3
auto-error

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@ var gulp = require('gulp');
     bower = require('gulp-bower'),
     prefixer = require('gulp-autoprefixer'),
     sass = require('gulp-sass'),
+    sourcemaps = require('gulp-sourcemaps'),
     qunit = require('node-qunit-phantomjs');
 
 var config = {
@@ -15,6 +16,7 @@ var config = {
 }
 
 var sass_config = {
+    outputStyle: 'compressed',
     includePaths: [
         config.bowerDir + 'foundation/scss',
         config.bowerDir + 'fontawesome/scss',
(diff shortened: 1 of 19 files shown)


vi4m:
I don't know where to post this comment, please address this issues as well:
1. Remove add button
2. Definitelly remove this refresh icon and ! - does it present ajax/autocomplete errors? Error reporting should look exactly like field validation error with the description text.
3. Replace adding button with the following scenario:
4. when no match is found, only then display plus button (maybe with the "Add new" description)

mkurek:
@vi4m:
1. these icons are displayed only when you type something (refresh) or there is server-side error (ex. 500) (!) - this is not related to validation error

About add button - do you mean something like this?

screen shot 2015-09-15 at 09 18 36

ar4s:
> Remove add button

@vi4m can you explain why? IMHO it's OK -- user can add something (`Rack` in this case) at any time

> Definitelly remove this refresh icon and ! - does it present ajax/autocomplete errors? Error reporting should look exactly like field validation error with the description text.

The icons will appear if server returns 5xx or 4xx

> Replace adding button with the following scenario:
> when no match is found, only then display plus button (maybe with the "Add new" description)

OK, but can I change this in other PR? Because I think this is beyond the scope of this PR.

vi4m:
@mkurek proposition looks exactly as i meant when it comes to adding icon

 </img>

vi4m:
@ar4s yes, we can postpone `add button refinements` for other pull request.
@ar4s but we should change the style of error reporting in case of server errors, to something like this:

zrzut ekranu 2015-09-18 o 15 41 00

ar4s:
> but we should change the style of error reporting in case of server errors, to something like this

I agree, but it's not just pure styling work it requires some changes in widget's logic. I propose merge all autocomplete suggestions and create issue for this and check this PR for CSS improvements.

mkurek:
I've created separated issue for that.

This PR is :ok: for me.

szok:
:ok:

PR #1788, opened 2015-09-11
Additional service and related models fields, team model introduced

mkurek:
- added description field to profit center
- added Team model
- user could been assigned to (one) team
- service has assigned (support) team
- added service `active` field


ar4s:
:ok:

szok:
:ok:

Issue #1791, opened 2015-09-16
REP 2: Ralph modules architecture


vi4m:
# Introduction

According to the [Zen of Ralph](https://github.com/allegro/ralph/wiki/The-Zen-of-Ralph) we should avoid introducing new requirements for the main ralph package. 

Recently we started work on generating PDF reports / transition outprints which doesn't cover ralph scope itself. 

Therefore i propose simple and flexible architecture for separating concerns, following industry standards microservices architecture to separater further ralph modules to separate processes.
# Ralph extensions architecture

Every new ralph extension, such as inky connector should be separated from main ralph package.
1. Separate repository with the ralph_ext_ prefix
2. Separate process, accessible from any network address
3. Architecture doesn't depend on concrete microservices transport, such as nameko, lymph, or raw redis invocations
4. Shared configuration using etcd daemon
# Example - inkpy
1. We create github repository called ralph_ext_inkpy with custom dependencies. This package doesn't depend on ralph itself, we don't want circural dependencies.
2. We introduce extensions index in the documentation which explains what this extensions does.
3. To simplify adoption of microservices, let's start with something simple. I thereby propose using just raw redis - rq  sync queues which gives us separation of processes, boundaries of services, but on the other hand -  doesn't introduce any new complex dependency such as rabbitmq, or zookeeper to maintain, and should be sufficient for our needs.

mkurek:
:+1: for using just raw redis.

We could use redis not only as job queue (we're using it right now through RQ - is simple, but covers all our requirements) - we could use it as a message broker/cache in microservices world. 

Inkpy is IMO not perfect example for presenting microservices architecture - inpky could be used as (synchronous) job executor using just RQ - everything we need is to schedule job on specific queue (ex. `ext_pdf`) with template and data passed. Inkpy should be plugged into redis (using python-rq) and listed on this queue. It's simple usage of rq jobs just like in Ralph 2.

If we want to have real microservices (ex. display content from external service), we could also use redis to do this (assuming for example only static content displayed by external service). 

Proposed requirements:
- content from external (micro)service is displayed on separated tab attached to model view (using existing extra views mechanism) - it's only static content
- every service which want to display content in Ralph should be approved by Ralph maintainers and authorized (using appropriate data in settings, ex. service auth token and possible models/tabs to work on)
[...]

PR #1793, opened 2015-09-16
[WIP] Added django_rq and ``ext`` queue for extensions

ar4s:
It's first step to microservices. It's closely related with #1791.

### DIFF
--- a/src/ralph/settings/base.py
+++ b/src/ralph/settings/base.py
@@ -25,6 +25,7 @@ INSTALLED_APPS = (
     'import_export',
     'reversion',
     'sitetree',
+    'django_rq',
     'ralph.accounts',
     'ralph.assets',
     'ralph.attachments',
@@ -167,4 +168,13 @@ REST_FRAMEWORK = {
     ),
     'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',  # noqa
     'PAGE_SIZE': 10,
+}
+
+RQ_QUEUES = {
+    'ext': {
+        'HOST': 'localhost',
+        'PORT': 6379,
+        'DB': 0,
+        'ASYNC': False,
+    },
 }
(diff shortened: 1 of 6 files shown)


vagrant/upstart/rqworker.conf
```diff
@@ -0,0 +1,6 @@
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+respawn
+chdir /home/vagrant
+exec bin/ralph rqworker ext--settings=ralph.settings.dev
```
mkurek: don't you miss space here?

ar4s: yep :)

src/ralph/settings/base.py
```diff
@@ -168,3 +169,12 @@
     'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',  # noqa
     'PAGE_SIZE': 10,
 }
+
+RQ_QUEUES = {
+    'ext': {
+        'HOST': 'localhost',
```
mkurek: use `os.environ.get('REDIS_HOST', 'localhost')` and make it working in docker too (in docker compose run redis container)

src/ralph/settings/base.py
```diff
@@ -168,3 +169,12 @@
     'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',  # noqa
     'PAGE_SIZE': 10,
 }
+
+RQ_QUEUES = {
+    'ext': {
+        'HOST': os.environ.get('REDIS_HOST', 'localhost'),
+        'PORT': os.environ.get('REDIS_PORT', '6379'),
+        'DB': 0,
+        'ASYNC': False,
```
mkurek, line 178: why not async?

ar4s, line 178: because we want run task and gets results immediately

PR #1795, opened 2015-09-16
Change controller behaviour depending whether rack is added or edited

damian1976:
Change the message from "Rack has been successfully added" in case when the rack is edited...

### DIFF
--- a/src/ralph/dc_view/static/js/app/data_center/controllers.js
+++ b/src/ralph/dc_view/static/js/app/data_center/controllers.js
@@ -44,11 +44,14 @@
             $scope.addOrEdit = function(rack) {
                 var rack_model = new RackModel(rack);
                 var rack_promise = null;
-                if (rack.id !== undefined) {
+                var success_msg = '';
+                if (typeof(rack.id) !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/ralph/dc_view/static/js/app/data_center/controllers.js
```diff
@@ -59,7 +59,12 @@
                         rack.new = false;
                         rack.saved = true;
                         $scope.forms.edit_form.$error.all = null;
-                        $scope.forms.edit_form.$success = ['The rack has been successfully added to data center.'];
+                        if (rack.id !== undefined) {
```
vi4m: if typeof(rack.id) !== "undefined"

damian1976: OK. If so, the line no. 47 (original) should also be changed in the same way (?)

vi4m: Congrats - you've found a potential mistake - you can fix it as well :)

ar4s:
:ok:

mkurek:
:ok:

Issue #1799, opened 2015-09-17
Added rack in DC Visualization without server room


ar4s:
User can add rack without server room beacause backend doesn't validate `server_room`

PR #1800, opened 2015-09-17
Added required condition to server room in DC view

ar4s:
Server room must be specify when user add rack from DC view.
Connected to #1799

### DIFF
--- a/src/ralph/dc_view/static/js/app/data_center/controllers.js
+++ b/src/ralph/dc_view/static/js/app/data_center/controllers.js
@@ -23,8 +23,8 @@
                 rack.visualization_col = x;
                 rack.visualization_row = y;
                 rack.orientation = 'top';
+                rack.server_room = '';
                 rack.name = 'New rack';
-                rack.data_center = data_center.id;
                 rack.new = true;
                 data_center.rack_set.push(rack);
                 $scope.$emit('edit_rack', rack);
@@ -45,6 +45,7 @@
                 var rack_model = new RackModel(rack);
                 var rack_promise = null;
                 var success_msg = '';
+
                 if (typeof(rack.id) !== 'undefined') {
                     rack_promise = rack_model.$update();
                     success_msg = 'updated';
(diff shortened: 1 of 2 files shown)


src/ralph/dc_view/static/partials/data_center/data_center.html
```diff
@@ -46,13 +46,15 @@
                   </div>
                 </div>
                 <div class="row">
-                  <div class="large-12 columns">
+                  <div class="large-12 columns" ng-class="{'error': forms.edit_form.server_room.$invalid}">
                       <label for="id_name">Server room
-                        <select name="server_room" ng-model="rack.server_room"  required>
+                        <select name="server_room" ng-model="rack.server_room" required>
+                          <option value="">-----</option>
```
mkurek: if this field is required why do you add empty option?

src/ralph/dc_view/static/js/app/data_center/controllers.js
```diff
@@ -28,6 +29,20 @@ angular
...
         };
+=======
+            $scope.newRack = function(data_center, x, y) {
+                var rack = new RackModel();
+                rack.visualization_col = x;
+                rack.visualization_row = y;
+                rack.orientation = 'top';
+                rack.server_room = '';
+                rack.name = 'New rack';
+                rack.data_center = data_center.id;
```
mkurek: in ng model rack is not connected to dc directly, but through server room

src/ralph/dc_view/static/partials/data_center/data_center.html
ar4s: ask angular
mind_blow

src/ralph/dc_view/static/js/app/data_center/controllers.js
ar4s: you right.. there are some legacy code from 2.x

mkurek:
:ok:

ghost:
:ok:

Issue #1665, opened 2015-08-03
Export/Import from/to 2.x/3.x


damian1976:
All assets models are imported as **back office** type.

mkurek:
Also BO hostname is not exported/imported

mkurek:
- not all bo assets, supports and licences are exported because of regions (admin_objects should be used)

mkurek:
Thanks @szok - i almost have ready fix for it, but I didn't work out it finally before my leave. I'll tweak it and push in on monday.

PR #1801, opened 2015-09-18
Use admin_objects_bo to export all objects to NG (without checking user region)

mkurek:
Regular `objects` manager excluded some of assets during export because there is no user in shell, so default region was assumed for this manager, what causes limiting exported assets only to assets with default region set.


ar4s:
:ok:

ar4s:
connected to #1665

vi4m:
Can you move PR comment into the commit message itself ?

mkurek:
@vi4m Done

vi4m:
LGTM then.

Issue #1789, opened 2015-09-11
BaseObject in API


mkurek:
BaseObject should be exposed in REST API
- it should behave similar to https://github.com/allegro/ralph/pull/1725 mechanism - return list of target objects (and use target serializers for them)
- allow to filter by service
- parent field should use this serializer too (but only for depth 1)

PR #1804, opened 2015-09-18
Polymorphic objects in API

mkurek:
Return polymorphic objects to API using dedicated serializers for each inheriting model.
Use this polymorphic logic to BaseObject in API.
Added `polymorphic_select_related` and `polymorphic_prefetch_related` to Polymorphic queryset.
Minor performance improvements in polymorphic logic.

Connected to #1789

### DIFF
--- a/src/ralph/api/serializers.py
+++ b/src/ralph/api/serializers.py
@@ -143,8 +143,22 @@ class RalphAPISaveSerializer(serializers.ModelSerializer):
         return field_class, field_kwargs
 
 
+serializers_registry = {}
+
+
+class RalphAPISerializerMetaclass(serializers.SerializerMetaclass):
+    def __new__(cls, name, bases, attrs):
+        attrs['_serializers_registry'] = serializers_registry
+        new_cls = super().__new__(cls, name, bases, attrs)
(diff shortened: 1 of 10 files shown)


src/ralph/api/serializers.py
```diff
@@ -143,8 +143,22 @@ def build_field(self, *args, **kwargs):
...
 
+serializers_registry = {}
+
+
+class RalphAPISerializerMetaclass(serializers.SerializerMetaclass):
+    def __new__(cls, name, bases, attrs):
+        attrs['_serializers_registry'] = serializers_registry
+        new_cls = super().__new__(cls, name, bases, attrs)
+        meta = getattr(new_cls, 'Meta', None)
+        if getattr(meta, 'model', None):
```
ar4s, line 154: `if meta and getattr(meta, 'model', None):` to prevent `AttributeError`

mkurek, line 154: getattr is working fine with None:

```
>>> x = None
>>> print(getattr(x, 'model', None))
None
```

ar4s:
:ok:

szok:
:ok:

PR #1805, opened 2015-09-18
Added some test to DC Visualization

ar4s:
Motivation: Code without tests is bad code.
Added way to test code written in JS by QUnit and PhantomJS.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,4 +1,4 @@
-var gulp = require('gulp');
+var gulp = require('gulp'),
     watch = require('gulp-watch'),
     runSequence = require('run-sequence'),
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -77,8 +77,9 @@ gulp.task('js', function(){
 });
```
pnuckowski, line 77: First line of this file should ends with comma, not semicolon.

ar4s, line 77: Thanks for review!

mkurek:
is it still WIP? If not, it's :ok: for me

szok:
:ok:

PR #1807, opened 2015-09-22
New custom filters

szok:
Change support for custom filters in Django Admin

Added:
- ForeignKey support
- better verbose name


ar4s:
:ok:

src/ralph/accounts/admin.py
```diff
@@ -91,6 +92,21 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class StaffStatusFilter(BooleanFilter):
```
vi4m: Do we need to declare dozens of classes every time we want to add new filters? Maybe it would be better to create simple metaclass which will take away trivial params cases like (title, param_name) ?

For example:

```
filters.add(
    ('Active', 'is_active'), 
    ('Superuser', 'is_superuser'), 
    ....
```

szok: It's :ok: for me.
But in some filters we have greater logic.
Maybe generator classes for filters? RalphFilter(name, params, type) class return (SuperuserFilter(BooleanFilter))

@ar4s, @mkurek, @xliiv What do you think?

mkurek: in my opinion we should restore default django behaviour here for simple cases and use field name as @vi4m suggest. We'd not need metaclass for this - there is mechanism to register appropriate filter class for particular field type using `FieldListFilter` (see `django.contrib.admin.filters` and `django.contrib.admin.views.main.ChangeList:get_filters` for details). This would probably solve 90% of our cases and then we'd use custom filter classes when complex logic is required.

quamilek:
Write the description of the patch, now it is a black box :smiley:

src/ralph/data_center/admin.py
```diff
@@ -132,9 +118,9 @@ class DataCenterAssetAdmin(
     bulk_edit_list = list_display
     search_fields = ['barcode', 'sn', 'hostname', 'invoice_no', 'order_no']
     list_filter = [
-        StatusFilter, BarcodeFilter, SNFilter, HostnameFilter, InvoiceNoFilter,
-        InvoiceDateFilter, OrderNoFilter, ModelFilter, DepreciationDateFilter,
-        ForceDepreciationFilter, RemarksFilter, RackFilter
+        'status', 'barcode', 'sn', 'hostname', 'invoice_no', 'invoice_date',
```
vi4m, line 121: what happens if i mis-spell something here? Will I get an error during the application startup?

szok, line 121: Application startup. Django checks verify.

src/ralph/admin/filters.py
```diff
@@ -3,13 +3,29 @@
...
-from django.contrib.admin import SimpleListFilter
+from django.contrib.admin.filters import FieldListFilter
+from django.db import models
 from django.db.models import Q
 from django.utils.encoding import smart_text
 from django.utils.formats import get_format
 from django.utils.translation import ugettext_lazy as _
 
 
+class RalphFieldListFilter(FieldListFilter):
```
vi4m: this class doesn't have meaningful docstring

src/ralph/admin/filters.py
```diff
@@ -32,7 +48,17 @@ def date_format_to_human(value):
     return value
 
 
-class BaseCustomFilter(SimpleListFilter):
+class BaseCustomFilter(RalphFieldListFilter):
```
vi4m: class docstring missing?

src/ralph/admin/filters.py
```diff
@@ -78,14 +111,21 @@ class BooleanFilter(ChoicesFilter):
     ]
 
 
+RalphFieldListFilter.register(
+    lambda f: isinstance(f, (models.BooleanField, models.NullBooleanField)),
```
vi4m: Not readable enough.

src/ralph/admin/filters.py
```diff
@@ -40,6 +66,9 @@ def has_output(self):
     def lookups(self, request, model_admin):
         return None
 
+    def expected_parameters(self):
```
vi4m: what's that?

vi4m: ok, never mind, i've found `Returns the list of parameter names that are expected from the request’s query string and that will be used by this filter.`

src/ralph/admin/filters.py
```diff
@@ -48,6 +77,9 @@ class ChoicesFilter(BaseCustomFilter):
     template = 'admin/filters/choices_filter.html'
     choices_list = []
 
+    def get_choices_list(self):
```
vi4m: java programming?

vi4m: what about the @property ?

szok: :+1:

src/ralph/admin/filters.py
vi4m: Nor class name gives a hint what it does RalphFieldListFilter ?

vi4m:
I like this pull request - please polish the edges and ship it :)

vi4m:
+ documentation for developer sections how to define custom filter

src/ralph/admin/filters.py
ar4s: IMO this class name is enough descriptive - `RalphFieldListFilter` is our `FieldListFilter` with some overrides

vi4m: I mean, FieldListFilter is ok, but why follow Ralph.\* convension? Why django doesn't do DjangoFieldListFilter?

src/ralph/admin/filters.py
```diff
@@ -164,3 +236,25 @@ def choices(self, cl):
             'current_value': self.value(),
             'parameter_name': self.parameter_name
         },)
+
+
+RalphFieldListFilter.register(
+    lambda f: isinstance(f, (models.CharField, models.TextField)),
```
mkurek: what about numeric fields like `IntegerField`?

src/ralph/admin/filters.py
```diff
@@ -164,3 +236,25 @@ def choices(self, cl):
...
+
+
+class RelatedFieldListFilter(ChoicesFilter):
+
+    """Filter for Foregin key field."""
+
+    @property
+    def choices_list(self):
+        model = get_model_from_relation(self.field)
+        return [(i.id, str(i)) for i in model.objects.all()]
```
mkurek: we should probably use default manager instead of `objects`

src/ralph/admin/filters.py
vi4m: Please think about changing register priorities to simplify our prefixed classes maybe?

src/ralph/admin/filters.py
```diff
@@ -32,7 +34,33 @@ def date_format_to_human(value):
...
 
 
-class BaseCustomFilter(SimpleListFilter):
+class BaseCustomFilter(FieldListFilter):
+
+    """Base class for custom filters."""
+
+    def __init__(self, field, request, params, model, model_admin, field_path):
+        self.lookup_kwarg = field_path
+        self.parameter_name = field_path
```
mkurek: since we're inheriting from `FieldListFilter` now, parameter_name field is no longer necessary - you could use `self.field_path` instead

ar4s:
:ok:

mkurek:
:ok:

PR #1808, opened 2015-09-22
Changed order step in Dockerfile

ar4s:
To cache `provision.sh`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,10 @@ ENV RALPH_STATIC=/root/static
 ENV RALPH_DOCS=$RALPH_DIR/docs
 ENV SCRIPTS_PATH=/root
 
-ADD docker/* $SCRIPTS_PATH/
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -15,11 +15,11 @@ ENV RALPH_STATIC=/root/static
 ENV RALPH_DOCS=$RALPH_DIR/docs
 ENV SCRIPTS_PATH=/root
 
-ADD docker/* $SCRIPTS_PATH/
-
 # basic provisioning
 RUN $SCRIPTS_PATH/provision.sh
 
+ADD docker/* $SCRIPTS_PATH/
```
mkurek, line 21: this will not work - first you run provision script, then you add this script. How should it work? Run `provision.sh` is cached until someone change something  in `docker/` directory.

vi4m, line 21: @mkurek is right, but ...

PR #1810, opened 2015-09-25
Bumped django's version in travis

ar4s:


mkurek:
please allow failures for Django 1.9 for now (since not all dependent packages support it). Also bump Django version to 1.8.5 (https://www.djangoproject.com/weblog/2015/oct/03/bugfix-release-issued/)

Issue #1693, opened 2015-08-11
Additional views permissions


mkurek:
There should be possibility to specify view/change permissions for additional views (ex. model admin page tabs, custom views like visualisation etc).

mkurek:
Permissions for visualisation API should be the same as visualisation UI (when someone has access to visualisation, he has access to visualisation API)

PR #1813, opened 2015-09-30
Additional view permissions.

szok:
Added new metaclass for Ralph extra view with permission.

Connected to #1693

### DIFF
--- a/src/ralph/lib/permissions/views.py
+++ b/src/ralph/lib/permissions/views.py
@@ -1,0 +1,86 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+from django.http import HttpResponseForbidden
+
+from ralph.admin.sites import ralph_site
+
+logger = logging.getLogger(__name__)
+
+_permission_views = []
+
+
+def view_permission_dispatch(func):
+    """
+    Adding to check the user has permission to dispatch method.
+    """
+    def wraps(self, request, *args, **kwargs):
+        model = kwargs.get('model', None)
+        if not model:
+            model = get_user_model()
+        perm_name = '{}.{}'.format(
+            model._meta.app_label, self.permision_codename
+        )
+        if not request.user.has_perm(perm_name):
+            return HttpResponseForbidden()
+        return func(self, request, *args, **kwargs)
+    return wraps
+
+
+class PermissionViewMetaClass(type):
+
+    """
+    Adding permission to additional views.
+    """
+
+    def __new__(cls, name, bases, attrs):
+        codename = 'can_view_extra_{}'.format(name.lower())
+
+        attrs['permision_codename'] = codename
+        new_class = super().__new__(cls, name, bases, attrs)
+        dispatch = getattr(new_class, 'dispatch', None)
+        setattr(new_class, 'dispatch', view_permission_dispatch(dispatch))
+        _permission_views.append((new_class, codename))
+        return new_class
+
+
+def update_extra_view_permissions(**kwargs):
+    """
+    Get all views that inherit the PermissionViewMetaClass and
+    adding them permission.
+    """
+    admin_classes = {}
+    for model, admin_class in ralph_site._registry.items():
+        for change_view in admin_class.change_views:
+            admin_classes[change_view] = model
+
+    old_permission = Permission.objects.filter(
+        codename__startswith='can_view_extra_'
+    ).values_list('id', flat=True)
+    current_permission = []
+    for class_view, codename in _permission_views:
+        model = admin_classes.get(class_view, None)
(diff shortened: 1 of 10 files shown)


src/ralph/accounts/admin.py
```diff
@@ -61,7 +62,7 @@ def url(self, item):
     url.title = _('Link')
 
 
-class UserInfoView(RalphDetailView):
+class UserInfoView(RalphDetailView, metaclass=PermissionViewMetaClass):
```
mkurek: why `RalphDetailView` cannot use `PermissionViewMetaClass` metaclass directly?

src/ralph/lib/permissions/views.py
```diff
@@ -0,0 +1,53 @@
...
+class PermissionViewMetaClass(type):
+
+    def __new__(cls, name, bases, attrs):
+        codename = 'can_view_{}'.format(name.lower())
+
+        attrs['permision_codename'] = codename
+        new_class = super().__new__(cls, name, bases, attrs)
+        dispatch = getattr(new_class, 'dispatch', None)
+
+        def new_dispatch(self, request, *args, **kwargs):
```
mkurek: first of all, this looks just like decorator - you could use mixin class to do this, instead of metaclass. Second of all - what if someone overwrite dispatch without calling super? It's not super-safe :)

src/ralph/accounts/admin.py
szok: Because other class inherit from RalphDetailView and we have conflict (example: RalphDetailViewAdmin)

src/ralph/lib/permissions/views.py
```diff
@@ -0,0 +1,53 @@
...
+    for model, admin_class in ralph_site._registry.items():
+        for change_view in admin_class.change_views:
+            admin_classes[change_view] = model
+
+    for class_view, codename in _permission_views:
+        model = admin_classes.get(class_view, None)
+        if not model:
+            model = get_user_model()
+        ct = ContentType.objects.get_for_model(model)
+        Permission.objects.get_or_create(
```
mkurek: what with permissions that are no longer necessary? For example associated view is no longer existing in source code.

src/ralph/lib/transitions/templates/transitions/templatetags/available_transitions.html
```diff
@@ -8,7 +8,7 @@
   <ul id="transitions_status" class="f-dropdown" data-dropdown-content aria-hidden="true" tabindex="-1">
     {% for transition in transitions %}
       <li>
-        <a href="{% url transition_url_name original.pk transition.pk %}">{{ transition.name }}</a>
+        <a href="{# {% url transition_url_name original.pk transition.pk %} #}">{{ transition.name }}</a>
```
mkurek: ?

src/ralph/accounts/admin.py
mkurek: using only `PermissionViewMetaClass` for `RalphDetailView` and compound metaclass for `RalphDetailViewAdmin` should do the job

src/ralph/lib/permissions/views.py
szok: But it remains only to add mixin to class  that have meta class.

I can not add by meta class Python mixin it does not support it.

Exception:

```
TypeError: Cannot create a consistent method resolution
order (MRO) for bases PermissionViewMixin, RalphDetailViewAdmin
```

szok: Solution:

```
add_mixin = True
        for base_class in bases:
            if PermissionViewMixin in base_class.__mro__:
                add_mixin = False
                break

        if add_mixin:
            bases = (PermissionViewMixin,) + bases
```

But is not working correctly.
Any suggestion?

mkurek: when did you get this TypeError? Try sth like this:

```
class RalphDetailView(RalphExtraViewMixin, RalphTemplateView, metaclass= PermissionViewMetaClass):
...

PermissionAdminViewBase = type('PermissionAdminViewBase', (AdminViewBase, PermissionViewMetaClass), {})
class RalphDetailViewAdmin(RalphDetailView, metaclass=PermissionAdminViewBase):
...
```

ar4s:
:ok:

src/ralph/lib/permissions/views.py
```diff
@@ -0,0 +1,76 @@
...
+            content_type=ct,
+            codename=codename,
+            defaults={'name': 'Can view {}'.format(class_view.__name__)}
+        )
+        current_permission.append(perm.id)
+
+    # Remove old permission codename views.
+    permission_ids = set(old_permission) - set(current_permission)
+    if permission_ids:
+        Permission.objects.filter(id__in=permission_ids).delete()
```
mkurek, line 86: please add some logging here (with warning level)

mkurek:
add this logging and it's :ok: for me

PR #1814, opened 2015-09-30
Domains contracts management

vi4m:
## TOC
- Added new Domain Management module for domains purchases tracking.
- Some love to the make flake target - clean stage before flake was _fucking_ slow, and in fact not needed, so I removed it. You can always do this in a explicit way(`make clean flake`). 

From now on, beautiful ascii art flake will show, when `make flake` passes succesfully.
## Domain management Q&A

Q: What services/env will be used for?
A: All domains will have Service/Env set to -> Domains Mangement Service / prod / test. Scrooge will take care of prices management just like "Virtualization Systems"

Q: Why Technical and Business owner are separate from Service's owners. 
A: Because they are just responsible for a domain.

Q: Why separation of Domain and Contracts?
A: 1. Domain will be reused by django-powerdns-*. 2. We want to maintain history in a clean way, and managing not only renewal actions, but other types as well, such as transfer costs (another contract you can add in the future)


src/ralph/admin/sitetrees.py
```diff
@@ -123,6 +123,17 @@ def section(section_name, app, model):
...
         ),
         ralph_item(
+            title=_('Domains'),
+            url='#',
+            url_as_pattern=False,
+            perms_mode_all=False,
+            children=[
+                section(_('Domain contracts'), 'domains', 'DomainContract'),
+                section(_('Domain registrants'), 'domains', 'DomainRegistrant'),
+                section(_('Domain owners'), 'accounts', 'RalphUser'),
```
mkurek: is this a mistake? what is the purpose of this menu entry, since we have 'Users' already?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
```
mkurek: since there is only one file with only few models you could use `models.py` directly instead of creating models directory. We were using models directory iff there were lot of models in single app.

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,80 @@
...
+        verbose_name=_('Domain name'),
+        help_text=_('Full domain name'),
+        unique=True,
+        max_length=255
+    )
+    domain_status = models.PositiveIntegerField(
+        choices=DomainStatus(),
+        default=DomainStatus.active.id,
+    )
+    service = models.ManyToManyField(Service)
```
mkurek: consider, if it should point to `ServiceEnvironment` model (`ralph.assets.models.ServiceEnvironment`) - then it'll store information about environment either (at the same time).

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,80 @@
...
+    expiration_date = models.DateField(null=True, blank=True)
```
mkurek: what you think about storing history of expiration dates (and maybe additional info too, for example some contract id, price etc)?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,80 @@
...
+    business_segment = models.ForeignKey(BusinessSegment)
```
mkurek: business segment is defined indirectly by service (service -> profit center -> business segment)

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,80 @@
...
+    _ = Choices.Choice
+
+    active = _('Active')
+    pending_lapse = _('Pending lapse')
+    pending_transfer = _('Pending transfer away')
+    lapsed = _('Lapsed (inactive)')
+    transfered_away = _('Transfered away')
+
+
+class DomainContract(
```
mkurek: do we want to store only public domains? If not, then what about distinction between public and private domains? Or this would be stored in django_powerdns_dnssec and related?

src/ralph/admin/sitetrees.py
vi4m: There are multiple problems with the "Users" sitting somewhere in menu settings. 
1. "Don't let me think". We don't want the users to think about what's underlaying model for Domain owner field (User) and where to find it in settings. The fact that we use the same object is just a technical coincidence IMO. 
2. The fact, that you have to create ralph account(with password) just to assign "Domain owner" is a problem too, but this PR doesn't touch it

src/ralph/domains/models/domains.py
vi4m: No, it's revelant only to service as business wanted.

src/ralph/domains/models/domains.py
vi4m: It could be many services assigned to one domain contract, but only one bussines segment is the owner, that's why it's here "again".

src/ralph/admin/sitetrees.py
mkurek: if you mean Domain Driven Design, then not every user is domain owner (i know it could be, but only specific subset of users actually would be owning some domains), so IMO we should filter this view to show only real domain owners. What do you think?

src/ralph/domains/models/domains.py
mkurek: i don't know if there are any plans to charge for domains in scrooge, but if there will be such in future then it'll be impossible in current model, since scrooge requires both service and environment.

src/ralph/domains/models/domains.py
mkurek: ok, so shouldn't there be validation if selected services are descendants of selected business segment?

src/ralph/domains/models/domains.py
vi4m: It depends on searching/reporting, it has to be clarified if users prefers explicit dates management, or django history will be sufficient for auditable purposes. Has to be decided after end users tests and expectations.

src/ralph/admin/sitetrees.py
vi4m: It could be great to have either subclass of User for custom cases, or other "filter" like mechanism. I'm not sure what's better.

src/ralph/domains/models/domains.py
vi4m: Good point, if this PR ends up so simple, i will just live one file for models.

src/ralph/domains/models/domains.py
vi4m: I'm not sure if the reality will be so simple.TBD further. (we have 2k records to import with history baggage included and ownerships are not so simple in reality ;-))

src/ralph/admin/sitetrees.py
mkurek: filter (consider using django proxy model with overwritten db manager (objects): https://docs.djangoproject.com/en/1.8/topics/db/models/#proxy-models)

src/ralph/domains/filters.py
```diff
@@ -0,0 +1,16 @@
+# # -*- coding: utf-8 -*-
```
mkurek: when https://github.com/allegro/ralph/pull/1807 will be merged, these filters won't be necessary

Makefile
```diff
@@ -35,7 +35,8 @@ isort:
 test: clean
 	test_ralph test $(TEST)
 
-flake: clean isort
+flake: isort
+	@cat scripts/flake.txt
```
mkurek: what about this file? why didn't you commited it?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    technical_owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='domaincontract_technical_owner',
+        blank=True,
+        null=True,
+        help_text=_("Technical owner of a service")
+    )
+    domain_holder = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
```
mkurek: this should point to `AssetHolder` model

src/ralph/admin/sitetrees.py
vi4m: I just left Users table without filters for now, to not complicate things.

src/ralph/domains/models/domains.py
vi4m: Fixed.

src/ralph/domains/models/domains.py
vi4m: See the notes in the description field.

src/ralph/domains/models/domains.py
vi4m: django_powerdns_dnssec

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@register(DomainContract)
+class DomainContract(RalphAdmin):
+    list_select_related = ['domain', 'domain__service_env']
+    list_filter = [
+        'domain__name', 'domain__service_env__service__name'
+    ]
+    list_display = [
+        'domain', 'expiration_date', 'price'
```
mkurek, line 54: what about service? if we allow to filter by service name shouldn't it be one of the list columns too?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+        verbose_name=_('Domain name'),
+        help_text=_('Full domain name'),
+        unique=True,
+        max_length=255
+    )
+    domain_status = models.PositiveIntegerField(
+        choices=DomainStatus(),
+        default=DomainStatus.active.id,
+    )
+    business_segment = models.ForeignKey(
```
mkurek: i don't remember if i've asked it before but what do you think about many-to-many here? will there be one (and only one) contact person for domain? In case of services we allow for many owners.

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+    pass
+
+
+class DomainStatus(Choices):
+    _ = Choices.Choice
+
+    active = _('Active')
+    pending_lapse = _('Pending lapse')
+    pending_transfer = _('Pending transfer away')
+    lapsed = _('Lapsed (inactive)')
```
mkurek, line 34: maybe `expired` will better suit here?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+    domain_status = models.PositiveIntegerField(
+        choices=DomainStatus(),
+        default=DomainStatus.active.id,
+    )
+    business_segment = models.ForeignKey(
+        BusinessSegment, blank=True, null=True,
+        help_text=_("Business segment for a domain")
+    )
+    business_owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
```
mkurek, line 54: i don't remember if i've asked it before but what do you think about many-to-many here? will there be one (and only one) contact person for domain? In case of services we allow for many owners.

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+    domain_holder = models.ForeignKey(
+        AssetHolder,
+        blank=True,
+        null=True,
+        help_text=_("Company which receives invoice for the domain")
+    )
+
+    def __str__(self):
+        return "{name}".format(
```
mkurek: just `return self.name` will be enough :)

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+    ]
+    fieldsets = (
+        (_('Basic info'), {
+            'fields': (
+                'name', 'parent', 'remarks', 'domain_status',
+            )
+        }),
+        (_('Ownership info'), {
+            'fields': (
+                'service_env', 'domain_status',
```
mkurek: you have duplicated `domain_status` field

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class DomainContractInline(RalphTabularInline):
+    model = DomainContract
+    extra = 0
+
+
+@register(Domain)
+class DomainAdmin(RalphAdmin):
+    pass
```
mkurek: remove this `pass`

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+
+    list_select_related = ['technical_owner', 'business_owner', 'domain_holder']
+    list_filter = [
+        'name', 'service_env',
+    ]
+    list_display = [
```
mkurek: what about service_env?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+            name=self.name,
+        )
+
+
+class DomainContract(
```
mkurek, line 78: what about some asset-like fields like invoice date, invoice number, order number, task url etc?

src/ralph/domains/models/domains.py
```diff
@@ -0,0 +1,94 @@
...
+    _ = Choices.Choice
+
+    active = _('Active')
+    pending_lapse = _('Pending lapse')
+    pending_transfer = _('Pending transfer away')
+    lapsed = _('Lapsed (inactive)')
+    transfered_away = _('Transfered away')
+
+
+class Domain(BaseObject):
```
mkurek: mix `AdminAbsoluteUrlMixin` here (`ralph.lib.mixins.models.AdminAbsoluteUrlMixin`)

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+)
+
+
+class DomainContractInline(RalphTabularInline):
+    model = DomainContract
+    extra = 0
+
+
+@register(Domain)
+class DomainAdmin(RalphAdmin):
```
mkurek: don't we want to use attachments here? if yes, mix `AttachmentsMixin` (`ralph.attachments.admin.AttachmentsMixin`)

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,75 @@
...
+                'technical_owner', 'domain_holder'
+            )
+        })
+    )
+    search_fields = ['name', ]
+    inlines = (DomainContractInline, )
+
+
+@register(DomainContract)
+class DomainContract(RalphAdmin):
```
mkurek: consider attachments here too

src/ralph/domains/models/domains.py
vi4m, line 34: NO. "lapsed" is used in domain management dictionary. Expired is the process "in the middle"

src/ralph/domains/models/domains.py
vi4m, line 54: We don't need this, since it's the person just for a single domain management.

src/ralph/domains/models/domains.py
vi4m, line 78: It's automatically handled by auto-renewal process of CNC. If we want it in the future, we will just extend the model.

src/ralph/domains/admin.py
vi4m: good point :)

src/ralph/domains/admin.py
vi4m, line 54: service name search is still present for open source in order to use this field, but table view doesn't show that because in our case it will show "Domain Management Service" for all records.

mkurek:
:ok:

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class DomainContractInline(RalphTabularInline):
+    model = DomainContract
+    extra = 0
+
+
+@register(Domain)
+class DomainAdmin(AttachmentsMixin, RalphAdmin):
+    change_views = []
```
ar4s: pls remove - https://github.com/allegro/ralph/pull/1944

src/ralph/domains/admin.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+        })
+    )
+    search_fields = ['name', ]
+    inlines = (DomainContractInline, )
+
+
+@register(DomainContract)
+class DomainContractAdmin(AttachmentsMixin, RalphAdmin):
+    change_views = []
```
ar4s: ditto

src/ralph/domains/admin.py
vi4m: removed

src/ralph/domains/admin.py
vi4m: removed

ar4s:
:ok:

PR #1815, opened 2015-10-05
Split AssetStatus for separated statuses for BO Asset and DC Asset

mkurek:
- removed AssetStatus choices
- added new choices class for BackOfficeAsset status (BackOfficeAssetStatus)
- added new choices class for DataCenterAsset status (DataCenterAssetStatus)

Connected to #1665

PR #1816, opened 2015-10-05
Fixed exporting asset type and status

mkurek:
In Ralph NG, there is separated asset status for BO and DC (introduced in #1815). This PR fixes export of BO and DC asset statuses to be compatible with NG.

### DIFF
--- a/src/ralph/export_to_ng/resources.py
+++ b/src/ralph/export_to_ng/resources.py
@@ -12,6 +12,7 @@ from django.conf import settings
 from django.db.models import Q
 from import_export import fields
 from import_export import resources
+from lck.django.choices import Choices
 
 from ralph.account.models import Region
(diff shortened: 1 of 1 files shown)


szok:
:ok:

vi4m:
Your commit message: `This PR fix export` -> `This PR fixes export`

mkurek:
@vi4m done

vi4m:
It should be done with separate package called ralph-interop with DI. Instead, we did just a bastard child of two systems, mixing them, and breaking Classes Encapsulation rules . This horrible practice should be punished, and this pull request should not be merged. But...

Fuck it. New world is coming, let's spend our energy on Ralph-NG instead! :)

vi4m:
LLCBDSOTOT
Looks Like Crap But Don't Spend Our Time On It

Issue #1809, opened 2015-09-23
Bulk add/edit


damian1976:
1/ BULK ADD
To definetely close the issue of bulk add it would great to add also Inventory number textarea next to SN, barcode and Position ones. In this way we will be able (without multiple asset edit process) to add multiple assets in a single step. BUT I'd like the Inventory number to be optional. 
Example.
4 assets. All fo them have SN, barcodes and positions. But only 1st and 4th asset has Inventory number (second and third line in the Inventory Number textarea is empty). Validation should let the things go with these lines to be empty.

2/ BULK EDIT

Like in 2.x, my proposition is to add another option like Bulk location edit  in /data_center/datacenterasset/ in Actions form OR in existing Bulk edit option, when clicked, TWO additional columns should be added: Rack  and Position (probably also Slot is to be considered ?)

What do you think about it? In my opinion this funcionality will be very helpful and intuitive.

mkurek:
@szok would you like to take care of this? As far as i remember second issue has been already resolved.

damian1976:
Yes, 2nd issue is resolved. We may think of adding just one more column: 'Slot'. Just in case of e.g. blade assets. Them it will work for all cases. Can we add this column, please?

mkurek:
1 is resolved too (you could select bulk add fields using MULTIADD_DATA_CENTER_ASSET_FIELDS settings.

PR #1819, opened 2015-10-05
Multiadd / duplicate improvements

mkurek:
- fixed delimiting by pipe (|)
- added rack in dc asset description table
- improved saved message - now shows items count and links to added objects
- added ability to make field non-required
- using 'one_of_mulitvalue_required' in MultiAddView
- added 'niw' field to DC Asset multiadd

Connected to #1809

### DIFF
--- a/src/ralph/admin/fields.py
+++ b/src/ralph/admin/fields.py
@@ -16,13 +16,14 @@ class MultilineField(forms.CharField):
         - separated values cannot duplicate each other,
     """
     widget = forms.Textarea
-    separators = ",|\n"
+    separators = r',|\n|\|'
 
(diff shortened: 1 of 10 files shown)


szok:
:ok:

vi4m:
What about the "position" field ? Is it sill required? Should be optional IMHO.

mkurek:
Yes it is - it's required both in Ralph 2.0 (assets) and NG, just like Rack field.

vi4m:
Why?

mkurek:
As discussed, leave position field right now as it is - we can always change it to not being required.

ar4s:
:ok:

Issue #1512, opened 2015-06-23
Tagging mechanism


mkurek:
As an user I want to specify tags (values-only) for some objects. Tags should be working for models:
- [x] Base object and descendants (Asset, Tenant etc)
- [ ] Licence
- [ ] Support
- [ ] Service

...

 (and).

Requirements:
- tags should be accessible in API
- django admin integration by-default (single field for all tags, separated by comma)

For not-MVP version:
- tags hints during typing
- user should be able to filter objects by tags values in API
- user should be able to search by tags in GUI

PR #1820, opened 2015-10-06
Added tagging mechanism

mkurek:
- used django-taggit as tagging mechanism
- used django-taggit-serializer for cooperation with DRF
- tags are automatically shown in django admin and API (if model inherits `ralph.lib.mixins.models.TaggableMixin`).

Connected to #1512


ar4s:
:ok:

szok:
:ok:

PR #1821, opened 2015-10-06
Fixed help text (tooltip) position

mkurek:
In 4ef3e08 help text was (wrongly) moved after label. This commit moves it back after field.


ar4s:
:ok:

szok:
:ok:

PR #1823, opened 2015-10-06
list_filter attribute can be string or tuple.

msuszko:
Extract name in the second case, like it's done in
django.contrib.admin.views.main.ChangeList.get_filters


mkurek:
:ok:

szok:
:ok:

PR #1826, opened 2015-10-07
Remove python-ldap from requirements

szok:
django-auth-ldap in version 1.2.7 requires pyldap (2.4.20), which is implementation of python-ldap (http://www.python-ldap.org/index.html) in python3. Branch py3 was removed from python-ldap repo. See https://github.com/rbarrois/python-ldap/tree/release for details


mkurek:
:ok:

mkurek:
Please describe why you've changed it (about django-auth-ldap and python-ldap -> pyldap)

vi4m:
LGTM

PR #1830, opened 2015-10-19
Fixed LDAP groups sync

mkurek:
Added casting bytes to str when checking user groups (pyldap returns all values as bytes)


ar4s:
:ok:

szok:
:ok:

PR #1833, opened 2015-10-21
DC assets 'Bulk edit': rack, orientation and position fields added to the form

damian1976:
When updating location of multiple items (hardware/assets) in 'Bulk edit' mode it's very convenient (desirable) to be able to do it in one, single step. Otherwise, I have to, one-by-one, edit every single DC asset and change the location respectively. This PR makes the case easy and allows to do it in one step.


szok:
:ok:

mkurek:
:ok:

Issue #1690, opened 2015-08-11
Bulk edit "fill all" for FK fields (and other fill all improvements)


mkurek:
- [x] fill all should also work for FK fields (with autocomplete widget)
- [x] fill all button should be displayed on hover (at element), not after focus (after click)
- [x] when message is displayed at the top (ex. X assets saved successfully), "fill all" button position is messed up (it should be positioned relatively to input/td, not whole page)
- [x] there should be possibility to turn off "fill all" for some (defined) fields, ex. barcode, sn

mkurek:
Done in #1835

PR #1835, opened 2015-10-21
Bulk edit improvements

szok:
Added:
- more fields to Back Office asset
- exclude fill fields in bulk edit

Connected to #1690

### DIFF
--- a/src/ralph/admin/mixins.py
+++ b/src/ralph/admin/mixins.py
@@ -120,6 +120,8 @@ class RalphAdminMixin(RalphAutocompleteMixin):
 
     checks_class = RalphAdminChecks
     form = RalphAdminForm
+    # List of fields that are to be excluded from fillable on bulk edit
+    bulk_edit_no_fillable = []
 
     def __init__(self, *args, **kwargs):
         self.list_views = copy(self.list_views) or []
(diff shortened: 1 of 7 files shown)


src/ralph/data_center/admin.py
```diff
@@ -117,6 +117,7 @@ class DataCenterAssetAdmin(
     multiadd_info_fields = list_display + ['rack']
     one_of_mulitvalue_required = ['sn', 'barcode']
     bulk_edit_list = list_display
+    bulk_edit_list_exclude = ['barcode', 'sn']
```
mkurek: so exclude or no_fillable?

src/ralph/admin/mixins.py
```diff
@@ -203,6 +203,19 @@ def formfield_for_manytomany(self, db_field, request=None, **kwargs):
             return db_field.formfield(**kwargs)
         return super().formfield_for_manytomany(db_field, request, **kwargs)
 
+    def get_changelist_form(self, request, **kwargs):
```
mkurek, line 208: could you add some docstring here (what this method do, why this weird class `vTextField no-fillable` etc)?

szok, line 208: Done :)

mkurek, line 208: i mean description of bulk_edit_no_fillable option (you could btw add empty list as default value in this class and describe it there)

szok, line 208: OK, done

ar4s:
:ok:

vi4m:
Nicely done!

vi4m:
LGTM

PR #1839, opened 2015-10-22
fixed upgrading docker containers using init script

mkurek:
mute error during creating superuser (for example when user already exists) - if superuser was not created, error will be thrown in createsuperuser.py either way;


szok:
:ok:

ar4s:
:ok:

PR #1841, opened 2015-10-22
Add backoffice filter from old ralph

szok:
- Added missing filter to backoffice hardware
- Added support foreign key field in filter
- Added AssetHolder model
- Added property of field to BackOfficeAsset

### DIFF
--- a/src/ralph/admin/filters.py
+++ b/src/ralph/admin/filters.py
@@ -5,11 +5,15 @@ from functools import lru_cache
 
 from django.contrib.admin.filters import FieldListFilter
 from django.contrib.admin.utils import get_model_from_relation
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models import Q
+from django.forms.utils import flatatt
 from django.utils.encoding import smart_text
 from django.utils.formats import get_format
 from django.utils.translation import ugettext_lazy as _
+
+from ralph.admin.autocomplete import DETAIL_PARAM, QUERY_PARAM
 
 
 @lru_cache()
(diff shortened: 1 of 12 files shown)


src/ralph/assets/models/assets.py
```diff
@@ -150,7 +150,7 @@ class Meta:
         verbose_name_plural = _('models')
 
     def __str__(self):
-        if self.category:
+        if self.category_id:
```
mkurek, line 153: :+1:

src/ralph/back_office/models.py
```diff
@@ -27,6 +27,10 @@ class Warehouse(NamedMixin, TimeStampMixin, models.Model):
     pass
 
 
+class AssetOwner(NamedMixin.NonUnique, TimeStampMixin, models.Model):
```
mkurek: it is confusing with user/owner of the asset - IMO we should rename it. @vi4m any idea how to name it?

ar4s:
:ok:

src/ralph/admin/filters.py
```diff
@@ -205,14 +209,62 @@ def choices(self, cl):
...
 
-    @property
-    def choices_list(self):
-        model = get_model_from_relation(self.field)
-        return [(i.id, str(i)) for i in model._default_manager.all()]
+    template = "admin/filters/related_filter.html"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.model_field = get_model_from_relation(self.field)
```
mkurek: wouldn't `field_model` be better here?

mkurek:
:ok:

PR #1843, opened 2015-10-23
Add select_related support to Polymorphic models

szok:
Adding support for select_related in the Django admin change lists for
models that use PolymorphicQuerySet


mkurek:
Couldn't you "somehow" use `polymorphic_select_related` just like in API (for `BaseObject`)? 

https://github.com/allegro/ralph/blob/ng/src/ralph/api/utils.py
https://github.com/allegro/ralph/blob/ng/src/ralph/lib/polymorphic/models.py#L59

mkurek:
:ok:

ar4s:
:ok:

PR #1845, opened 2015-10-23
Added attachments to BackOfficeAsset, Licence and Support

mkurek:
- added attachments to BackOfficeAsset, Licence and Support
- fixed attaching the same view to multiple admin sites (ImproperlyConfigured exception will be raised during application start if this happens again in the future)
- fixed serving attachments with space in the name

### DIFF
--- a/src/ralph/admin/mixins.py
+++ b/src/ralph/admin/mixins.py
@@ -130,6 +130,14 @@ class RalphAdminMixin(RalphAutocompleteMixin):
         else:
             self.change_views = copy(self.change_views) or []
         super().__init__(*args, **kwargs)
+        # create copy (subclass) of view class to assign unique
+        # class to each admin when it's used (prevents conflicts between urls
+        # etc). See ralph.admin.extra.RalphExtraViewMixin.post_register for
+        # details
+        self.list_views = [type(c.__name__, (c,), {}) for c in self.list_views]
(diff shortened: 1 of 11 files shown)


src/ralph/admin/mixins.py
```diff
@@ -128,6 +128,14 @@ def __init__(self, *args, **kwargs):
         else:
             self.change_views = copy(self.change_views) or []
         super().__init__(*args, **kwargs)
+        # create copy (subclass) of view class to assign unique
```
vi4m, line 133: if you need to comment it like that - you probably want to have a separate sub-function with nice name `copy_class_list` or something

src/ralph/attachments/views.py
```diff
@@ -7,6 +7,19 @@
 from ralph.helpers import add_request_to_form
 
 
+def get_attachment_url_name(model, with_namespace=True):
```
szok: We already have a similar method in the MulitiAddAdminMixin(get_url_name) can write generic function?

szok:
:ok:

ar4s:
:ok:
https://www.youtube.com/watch?v=xqfL6_6qEJY

ar4s:
https://www.youtube.com/watch?v=xqfL6_6qEJY

PR #1846, opened 2015-10-27
There are fileds without flatchoices property

msuszko:


src/ralph/admin/filters.py
```diff
@@ -41,8 +41,11 @@ class BaseCustomFilter(FieldListFilter):
     def __init__(self, field, request, params, model, model_admin, field_path):
         self.lookup_kwarg = field_path
         self.model = model
-        if field.flatchoices:
-            self.choices_list = field.flatchoices
+        try:
+            if field.flatchoices:
+                self.choices_list = field.flatchoices
+        except AttributeError:
+            pass
```
vi4m, line 52: Put explanation in comment here, why you mute this exception.

msuszko, line 52: .flatchoices is property of `django.db.models.Field`, but `django.db.models.fields.related.ForeignObjectRel` doesn't have it.
This check breaks filters on ForeignKey fields.

mkurek, line 52: @vi4m probably meant to add this as a comment to source code and / or commit message

either way, if you add it, it's :ok: for me

msuszko:
updated in #1954

PR #1848, opened 2015-10-28
Added reports from Ralph 2

szok:
- Added asset relation report
- Added licence relation report

### DIFF
--- a/src/ralph/reports/views.py
+++ b/src/ralph/reports/views.py
@@ -1,15 +1,20 @@
 # -*- coding: utf-8 -*-
 import logging
 
+import tablib
+from django.contrib.contenttypes.models import ContentType
 from django.db.models import Count
-from django.http import Http404
+from django.http import Http404, HttpResponse
+from django.utils.encoding import smart_str
 from django.utils.translation import ugettext_lazy as _
 
+from ralph.admin.helpers import getattr_dunder
 from ralph.admin.mixins import RalphTemplateView
 from ralph.assets.models.assets import Asset, AssetModel
 from ralph.assets.models.choices import ObjectModelType
 from ralph.back_office.models import BackOfficeAsset
 from ralph.data_center.models.physical import DataCenter, DataCenterAsset
+from ralph.licences.models import Licence
 from ralph.reports.base import ReportContainer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


vi4m:
Nice html explanation needed, like in present reports.

src/ralph/reports/views.py
```diff
@@ -173,6 +232,106 @@ def prepare(self, model, dc=None):
             )
 
 
+class BaseRelationsReport(BaseReport):
+    template_name = 'reports/report_relations.html'
+    with_modes = True
+    links = False
+
+    def prepare(self, model, *args, **kwargs):
```
mkurek: do we need two functions for this: `prepare` and `get_result`?

src/ralph/reports/views.py
```diff
@@ -173,6 +232,106 @@ def prepare(self, model, dc=None):
...
+        pass
+
+    def get_response(self, request, result):
```
mkurek: what do you think about extracting it to separated class (`CSVReportMixin`)?

src/ralph/reports/views.py
```diff
@@ -173,6 +232,106 @@ def prepare(self, model, dc=None):
...
+    slug = 'asset-relations'
+    name = _('Asset - relations')
+    description = _('Asset list of information about the user, owner, model.')
+    filename = 'asset_relations.csv'
+    headers = ASSETS_COLUMNS
+
+    def get_assets_rows(self, model):
+        queryset = model.objects.all()
+        for asset in queryset:
+            row = [getattr_dunder(asset, column) for column in ASSETS_COLUMNS]
```
mkurek: how many SQLs are executed during generating this report? wouldn't it be better to pass these fields to `values_list` and do everything on SQL level?

src/ralph/reports/views.py
```diff
@@ -173,6 +232,106 @@ def prepare(self, model, dc=None):
...
+
+
+class AssetRelationsReport(BaseRelationsReport):
+    slug = 'asset-relations'
+    name = _('Asset - relations')
+    description = _('Asset list of information about the user, owner, model.')
+    filename = 'asset_relations.csv'
+    headers = ASSETS_COLUMNS
+
+    def get_assets_rows(self, model):
```
mkurek: there should be no option to generate report for ALL assets - it's useless; just maintain two separated reports for BO and DC and create separated columns for BO and DC.

src/ralph/reports/views.py
szok: :+1:

mkurek:
:ok:

src/ralph/reports/views.py
```diff
@@ -26,6 +31,33 @@ def get_choice_name(choices_class, key, default='------'):
         return 'Does not exist for key {}'.format(key)
 
 
+class CSVReportMixin(object):
+    """CSV report mixxin.
```
ar4s: mixxin -> mixin

ar4s:
:ok:

PR #1851, opened 2015-10-29
Add support and licences filter from old Ralph

szok:
- Added Support filters
- Added Licence filters
- Added new Decimal filter
- Rename Support SoftwareCategory model to Software
- Added asset_type to Support Software model

### DIFF
--- a/src/ralph/admin/filters.py
+++ b/src/ralph/admin/filters.py
@@ -187,6 +187,41 @@ class DateListFilter(BaseCustomFilter):
         }
 
 
+class NumberListFilter(DateListFilter):
+
+    """Renders filter form with decimal field."""
+
+    template = 'admin/filters/number_filter.html'
+
+    def queryset(self, request, queryset):
+        if any(self.value()):
+            value = self.value()
+            if value[0]:
+                queryset = queryset.filter(**{
+                    '{}__gte'.format(self.field_path): value[0]
+                })
+            if value[1]:
+                queryset = queryset.filter(**{
+                    '{}__lte'.format(self.field_path): value[1]
+                })
+        return queryset
+
+    def choices(self, cl):
+        value = self.value()
+        # default is decimal
+        step = 0.01
+        if isinstance(self.field, models.IntegerField):
+            step = 1
+
+        yield {
+            'parameter_name_start': self.parameter_name_start,
+            'parameter_name_end': self.parameter_name_end,
+            'start_value': value[0],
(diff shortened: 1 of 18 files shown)


src/ralph/admin/filters.py
```diff
@@ -275,6 +305,7 @@ def register_custom_filters():
     """
     field_filter_mapper = [
         (lambda f: bool(f.choices), ChoicesListFilter),
+        (lambda f: isinstance(f, models.DecimalField), DecimalListFilter),
```
mkurek: what about integer field?

src/ralph/admin/filters.py
```diff
@@ -187,6 +187,36 @@ def choices(self, cl):
...
+                    '{}__gte'.format(self.field_path): value[0]
+                })
+            if value[1]:
+                queryset = queryset.filter(**{
+                    '{}__lte'.format(self.field_path): value[1]
+                })
+
+            return queryset
+
+    def choices(self, cl):
```
mkurek, line 209: do we have to use choices here (and in date filter too)? I don't like this yielding-only-single-element.

src/ralph/admin/templates/admin/filters/decimal_filter.html
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+<h6>
+  {% blocktrans with filter_title=title %}
+    {{ filter_title }}
+  {% endblocktrans %}
+</h6>
+{% with choices.0 as i %}
+  <input type="text" name="{{ i.parameter_name_start }}" placeholder="From" value="{{ i.start_value }}" />
```
mkurek: what do you think about `<input type="number" step="any" ...>`?

src/ralph/licences/models.py
```diff
@@ -73,13 +74,14 @@ class Licence(
...
     )
-    # property_of = models.ForeignKey(
-    #     settings.AUTH_USER_MODEL,
-    #     on_delete=models.PROTECT,
-    #     null=True,
-    # )
-    software_category = models.ForeignKey(
-        SoftwareCategory,
+    property_of = models.ForeignKey(
+        'back_office.AssetHolder',
```
mkurek: wouldn't it be better to define `AssetHolder` in assets app? it'll be used in BO, DC, Domains and maybe somewhere else in the future

mkurek:
:ok:

ar4s:
:ok:

PR #1852, opened 2015-10-29
[WIP] added lightweight dependency form

ar4s:


src/ralph/static/src/js/autofill.js
```diff
@@ -0,0 +1,28 @@
+(function ($) {
+    'use strict';
+    var connections = [
```
szok, line 3: :+1:

src/ralph/admin/mixins.py
```diff
@@ -41,10 +41,12 @@ def get_common_media():
         ('admin', 'js', 'actions.js'),
         ('admin', 'js', 'admin', 'RelatedObjectLookups.js'),
         ('vendor', 'js', 'jquery.js'),
+        ('src', 'js', 'ralph.init.js'),
         ('vendor', 'js', 'foundation.min.js'),
         ('vendor', 'js', 'modernizr.js'),
-        ('src', 'js', 'fill-fields.js'),
         ('vendor', 'js', 'foundation-datepicker.js'),
+        # ('src', 'js', 'fill-fields.js'),
```
mkurek: what about it?

src/ralph/static/src/js/autofill.js
```diff
@@ -0,0 +1,27 @@
...
+        {
+            app: 'back_office', model: 'backofficeasset',
```
mkurek: wouldn't it be better to move model to next line?

src/ralph/static/src/js/autofill.js
```diff
@@ -0,0 +1,27 @@
...
+            conf: [
+                {
+                    src: 'user',
+                    func: function($form, $src, val) {
+                        $('[name=owner]').val(val);
```
mkurek: what if i have matching input outside this form? shouldn't scope of this function be limited only to form? in other words, shouldn't you filter only inside passed form?

mkurek:
:ok:

PR #1855, opened 2015-10-30
Added new action to transitions

ar4s:
alternative to https://github.com/allegro/ralph/pull/1852


src/ralph/back_office/models.py
```diff
@@ -195,3 +196,26 @@ def change_hostname(self, **kwargs):
             )
         }
     }
+
+    @transition_action
+    def change_user_and_owner(self, **kwargs):
+        UserModel = apps.get_model(*settings.AUTH_USER_MODEL.split('.'))
```
szok: Why do you not used : get_user_model() ?

src/ralph/lib/transitions/views.py
```diff
@@ -30,6 +30,9 @@ def get_form_fields_from_actions(self):
         for action in self.actions:
             action_fields = getattr(action, 'form_fields', {})
             for name, options in action_fields.items():
+                condition = options.get('condition', lambda x: True)
+                if not condition(self.obj):
+                    continue
```
szok, line 35: :+1: 

I think you have to added to documentation.

ghost:
:ok:

mkurek:
:ok:

Issue #1840, opened 2015-10-22
REP5: Operations


mkurek:
## Introduction

This is continuation of #1803 (CMDB Interface)

In Ralph 2.0 each CI could have attached incidents, problems and changed (from Jira). According to #1803 we want to move this functionality to Ralph NG. This REP will describe technical details of this feature.
## CMDB 2.0

In current CMDB version (Ralph 2.0) there are incidents, problems and changes, which use the same base class (but are stored in different database tables). Each of this models has exactly the same fields:
- cis (many to many for CI)
- summary
- description (don't know what's the difference between summary and description)
- jira_id 
- status
- assignee
- analysis (don't know what's the purpose of this field - IMO it's rather impact than analysis)
- problems (ditto)
- priority
- issue_type
- created / updated / resolved / planned_start / planned_end date
## Hardware failures

There is new feature request, to store (assets) hardware failures in Ralph NG (both for DC and BO assets). Each failure should have the following fields:
- failure type (for example hard drive, RAM, power supply etc)
- description
- jira (or probably general purpose) ticket id
- status 
## Proposed architecture

![img_20151022_143014](https://cloud.githubusercontent.com/assets/107437/10665179/91e8bd26-78c9-11e5-86ca-f4305f3cfb66.jpg)

Notes:
- `Issue` is generic type for problems, incidents, changes and failures.
- `Failure` is inheriting from `Issue`, adding one additional field: `failure_type`, which is foreign key to additional table `FailureType`
- `FailureType` contains name and `ContentType` - particular failure type could be assigned only to failure for one base object type, for example controller crash could be reported only for data center asset
- `Issue` could be attached to multiple `BaseObject`s (probably all with the same type)
- status has only few valid choices (use `dj.choices`), for example open, closed, resolved etc.
- choosing 'failure' issue type forces to use `Failure` type; using `Failure` class forcing to use `failure` type;
## Discussion
- is failure proper name for it?
- should failure inherit from issue?
- dependency and conditions between issue type and inheriting is a little bit complicated

vi4m:
Well I really like this approach in general.
I'm wondering if we really need separate Failure object.

What if we simplify that a little bi,  to just say that Failure = Incident, type=hardware incident, details=disk broken. You can than relate incidents together, for example nic card incident -> service not responding incident in the same place. I'm not sure how specific is Failure in comparision with Incident.

vi4m:
[...]

PR #1856, opened 2015-10-30
REP 5: Operation - implementation

mkurek:
Implementation of #1840 

Main feature of this commit is both-ends M2M in inlines (which is later used to create Operations directly from BaseObject and connect them together).
Other features:
- Operations hierarchy - incidents, problems, changes, failures
- Failures report by manufacturer

### DIFF
--- a/src/ralph/admin/mixins.py
+++ b/src/ralph/admin/mixins.py
@@ -87,6 +87,10 @@ class RalphAdminForm(RalphAdminFormMixin, forms.ModelForm):
     pass
 
 
+class RalphMPTTAdminForm(RalphAdminFormMixin, MPTTAdminForm):
+    pass
+
+
 class RalphAdminChecks(admin.checks.ModelAdminChecks):
     def _check_form(self, cls, model):
         """
@@ -104,10 +108,6 @@ class RalphAdminChecks(admin.checks.ModelAdminChecks):
                 id='admin.E016'
             )
         return result
-
-
-class RalphMPTTAdminForm(RalphAdminFormMixin, MPTTAdminForm):
-    pass
 
 
 class RalphAdminMixin(RalphAutocompleteMixin):
@@ -319,6 +319,10 @@ class RalphAdmin(
         return resource_class
 
 
+class RalphMPTTAdmin(MPTTModelAdmin, RalphAdmin):
+    form = RalphMPTTAdminForm
+
+
 class RalphTabularInline(
     RalphAutocompleteMixin,
     admin.TabularInline
@@ -347,10 +351,6 @@ class RalphTemplateView(
     RalphBaseTemplateView, metaclass=PermissionViewMetaClass
 ):
     pass
-
-
-class RalphMPTTAdmin(MPTTModelAdmin, RalphAdmin):
-    form = RalphMPTTAdminForm
 
 
 class BulkEditChangeListMixin(object):
--- a/src/ralph/api/tests/test_viewsets.py
+++ b/src/ralph/api/tests/test_viewsets.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from datetime import date, datetime
+from datetime import date
 from decimal import Decimal
 from urllib.parse import urlencode
 
@@ -21,7 +21,6 @@ from ralph.api.tests.api import (
 from ralph.api.viewsets import RalphAPIViewSet
 from ralph.tests import RalphTestCase
 from ralph.tests.factories import ManufacturerFactory
-from ralph.tests.models import Bar
 
 
 class ViewsetWithoutRalphPermission(RalphAPIViewSet):
@@ -50,21 +49,18 @@ class TestRalphViewset(RalphTestCase):
 
         Bar.objects.create(
             name='Bar11',
-            created=datetime(2015, 3, 1, 4, 30, 30),
             date=date(2015, 3, 1),
             price=Decimal('21.4'),
             count=1
         )
         Bar.objects.create(
             name='Bar22',
-            created=datetime(2014, 4, 1, 4, 30, 30),
             date=date(2014, 4, 1),
             price=Decimal('11.4'),
             count=2
         )
         Bar.objects.create(
             name='Bar33',
-            created=datetime(2013, 5, 1, 4, 30, 30),
             date=date(2013, 5, 1),
             price=Decimal('31.4'),
             count=3
@@ -166,12 +162,17 @@ class TestRalphViewset(RalphTestCase):
     def test_query_filters_datetimefield(self):
         request = self.request_factory.get('/api/bar')
         bvs = BarViewSet()
-        request.query_params = QueryDict(urlencode({'created__year': 2015}))
+        request.query_params = QueryDict(urlencode(
+            {'created__year': date.today().year})
+        )
         bvs.request = request
-        self.assertEqual(len(bvs.get_queryset()), 1)
+        self.assertEqual(len(bvs.get_queryset()), 3)
 
         request.query_params = QueryDict(
-            urlencode({'created__year': 2015, 'created__month': 3})
+            urlencode({
+                'date__year': 2014,
+                'created__month': date.today().month
+            })
         )
         bvs.request = request
         self.assertEqual(len(bvs.get_queryset()), 1)
--- a/src/ralph/data_importer/resources.py
+++ b/src/ralph/data_importer/resources.py
@@ -30,6 +30,7 @@ from ralph.licences.models import (
     LicenceUser,
     Software
 )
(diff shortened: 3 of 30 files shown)


src/ralph/admin/filters.py
```diff
@@ -244,15 +247,53 @@ def choices(self, cl):
...
+    select list (all options are fetched at once).
+    """
+    def label_for_instance(self, obj):
+        return smart_text(obj)
+
+    @property
+    def choices_list(self):
+        model = get_model_from_relation(self.field)
+        queryset = model._default_manager.all()
+        return [(x._get_pk_val(), self.label_for_instance(x)) for x in queryset]
```
vi4m: i

vi4m:
What's missing?

src/ralph/admin/mixins.py
```diff
@@ -305,11 +305,15 @@ def get_export_resource_class(self):
...
+class RalphMPTTAdmin(MPTTModelAdmin, RalphAdmin):
+    form = RalphMPTTAdminForm
+
+
 class RalphTabularInline(
     RalphAutocompleteMixin,
     admin.TabularInline
 ):
-    pass
+    view_on_site = False
```
mkurek: remove

src/ralph/operations/models.py
```diff
@@ -0,0 +1,145 @@
...
+
+class Failure(Operation):
+    objects = OperationDescendantManager(OperationType.choices.failure)
+
+    class Meta:
+        proxy = True
+
+
+@receiver(post_migrate)
+def rebuild_handler(sender, **kwargs):
```
ghost, line 140: what's the purpose of this?

src/ralph/api/tests/test_viewsets.py
```diff
@@ -166,12 +162,17 @@ def test_query_filters_datefield(self):
     def test_query_filters_datetimefield(self):
         request = self.request_factory.get('/api/bar')
         bvs = BarViewSet()
-        request.query_params = QueryDict(urlencode({'created__year': 2015}))
+        request.query_params = QueryDict(urlencode(
+            {'created__year': date.today().year})
```
szok, line 166: :+1:

src/ralph/operations/models.py
mkurek, line 140: in migration 0002 we set `{'lft': 0, 'rght': 0, 'level': 0, 'tree_id': 0}`, so after this migration (and possibly every other migration in the future which will add more types) we have to rebuild mptt tree.

src/ralph/operations/models.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+    status = models.PositiveIntegerField(
+        verbose_name=_('status'), choices=OperationStatus(),
+    )
+    asignee = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='operations',
+        verbose_name=_('asignee'),
+        null=True,
+        blank=True,
```
ghost, line 70: When `assignee` is deleted `operation` will be also removed, true?
I'm not sure if this is wanted.. . Or perhaps we don't allow removing users.

mkurek, line 70: You mean when we delete account from Ralph? As far as I remember we don't allow to remove user accounts. Anyway, I could add here `on_delete=models.PROTECT` if you like.

src/ralph/reports/views.py
```diff
@@ -453,3 +454,36 @@ def prepare(self, model, *args, **kwargs):
...
+    description = _('Failure types for each manufacturer.')
+
+    def prepare(self, model, dc=None):
+        queryset = model._default_manager
+        if dc:
+            queryset = queryset.filter(rack__server_room__data_center=dc)
+        operation_types = OperationType.objects.get(
+            pk=OperationType.choices.failure
+        ).get_descendants(include_self=True)
+        failures = Failure.base_objects.through.objects.filter(
```
xliiv: Couldn't `Failure` (https://github.com/allegro/ralph/pull/1856/files#diff-add77daee7adbac65288dc4367fd87b5R131) be used here?

src/ralph/data_importer/resources.py
```diff
@@ -515,3 +516,30 @@ class DomainResource(RalphModelResource):
 
     class Meta:
         model = Domain
+
+
+class OperationTypeResource(ImportForeignKeyMixin, resources.ModelResource):
+    name = fields.Field(
```
szok: Column name type not exist in OperationTYpe model.

src/ralph/reports/views.py
```diff
@@ -453,3 +454,36 @@ def prepare(self, model, *args, **kwargs):
...
+            baseobject__in=queryset.all(),
+            operation__type__in=operation_types
```
ghost, line 473: is this type required, since it's `Failure` manager?

src/ralph/data_importer/widgets.py
```diff
@@ -40,6 +40,10 @@ def get_imported_obj(model, old_pk):
                 content_type=content_type,
             )
         )
+
+        def imported_obj():
```
szok: to remove
Used now https://github.com/mkurek/ralph/blob/operations/src/ralph/data_importer/widgets.py#L164

src/ralph/reports/views.py
mkurek, line 473: it's complicated. Actually, we don't use `Failure`'s manager here - it's not `Failure`'s objects, but `base_objects.through` objects, which is M2M relationship intermediate table and we store there every `Operation` relation. Since there is M2M relation between `Failure` (`Operation`) and `BaseObject`, I didn't find a way to get data (manufacturer + number of failures/operation) directly from `Failure`.

src/ralph/data_importer/widgets.py
```diff
@@ -40,6 +40,10 @@ def get_imported_obj(model, old_pk):
...
+            pass
```
ghost: Could you write why this is empty?

mkurek: read @szok comment above - this is "old-way" to handle imported id/old id.

src/ralph/reports/views.py
```diff
@@ -453,3 +454,36 @@ def prepare(self, model, *args, **kwargs):
...
+        ).values(
```
szok, line 474: maybe add select_related for asset__model__manufacturer__name?

src/ralph/settings/base.py
```diff
@@ -270,3 +271,4 @@ def os_env_true(var, default=''):
 MY_EQUIPMENT_SHOW_BUYOUT_DATE = os_env_true('MY_EQUIPMENT_SHOW_BUYOUT_DATE')
 
 MAP_IMPORTED_ID_TO_NEW_ID = False
+ISSUE_TRACKER_URL = ''
```
ghost, line 281: I'd add a comment that it's related to `operation` feature.

src/ralph/reports/views.py
mkurek, line 474: no need for that since we call `values` on it - django will handle it in one query.

src/ralph/data_importer/resources.py
```diff
@@ -515,3 +516,30 @@ class DomainResource(RalphModelResource):
 
     class Meta:
         model = Domain
+
+
+class OperationTypeResource(ImportForeignKeyMixin, resources.ModelResource):
```
ghost: shouldn't this inherit from `RalphModelResource`?

src/ralph/data_importer/resources.py
```diff
@@ -515,3 +516,30 @@ class DomainResource(RalphModelResource):
...
+        column_name='parent',
+        attribute='parent',
+        widget=ImportedForeignKeyWidget(OperationType),
+    )
+
+    class Meta:
+        model = OperationType
+
+
+class OperationResource(ImportForeignKeyMixin, resources.ModelResource):
```
ghost: shouldn't this inherit from `RalphModelResource`?

mkurek: sure - it was written before `RalphModelResource` was introduced

src/ralph/operations/models.py
ghost, line 140: :+1:

src/ralph/operations/models.py
ghost, line 70: I'm not sure. If this is reasonable, I say yes. What do you think?

src/ralph/data_importer/widgets.py
ghost: ok

src/ralph/operations/models.py
mkurek, line 70: done

ar4s:
:ok:

ghost:
:ok:

PR #1860, opened 2015-11-02
Fixed nested serializers

mkurek:
- nested serializer is not registered as main model serializer
- exclude specific fields from nested serializers (ex. content_type)


szok:
:ok:

ghost:
:ok:

PR #1862, opened 2015-11-02
Use 64bit integer to store numeric address

msuszko:
PositiveIntegerField can store only 2^31 (127.255.255.255),
use BigIntegerField instead.


mkurek:
:ok:

szok:
:ok:

PR #1863, opened 2015-11-02
Add service env filter to datacenter asset api filter

szok:
Added service_env__service__uid to datacenter asset api filter


mkurek:
:ok:

ar4s:
:ok:

PR #1864, opened 2015-11-02
Fix forms in admin tabs

mkurek:
- Fixes bug created in ee5893e
- Removed subclassing mechanism for all change views and list views
- Now subclass of extra views are created only for attachments


szok:
:ok:

ghost:
:ok:

PR #1875, opened 2015-11-04
rename back UserLicence to LicenceUser

mkurek:
- rename back UserLicence to LicenceUser
- export Licence's software_category as software


szok:
:ok:

ar4s:
:ok:

Issue #1778, opened 2015-09-10
There is no access control in Reports


msuszko:
Reports are working without any permissions granted

mkurek:
@szok: there should be separated permission for each report; additionally there should be permission for all modes (All/DC/BO or whatever there will be in future reports) for each report.

mkurek:
Done in #1877

PR #1877, opened 2015-11-04
Add permission view to reports

szok:
Change generating report mechanism.
Added permissions to reports view.

Connected to #1778


ar4s:
:ok:

mkurek:
:ok:

PR #1883, opened 2015-11-06
user only need active LDAP permission (group) to log-in to Ralph

mkurek:
active permission (group) automatically assign staff permission


szok:
:ok:

ar4s:
:ok:

PR #1889, opened 2015-11-06
Add extend filter to REST API

szok:
Added extend_filter_fields to rest framework api

### DIFF
--- a/src/ralph/api/viewsets.py
+++ b/src/ralph/api/viewsets.py
@@ -2,6 +2,7 @@
 import inspect
 
 from django.contrib.admin import SimpleListFilter
+from django.db.models import Q
 from rest_framework import filters, permissions, relations, viewsets
 
 from ralph.admin.sites import ralph_site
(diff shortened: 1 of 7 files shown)


src/ralph/api/viewsets.py
```diff
@@ -116,4 +116,15 @@ class RalphAPIViewSet(
...
-    pass
+    extend_filter_fields = {}
+
+    def get_queryset(self):
+        queryset = self.queryset
+        for field, field_filters in self.extend_filter_fields.items():
+            value = self.request.query_params.get(field, None)
+            if value:
+                queryset = queryset.filter(
+                    **dict.fromkeys(field_filters, value)
```
mkurek: shouldn't this use `Q` to make or-condition at the level of single param?

src/ralph/api/viewsets.py
```diff
@@ -116,4 +119,15 @@ class RalphAPIViewSet(
...
-    pass
+    extend_filter_fields = {}
+
+    def get_queryset(self):
+        queryset = self.queryset
+        qs = []
+        for field, field_filters in self.extend_filter_fields.items():
+            value = self.request.query_params.get(field, None)
+            if value:
+                qs.append(Q(**dict.fromkeys(field_filters, value)))
```
mkurek: use AND between fields (meta-fields) and OR within one field

src/ralph/api/viewsets.py
```diff
@@ -116,4 +117,15 @@ class RalphAPIViewSet(
     viewsets.ModelViewSet,
     metaclass=RalphAPIViewSetMetaclass
 ):
-    pass
+    extend_filter_fields = {}
```
mkurek: this should probably be done in `__init__`

src/ralph/api/viewsets.py
```diff
@@ -116,4 +117,20 @@ class RalphAPIViewSet(
...
-    pass
+    extend_filter_fields = None
+
+    def __init__(self, *args, **kwargs):
+        if self.extend_filter_fields is None:
+            self.extend_filter_fields = {}
+        super().__init__(*args, **kwargs)
+
+    def get_queryset(self):
+        queryset = self.queryset
```
mkurek: why not super?

mkurek:
uni tests would be nice here :)

mkurek:
:ok:

ar4s:
:ok:

PR #1890, opened 2015-11-07
Docker image & DEB build in Travis

mkurek:
There is new phase in Travis - deploy. It's executed only for non-PR builds (and with our conditions only on commit to branch ng and when python 3.4 build is performed).
During deploy, there are two (independent) steps: processing DEB package and Docker image.

Version format for both packages: `<RALPH_VERSION>-snapshot-<DATE %Y%m%d>-<TRAVIS_BUILD_NUMBER>`, ex. `3.0.0-snapshot-20151106-70`

DEB package processing steps:
- installation of dh-virtualenv
- deb package build (using `make build-package`)
- upload of deb package to bintray (using curl)

Docker image processing steps:
- login to Docker Hub
- build of new image
- push image to docker hub (using version number specified above and additional `snapshot_latest` tag)

Environment variables that need to be set in travis:
- `DOCKER_IMAGE`
- `DOCKER_EMAIL`
- `DOCKER_USERNAME`
- `DOCKER_PASSWORD`
- `BINTRAY_USER`
- `BINTRAY_APIKEY`
- `BINTRAY_REPO_NAME`
- `BINTRAY_PACKAGE_NAME`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
-sudo: false
+sudo: required
 language: python
+
+services:
+    - docker
 
 cache:
   directories:
@@ -13,6 +16,8 @@ python:
 before_install:
     - pip install flake8 isort
     - make flake
+    - export RALPH_VERSION=$(cat VERSION)
+    - export NEW_TAG=$RALPH_VERSION-snapshot-$(date "+%Y%m%d")-$TRAVIS_BUILD_NUMBER
 
 install:
     - npm config set python /usr/bin/python2.7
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -29,6 +34,24 @@ script:
...
+    - sudo apt-get update
+    - sudo apt-get install -y devscripts python-virtualenv git equivs dh-virtualenv
+    - make build-package
+
+deploy:
+    -   provider: script
+        script: scripts/travis_deploy_docker.sh
+        on:
+            branch: ng
+            python: 3.4
```
vi4m, line 48: what about python 3.5 ?

mkurek, line 48: Docker image and DEB package are independent from python version on travis. Python version to use there should be specified either in Dockerfile or `debian/rules`. 

This condition is there only for one reason: to not deploy anything multiple times (in many builds on travis).

.travis.yml
```diff
@@ -13,6 +16,8 @@ python:
 before_install:
     - pip install flake8 isort
     - make flake
+    - export RALPH_VERSION=$(cat VERSION)
+    - export NEW_TAG=$RALPH_VERSION-snapshot-$(date "+%Y%m%d")-$TRAVIS_BUILD_NUMBER
```
vi4m, line 20: very nice.

scripts/travis_deploy_bintray.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
vi4m, line 1: Maybe add one line here to be more readable.

vi4m:
I **really** like it! _Finally_ travis supports docker to integrate continous building systems.

vi4m:
So now, what about other sub-components such as inkpy?

Probably we should do this for each dependent subrepository to create separate packages (deb + docker) in a very similar manner.

vi4m:
Ok, merging.

PR #1895, opened 2015-11-10
Service_env is not required for BackOffice asset

mkurek:
Overwrited BackOfficeAdmin form and turned off service_env requirement.


ar4s:
:ok:

szok:
:ok:

PR #1896, opened 2015-11-10
Licence relation report optimization

szok:


src/ralph/reports/views.py
```diff
@@ -360,38 +359,54 @@ class LicenceRelationsReport(BaseRelationsReport):
...
+    ]
 
     def prepare(self, model, *args, **kwargs):
         queryset = Licence.objects.all()
         if model._meta.object_name == 'BackOfficeAsset':
             queryset = queryset.filter(
-                base_objects__content_type=ContentType.objects.get_for_model(
-                    BackOfficeAsset
-                )
+                software__asset_type=ObjectModelType.back_office
```
mkurek: `software__asset_type__in=(ObjectModelType.back_office, ObjectModelType.all)`

src/ralph/reports/tests.py
```diff
@@ -177,6 +178,13 @@ def test_asset_relation(self):
         ]
         self.assertEqual(report_result, result)
 
+    def test_num_queries(self):
```
mkurek: could you make the same test for BackOffice?

mkurek: btw in this kind of test it'd be cool to provide sufficient amount of test data (ex. 1000 licences)

mkurek:
:ok:

src/ralph/reports/tests.py
```diff
@@ -155,11 +159,35 @@ def setUp(self):
...
+                niw='N/A' + str(i),
+                software__name=str(i)
+            )
+            BaseObjectLicence.objects.create(
+                licence=licence, base_object=bo.baseobject_ptr
+            )
+            LicenceUser.objects.create(
+                licence=licence,
+                user=user
+            )
```
ar4s: try `bulk_create`
https://docs.djangoproject.com/en/1.8/ref/models/querysets/#bulk-create

mkurek: :+1:

ar4s:
:ok:

PR #1898, opened 2015-11-10
New OfficeInfrastructure model

mkurek:
- OfficeInfrastructure is Service model from old ralph_assets (without profit center and cost center)
- added importer from old Ralph


szok:
:ok:

ar4s:
:ok:

PR #1899, opened 2015-11-10
Added new user view

ar4s:
User can view assigned equipment. The view is very similar to additional user
info but this view is strongly linked with logged user.
user_equipmentr

### DIFF
--- a/src/ralph/accounts/admin.py
+++ b/src/ralph/accounts/admin.py
@@ -2,6 +2,7 @@
 from django.contrib.auth.admin import GroupAdmin, UserAdmin
 from django.contrib.auth.models import Group
 from django.core.urlresolvers import reverse
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.accounts.models import RalphUser, Region, Team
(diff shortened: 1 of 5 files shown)


src/ralph/accounts/admin.py
```diff
@@ -61,19 +61,21 @@ def url(self, item):
...
+    def get_user(self):
+        if not self.user:
+            raise NotImplementedError('Please specify user.')
+        return self.user
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         asset_list_queryset = BackOfficeAsset.objects.filter(
-            user=self.object
+            user=self.get_user()
```
mkurek: here should be `Q(user=self.get_user()) | Q(owner=self.get_user())`

src/ralph/accounts/views.py
```diff
@@ -1,7 +1,28 @@
...
 class UserProfileView(RalphTemplateView):
-
     template_name = 'ralphuser/user_profile.html'
+
+
+class CurrentUserInfoView(UserInfoMixin, TemplateView):
+    template_name = 'ralphuser/my_equipment.html'
+
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.is_authenticated():
```
mkurek: wouldn't it be better to use https://docs.djangoproject.com/en/1.8/topics/auth/default/#the-login-required-decorator ?

mkurek:
:ok:

szok:
:ok:

PR #1902, opened 2015-11-10
Added Budget info to exporter

szok:


src/ralph/export_to_ng/resources.py
```diff
@@ -133,6 +135,13 @@ def dehydrate_imei(self, asset):
             imei = ''
         return imei
 
+    def dehydrate_office_infrastructure(self, asset):
+        try:
+            service = asset.service_id
```
vi4m: why not just `return getattr(asset, 'service_id', '')`

mkurek: first of all this field is named `service_name` in ralph assets. Second of all, `AttributeError` is not possible here (it will be if you'll call `asset.service.id`) - `asset.service_id` will be either number (>0) or `None`, so my suggestion is `service = asset.service_id or ''`, without any try-except

src/ralph/export_to_ng/resources.py
```diff
@@ -133,6 +135,9 @@ def dehydrate_imei(self, asset):
             imei = ''
         return imei
 
+    def dehydrate_office_infrastructure(self, asset):
+        return asset.service_id or ''
```
mkurek: `service_name_id`: https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/models_assets.py#L654

mkurek:
:ok:

ar4s:
:ok:

PR #1905, opened 2015-11-12
Session in Redis

szok:


requirements/base.txt
```diff
@@ -8,6 +8,7 @@ django-reversion==1.8.6
 django-taggit==0.17.1
 django-taggit-serializer==0.1.5
 djangorestframework==3.2.2
+django-redis-sessions==0.5.0
```
mkurek: this should be IMO in prod.txt

src/ralph/settings/base.py
```diff
@@ -180,6 +180,13 @@
     'DB': int(os.environ.get('REDIS_PORT_6379_TCP_DB', 0)),
 }
 
+SESSION_ENGINE = 'redis_sessions.session'
```
mkurek: ditto - in `prod.py` settings. @ar4s what do you think?

ar4s:
:ok:

mkurek:
:ok:

PR #1919, opened 2015-11-12
Add hostname_assigning signal to BackOfficeAsset model

szok:


src/ralph/back_office/models.py
```diff
@@ -297,3 +299,13 @@ def return_report(self, request, **kwargs):
...
     return_report.verbose_name = _('Return report')
+
+
+@receiver(pre_save, sender=BackOfficeAsset)
+def hostname_assigning(sender, instance, raw, using, **kwargs):
+    """A hook for assigning ``hostname`` value when an
+    back office asset is edited.
+    """
+    if getattr(settings, 'BACK_OFFICE_ASSET_AUTO_ASSIGN_HOSTNAME', None):
+        if instance.status == BackOfficeAssetStatus.in_progress:
```
mkurek, line 309: what if asset already has `in_progress` status? we don't want to change hostname in this case. You should fetch previous asset from db and check if status field has changed

src/ralph/back_office/models.py
```diff
@@ -296,3 +298,17 @@ def return_report(self, request, **kwargs):
         self._generate_report(name='return', request=request)
     return_report.return_attachment = True
     return_report.verbose_name = _('Return report')
+
+
+@receiver(pre_save, sender=BackOfficeAsset)
+def hostname_assigning(sender, instance, raw, using, **kwargs):
+    """A hook for assigning ``hostname`` value when an
+    back office asset is edited.
```
mkurek: change this comment - hostname is assigned for new assets with in_progress status or edited assets when status has changed to in_progress.

mkurek:
:ok:

src/ralph/back_office/models.py
```diff
@@ -296,3 +298,17 @@ def return_report(self, request, **kwargs):
         self._generate_report(name='return', request=request)
     return_report.return_attachment = True
     return_report.verbose_name = _('Return report')
+
+
+@receiver(pre_save, sender=BackOfficeAsset)
+def hostname_assigning(sender, instance, raw, using, **kwargs):
```
vi4m: hostname_assignment

ar4s:
:ok:

PR #1920, opened 2015-11-13
Added new condition to TransitionForm

ar4s:
Transition must have only one actions with attachment.

### DIFF
--- a/src/ralph/lib/transitions/forms.py
+++ b/src/ralph/lib/transitions/forms.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from django import forms
+from django.utils.translation import ugettext_lazy as _
 
 from ralph.lib.transitions.models import Action, Transition, TransitionModel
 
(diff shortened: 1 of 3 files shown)


src/ralph/lib/transitions/forms.py
```diff
@@ -29,6 +30,19 @@ def __init__(self, *args, **kwargs):
...
+    def clean(self):
+        cleaned_data = super().clean()
+        actions = Action.objects.in_bulk(cleaned_data['actions'])
+        attachment_counter = 0
+        for k, v in actions.items():
+            action = getattr(self.model, v.name)
+            if getattr(action, 'return_attachment', False):
+                attachment_counter += 1
+        if attachment_counter > 1:
+            msg = _('Please check exactly one action wich return attachment')
```
mkurek: Please select at most one action which return attachment.

szok:
:ok:

mkurek:
:ok:

PR #1922, opened 2015-11-13
Add date column to attachments

szok:


src/ralph/attachments/templates/attachments/attachments.html
```diff
@@ -29,6 +30,7 @@
                   {% endif %}
                 </td>
                 <td>{% field form.description is_admin=False show_label=False %}</td>
+                <td>{% if obj.created %}{{ obj.created }}{% else %}&mdash;{% endif %}</td>
```
mkurek: is there any case when `obj.created` will be evaluated as `False`?

szok: yes because its work as inline and obj is empty

mkurek:
:ok:

vi4m:
LGTM

PR #1926, opened 2015-11-13
Add region validation to BaseObject Licence

szok:


mkurek:
what about API? :)

ar4s:
:ok:

mkurek:
:ok:

PR #1931, opened 2015-11-13
Fixed ServiceEnvironment API

mkurek:
- environments are not longer removed when `environments` is missing in update request
- ServiceEnvironment ids are kept after update request
- small fix for error message color (working with foundation now)


szok:
:ok:

ar4s:
:ok:

PR #1933, opened 2015-11-16
Added possibility to filter by imported object id in API

mkurek:
- added possibility to filter by imported object id in API
- added additional filters for ServiceEnvironment and Service


szok:
:ok:

ar4s:
:ok:

PR #1935, opened 2015-11-16
Replaced static templatetag by staticfiles

ar4s:
https://github.com/django/django/blob/master/django/templatetags/static.py#L135 vs https://github.com/django/django/blob/master/django/contrib/staticfiles/templatetags/staticfiles.py#L20


szok:
:ok:

mkurek:
:ok:

PR #1938, opened 2015-11-17
Fixed transitions in reversion history

mkurek:
- removed tabs and transition button from revert view
- removed duplicated office infrastructure field in BO admin form


szok:
:ok:

ar4s:
:ok:

PR #1939, opened 2015-11-17
Improved list filters

mkurek:
- added tooltip about multivalue search
- improved fields and headers margins
- fixed datepicker
- value (whitespace) trimming
- created base filter template

ng_filter


ar4s:
:ok:

szok:
:ok:

PR #1945, opened 2015-11-18
[WIP] Cloud hosts in Ralph & OpenStack sync

bnogas:
- add virtual app
- add openstack_sync management script


src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,295 @@
...
+            )
+
+    def _cleanup(self):
+        """
+        Remove all projects that doesn't exist in openstack from ralph
+        """
+        for project_id in (set(self.ralph_projects.keys()) - set(
+            self.openstack_projects.keys())
+        ):
+            # TODO: Wrzucic do revison
```
ar4s: try use `revision_context_manager.start()` and `revision_context_manager.end()` from django-reversion (ex. https://github.com/etianen/django-reversion/blob/8c0113d1bb18cc9d06d7724df3102914877a3ff9/src/reversion/middleware.py#L22)

src/ralph/virtual/admin.py
```diff
@@ -0,0 +1,67 @@
...
+        return False
+
+
+@register(CloudProvider)
+class CloudProviderAdmin(RalphAdmin):
+    pass
+
+
+@register(CloudProject)
+class CloudProjectAdmin(RalphAdmin):
```
mkurek, line 145: add `list_select_related` here with `cloudprovider`

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,58 @@
...
+
+    class Meta:
+        verbose_name = _('Cloud host')
+        verbose_name_plural = _('Cloud hosts')
+
+    def __str__(self):
+        return 'Cloud Host: {}'.format(self.hostname)
+
+    @property
+    def ipaddress_set(self):
```
mkurek: wouldn't this property be added by django out of the box?

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,72 @@
...
+from ralph.data_center.models.networks import IPAddress
+from ralph.lib.mixins.models import NamedMixin
+
+
+class CloudProvider(NamedMixin):
+    class Meta:
+        verbose_name = _('Cloud provider')
+        verbose_name_plural = _('Cloud providers')
+
+    def __str__(self):
```
mkurek: this is default behaviour in `NamedMixin` - you don't have to duplicate it here

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,72 @@
...
+
+class CloudFlavor(NamedMixin):
+    cloudprovider = models.ForeignKey(CloudProvider)
+    flavor_id = models.CharField(unique=True, max_length=100)
+
+    @property
+    def component_set(self):
+        return Component.objects.filter(asset=self)
+
+    def __str__(self):
```
mkurek: see comment for `__str__` in `CloudProvider`

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def cloudhost_set(self):
+        return CloudHost.objects.filter(parent=self)
+
+
+class CloudHost(BaseObject):
+    host_id = models.CharField(unique=True, max_length=100)
+    hostname = models.CharField(max_length=100)
+    cloudflavor = models.ForeignKey(CloudFlavor, verbose_name="Instance Type")
```
mkurek: instance type should be translated

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.name
+
+
+class CloudFlavor(NamedMixin):
+    cloudprovider = models.ForeignKey(CloudProvider)
+    flavor_id = models.CharField(unique=True, max_length=100)
+
+    @property
+    def component_set(self):
+        return Component.objects.filter(asset=self)
```
mkurek: since you're not inheriting from Asset/BaseObject, this filter will not work

src/ralph/virtual/tests/test_openstack.py
```diff
@@ -0,0 +1,95 @@
+from ralph.assets.models.assets import ServiceEnvironment
```
vi4m: coding missing

vi4m:
I expect new functionality to have more verbose commit log (maybe even paragraph). Short but good documentation is needed too.

src/ralph/virtual/tests/test_openstack.py
bnogas: added

vi4m:
Please make separate branch for it called feature/openstack-sync and make pull request there

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+from ralph.assets.models.components import ComponentModel, GenericComponent
+from ralph.data_center.models.networks import IPAddress
+from ralph.virtual.models import (
+    CloudFlavor,
+    CloudHost,
+    CloudProject,
+    CloudProvider
+)
+
+setup()
```
mkurek: why do you have to call `setup`?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+                'tag':  site['tag'],
+                'servers': {},
+            }
+
+        if len(projects) == 0:
+            raise NotImplementedError('Got an empty list of projects from '
+                                      '{}'.format(site['auth_url']))
+        return projects
+
+    def _run_openstack_query(self):
```
mkurek: this method name could be more descriptive, ex. `_process_openstack_instances`

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+
+        if len(flavors) == 0:
+            raise NotImplementedError('Got an empty list of flavors from '
+                                      '{}'.format(site['auth_url']))
+        return flavors
+
+    @staticmethod
+    def _get_projects_map(site):
+        """
+        Returns a hash map tenant_id->tenant_name
```
mkurek: just map (not hash map)

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+        projects = {}
+        for project in keystone_client.tenants.list():
+            projects[project.id] = {
+                'name': project.name,
+                'tag':  site['tag'],
+                'servers': {},
+            }
+
+        if len(projects) == 0:
+            raise NotImplementedError('Got an empty list of projects from '
```
mkurek: there should be exception here IMO - we could log this, but it's possible (especially for some test instances)

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+                                      '{}'.format(site['auth_url']))
+        return projects
+
+    def _run_openstack_query(self):
+        for site in settings.OPENSTACK_INSTANCES:
+            nt = self._get_novaclient_connection(site)
+            tmp = self.openstack_projects.copy()
```
mkurek: why are you operating here on copy of `openstack_projects`?

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+    cloudprovider = models.ForeignKey(CloudProvider)
+    flavor_id = models.CharField(unique=True, max_length=100)
+
+    def __str__(self):
+        return format(self.name)
+
+    @property
+    def cores(self):
+        """Number of cores"""
+        try:
```
mkurek: there is, for sure, easier way to handle it. First of all, since `GenericComponent` has foreign key to `BaseObject, Django creates reversed reference in`BaseObject`(it should be named like`genericcomponents_set`if not stated otherwise in ForeignKey definition) - this will handle`GenericComponent.objects.filter(asset=self)` for you. Second of all - you don't have to iterate for each component to check if it's type is STH (ex. processor) - just create proper query. My proposal:

```
return self.genericcomponent_set.filter(model__type=ComponentType.processor).values_list('model__cores', flat=True).first()
```

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+class CloudProject(BaseObject):
+    cloudprovider = models.ForeignKey(CloudProvider)
+    project_id = models.CharField(unique=True, max_length=100)
+    name = models.CharField(max_length=100)
+
+    def __str__(self):
+        return 'Cloud Project: {}'.format(self.name)
+
+    @property
+    def cloudhost_set(self):
```
mkurek: doesn't Django create this reversed-foreign-key for you?

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+        verbose_name = _('Cloud host')
+        verbose_name_plural = _('Cloud hosts')
+
+    def __str__(self):
+        return 'Cloud Host: {}'.format(self.hostname)
+
+    @property
+    def ip_addresses(self):
+        ips = []
+        for ip in IPAddress.objects.filter(base_object=self):
```
mkurek: `return ','.join(self.ipaddress_set.values_list('address', flat=True))`

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+            ips.append(ip.address)
+        return ', '.join(ips)
+
+    @property
+    def service_env(self):
```
mkurek: i'm not sure if it's good idea to overwrite DB fields like that (`service_env` is `BaseObject`'s field); first of all sql query (or ORM query) will return something different than python property; second of all - it'd be hard to create compound queries with this field (ex. to filter all hosts with particular service). Maybe you could rewrite here service from it's parent? You could use Django signal to handle it (https://docs.djangoproject.com/en/1.8/topics/signals/) - just subsribe to any `CloudProject` updates (post_save) and update there service_env for all of its children.

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+        return self.parent.service_env
+
+
+class VirtualServer(BaseObject):
```
mkurek, line 102: please add some todo here that this model needs some development (ex. hostname field etc)

src/ralph/virtual/tests/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+    CloudFlavor,
+    CloudHost,
+    CloudProject,
+    CloudProvider
+)
+
+
+class OpenstackModelsTestCase(RalphTestCase):
+    def setUp(self):
+        self.envs = EnvironmentFactory.create_batch(1)
```
mkurek: just `self.env = EnvironmentFactory()` - you don't have to call `create_batch` if you're creating only single instance

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+            self._delete_object(GenericComponent.objects.get(
+                asset=flavor_obj,
+                model=model))
+
+        if set(ralph_names) != set(names):
+            return 1
+        else:
+            return 0
+
+    def _add_cpu(self, flavor, flavor_obj):
```
mkurek: `_add_cpu`, `_add_ram` and `_add_disk` methods looks very similar - could you create one base method and call it from `_add_cpu`, `_add_ram` and `_add_disk` only passing particular params values (ex. `_add_component(flavor['cores'], 'cores vCPU', ComponentType.processor)`?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+        except ObjectDoesNotExist:
+            self._save_object(
+                GenericComponent(asset=flavor_obj, model=model),
+                "Add disk component"
+            )
+
+        return name
+
+    def _add_flavor(self, flavor, flavor_id):
+        """Add/modify flavor ro ralph"""
```
mkurek: ro?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+        if flavor_id not in self.ralph_flavors:
```
mkurek, line 408: why do you have to collect all flavors before this method? what if some other process add another flavor in the meantime? I'll suggest sth like this:

```
mod = False
ralph_flavor, created = CloudFlavor.objects.get_or_create(flavor_id=flavor_id, cloudprovider=self.cloud_provider, defaults=dict(name=flavor['name']))
if ralph_flavor.name != flavor['name']:
    ralph_flavor.name = flavor['name']
    self._save_object(ralph_flavor, 'Change name')
    mod = True
mod = self._add_components(flavor, ralph_flavor) or mod
if created:
    self.summary['new_flavors'] += 1
elif mod:
    self.summary['mod_flavors'] += 1
```

(unfortunately this `get_or_create` will probably doesn't work with your reversion mechanism)

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -0,0 +1,532 @@
...
+            project_id = project.project_id
+            self.ralph_projects[project_id] = {
+                'name': project.name,
+                'servers': {},
+                'tags': project.tags.names()
+            }
+            try:
+                self.ralph_projects[project_id][
+                    'cloudprovider'] = project.cloudprovider
+            except ObjectDoesNotExist:
```
mkurek: when could this error occur? Since `cloudprovider` is mandatory field for `CloudProject`, this try-except is useless here

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+    flavor_id = models.CharField(unique=True, max_length=100)
+
+    def __str__(self):
+        return format(self.name)
+
+    @property
+    def cores(self):
+        """Number of cores"""
+        try:
+            for component in GenericComponent.objects.filter(asset=self):
```
ar4s: NOTE: field `asset` will be rename to `base_object` (in [Component](https://github.com/allegro/ralph/blob/ng/src/ralph/assets/models/components.py#L44))

mkurek: thats exactly why using `self.genericcomponent_set` will be better here - you don't have to worry about field name on the other side :)

src/ralph/virtual/models.py
```diff
@@ -0,0 +1,104 @@
...
+                if component.model.type == ComponentType.processor:
+                    return component.model.cores
+            return None
+        except ObjectDoesNotExist:
```
ar4s: why not `GenericComponent.DoesNotExist`?

ar4s: When `.filter` raise `DoesNotExist`?

src/ralph/virtual/tests/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+        CloudHost.objects.create(
+            host_id='host_id1',
+            hostname='host1',
+            parent=project1,
+            cloudflavor=flavor1
+        )
+
+    def test_check_cloudflavor_cores(self):
```
ar4s: use `ddt` for `test_check_cloudflavor_*` ([example of use](https://github.com/allegro/ralph/blob/ng/src/ralph/data_center/tests/test_models.py#L199-L214)) 
NOTE: remember about decorating test case by `ddt`

src/ralph/virtual/tests/test_openstack.py
```diff
@@ -0,0 +1,198 @@
...
+                ),
+                host.created,
+            )
+            self.assertIn(TEST_HOSTS[host_id]['tag'], host.tags.names())
+            for ip in TEST_HOSTS[host_id]['ips']:
+                self.assertIn(ip, list(ips))
+
+    def test_check_add_flavor(self):
+        """Check if flavors are added and modified correctly"""
+        for flavor_id in OPENSTACK_FLAVOR:
```
ar4s: use `.items()` for `OPENSTACK_FLAVOR` then you can iter key and value at the same time
`for flavor_id, flavor in OPENSTACK_FLAVOR.items()`

src/ralph/virtual/tests/test_openstack.py
```diff
@@ -0,0 +1,198 @@
...
+            asset=flavor_obj, model=self.test_model
+        )
+
+    def test_check_complete(self):
+        """Check the whole run of the script"""
+        self.cmd._update_ralph()
+        self.cmd._cleanup()
+
+        # Objects add/modification
+        for flavor_id in OPENSTACK_FLAVOR:
```
ar4s: ditto

src/ralph/virtual/tests/test_openstack.py
```diff
@@ -0,0 +1,198 @@
...
+            flavor = OPENSTACK_FLAVOR[flavor_id]
+            ralph_flavor = CloudFlavor.objects.get(flavor_id=flavor_id)
+            self.assertEqual(ralph_flavor.name, flavor['name'])
+            self.assertEqual(ralph_flavor.cloudprovider, self.cloud_provider)
+
+        for project_id in OPENSTACK_DATA:
```
ar4s: ditto

mkurek:
I'm closing it for now - we'll reopen it when we'll merge it into ng branch; check #1999 now

PR #1948, opened 2015-11-18
Added deleted field to exporter

ar4s:


szok:
:ok:

src/ralph/export_to_ng/resources.py
```diff
@@ -302,6 +303,14 @@ def dehydrate_hostname(self, asset):
             hostname = ''
         return hostname
 
+    def dehydrate_deleted(self, asset):
+        deleted = False
+        try:
+            deleted = asset.device_info.ralph_device.deleted
```
mkurek: since asset is SoftDeletable (https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/models_assets.py#L561) IMO we should have condition if device is deleted or asset is deleted

mkurek:
:ok:

mkurek:
:ok:

PR #1951, opened 2015-11-19
Changed dehydrate_deleted to return 1 or 0

ar4s:
Now `dehydrate_deleted` returns 1 or instead of True/False

### DIFF
--- a/src/ralph/export_to_ng/resources.py
+++ b/src/ralph/export_to_ng/resources.py
@@ -309,7 +309,7 @@ class DataCenterAssetResource(AssetResource):
             deleted = asset.deleted or asset.device_info.ralph_device.deleted
         except AttributeError:
             pass
-        return deleted
+        return '1' if deleted else '0'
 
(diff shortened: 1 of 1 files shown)


szok:
:ok:

src/ralph/export_to_ng/resources.py
```diff
@@ -309,7 +309,7 @@ def dehydrate_deleted(self, asset):
             deleted = asset.deleted or asset.device_info.ralph_device.deleted
         except AttributeError:
             pass
-        return deleted
+        return '1' if deleted else '0'
```
mkurek, line 312: what about deleted for service or licence?

ar4s, line 312: by default `Resource` returns 1/0 for boolean type but if you using dehydrate then you must care of this - BackOfficeAsset, Support and Licence return correct values

mkurek:
:ok:

PR #1954, opened 2015-11-19
There are fileds without flatchoices property

msuszko:
.flatchoices is property of `django.db.models.Field`, but
`django.db.models.fields.related.ForeignObjectRel` doesn't have it.
This check breaks filters on ForeignKey fields.


mkurek:
:ok:

szok:
:ok:

PR #1956, opened 2015-11-19
Clean-up in Support models

mkurek:
- added property_of and asset_type to support
- changed contract_terms to TextField
- added budget_info and property_of to support importer


szok:
:ok:

ar4s:
:ok:

PR #1958, opened 2015-11-20
deleted configuration_path

ar4s:
Now it's not necessary.


mkurek:
:ok: (we'll get back to this field when we'll migrate ralph core)

szok:
:ok:

PR #1964, opened 2015-11-20
Management IP for DataCenterAsset

mkurek:
TEMPORARY SOLUTION! (until core part of Ralph will not be fully migrated to NG)
- added management_ip and management_hostname to DataCenterAsset
- fixed attachment migration


ar4s:
:ok:

szok:
:ok:

PR #1965, opened 2015-11-20
Added management IP/hostname to exporter

mkurek:
- added management IP/hostname to DataCenterAsset resource
- fixed exporting Network when host bits are set
- allow to exclude some models from export


ar4s:
:ok:

szok:
:ok:

PR #1972, opened 2015-11-22
ServiceEnvironment inheriting from BaseObject

mkurek:
ServiceEnvironment now inherits from BaseObject!

Added new migration operation which will help to inherit more models from BaseObject in the future. Basic steps:
- rename id to baseobject_ptr_id
- generate new ids for each instance
- fix foreign keys pointing to the old id

I've added new package 'django_migration_testcase' (MIT licence) to better test migrations in the future.

### DIFF
--- a/src/ralph/assets/migrations/_helpers.py
+++ b/src/ralph/assets/migrations/_helpers.py
@@ -1,0 +1,152 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from collections import defaultdict
+from functools import partial
+
+from django.db import connection, migrations, models
+
+
+def baseobject_migration(
+    apps, schema_editor, app_name, model_name, rewrite_fields=None
+):
+    """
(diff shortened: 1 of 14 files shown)


src/ralph/assets/migrations/_helpers.py
```diff
@@ -0,0 +1,197 @@
...
+):
+    """
+    Handle ids (duplication) for new BaseObject inherited model.
+
+    Create new BaseObject instance for each old-type objects, then update
+    baseobject_ptr_id to point to new BaseObject id. At the end, update all
+    related models to new point to new ids.
+    """
+    model_str = '{}.{}'.format(app_name, model_name)
+    print('Iheriting {} from baseobject'.format(model_str))
```
tootiefrootie: Inheriting

szok:
:ok:

ar4s:
:ok:

PR #1977, opened 2015-11-23
Password bugfix in Service Serializer and add fields to Service Resource

szok:
- Password bugfix in Service api serializer.
- Add technical_owners and business_owners to Service Resource.


ar4s:
:ok:

mkurek:
:ok:

PR #1984, opened 2015-11-24
Filter DataCenterAsset by service-env

mkurek:
- added filtering DC Asset by serviceenv
- fixed my equipment links to load json from env variables


ar4s:
:ok:

szok:
:ok:

PR #1997, opened 2015-11-25
Fixed TimeStampMixin created and modified fields

bnogas:
- change auto_now=True to auto_now_add=True in created field
- change auto_now_add=True to auto_now=True in modified field


mkurek:
:ok:

ar4s:
:ok:

PR #2004, opened 2015-11-26
fixed transitions and history button layout

mkurek:
screen shot 2015-11-26 at 13 39 26

In the future we'll definitely have to change this tabs layout to flex.


szok:
:ok:

ar4s:
:ok:

PR #2005, opened 2015-11-26
Fixed bulk-edit permissions

mkurek:
- fixed permissions on list view
- fixed permissions (readonly) in bulk-edit mode
- fixed form for BO Asset when user does not have permissions to hostname field


szok:
:ok:

ar4s:
:ok:

PR #2007, opened 2015-11-26
Asset fields fixes

mkurek:
- add default value for force_depreciation field
- allow management_ip field to save null values if empty


ar4s:
:ok:

szok:
:ok:

PR #1994, opened 2015-11-25
bulk transition

ar4s:
- [x] working old tests
- [ ] new tests
- [x] refactoring

PR #2012, opened 2015-11-27
Update owner after user select in transition

szok:


mkurek:
:ok: for me but wait for #1994 to be merged

ar4s:
#1994 now contains @szok's changes

PR #2015, opened 2015-11-27
fixed polymorphic select related if not defined

mkurek:


src/ralph/lib/polymorphic/models.py
```diff
@@ -63,11 +63,17 @@ def iterator(self):
                 if select_related:
                     model_query.query.select_related = select_related.copy()
 
-                if model_name in self._polymorphic_select_related:
+                if (
+                    model_name in self._polymorphic_select_related and
+                    self._polymorphic_select_related[model_name]
+                ):
```
szok: it may be better: 

```
if self._polymorphic_select_related.get(model_name, None):

```

ar4s:
:ok:

szok:
:ok:

PR #2019, opened 2015-12-01
Simplify permissions widget v1

mkurek:
- exclude some permissions from widget in user and group form
- fixed permissions ordering on widget


ar4s:
:ok:

szok:
:ok:

PR #2024, opened 2015-12-01
Skip country field in change hostname transition if owner field is used

mkurek:
- transition field condition takes additional param: list of actions
- modified change hostname action: if owner field is used in the same transition, don't show country field - take owner's country instead


szok:
:ok:

ar4s:
:ok:

PR #2027, opened 2015-12-02
Fixed transitions permissions

mkurek:
- fixed transition perm name (added app_label)
- fixed view perm (check both in model and user model)
- show only transitions available for user


szok:
:ok:

ar4s:
:ok:

PR #2028, opened 2015-12-02
Added view permission by default for each model (monkey patch)

ar4s:
Sorry for that but it's only way :crying_cat_face: (excluding PR to Django)


szok:
Description is the best :) 
:ok:

mkurek:
:ok:

PR #2030, opened 2015-12-02
Respect user permission in actions

ar4s:


src/ralph/lib/transitions/models.py
```diff
@@ -346,9 +352,15 @@ def permission_info(self):
         }
 
     @classmethod
-    def transitions_for_model(cls, model):
+    def transitions_for_model(cls, model, user=None):
         content_type = ContentType.objects.get_for_model(model)
-        return cls.objects.filter(model__content_type=content_type)
+        transitions = cls.objects.filter(model__content_type=content_type)
+        if not user:
```
mkurek: no need for this if (`_check_user_perm_for_transition` will handle it for you)

src/ralph/lib/transitions/models.py
```diff
@@ -157,6 +157,15 @@ def _check_action_with_instances(instances, transition):
         validation_func(instances)
 
 
+def _check_user_perm_for_transition(user, transition):
+    if not user:
+        return True
```
szok, line 162: Why true?

ar4s, line 162: it's useful when execute from CLI whithout user

mkurek:
:ok:

szok:
:ok:

PR #2033, opened 2015-12-03
Licences: show used-free

mkurek:
- show used / free licences in list
- show used / free licences in change form
- show free licences in autocomplete (**str**)
- new manager for Licence which select in one query number of used licences by user and baseobjects


szok:
:ok:

ar4s:
:ok:

PR #2042, opened 2015-12-03
changed run_before -> run_after

ar4s:
https://github.com/allegro/ralph/blob/ng/src/ralph/lib/transitions/models.py#L177 run_before is not implemented


mkurek:
:ok:

szok:
:ok:

PR #2047, opened 2015-12-04
Add split value to autocomplete widget

szok:
Add split value to autocomplete widget

### DIFF
--- a/src/ralph/admin/autocomplete.py
+++ b/src/ralph/admin/autocomplete.py
@@ -1,4 +1,5 @@
 import operator
+import re
 from functools import reduce
 
 from django.conf.urls import url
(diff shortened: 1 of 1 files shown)


src/ralph/admin/autocomplete.py
```diff
@@ -114,6 +114,27 @@ def dispatch(self, request, *args, **kwargs):
...
+
+        Args:
+            queryset: Django queryset
+            query: Query string
+            search_fields: List of fields
+
+        Returns:
+            Django queryset
+        """
+        for value in query.split():
```
mkurek: use `re` to split by space and dot

ar4s:
:ok:

mkurek:
:ok:

PR #2051, opened 2015-12-07
Add user and owner initial value to transition

szok:


src/ralph/back_office/models.py
```diff
@@ -84,6 +84,28 @@ def _check_user_assigned(instances):
...
+    return None
+
+
+def autocomplete_owner_if_release_report(actions, objects):
+    try:
+        obj = objects[0]
+    except IndexError:
+        return None
+
+    if [True for i in actions if i.__name__ == 'release_report'] and obj.owner:
```
mkurek: `if obj.owner and any([a.__name__ == 'release_report' for a in actions]):`

src/ralph/back_office/models.py
```diff
@@ -84,6 +84,32 @@ def _check_user_assigned(instances):
     return errors
 
 
+def autocomplete_user_if_release_report(actions, objects):
```
mkurek: extract this logic to single function where you'll pass attribute name (owner, user)

ar4s:
:ok:

mkurek:
:ok:

PR #2055, opened 2015-12-07
WIP Add focus to autocomplete filter, bugfix manufacturer filter

szok:


mkurek:
This will be done as a part of m2m in autocomplete.

PR #2057, opened 2015-12-07
Autocomplete in transition duplicate user bugfix

szok:


mkurek:
:ok:

src/ralph/lib/transitions/templates/transitions/run_transition.html
```diff
@@ -47,7 +47,6 @@
 {% block extra_scripts %}
   {{ block.super }}
   {{ form.media.js }}
-  <script type="text/javascript" src="{% static 'src/js/ralph-autocomplete.js' %}"></script>
```
ar4s, line 50:
```
$.fn.autocomplete = function(options) {
        var instance = $.data(this, 'autocomplete');
        if (instance === undefined) {
            instance = this.each(function() {
                $.data(this, 'autocomplete', new AutocompleteWidget(this, options));
            });
        }
        return instance
    };
```

ar4s:
:ok:

PR #2064, opened 2015-12-07
Added tooltip when transition is without form

ar4s:


szok:
:ok:

src/ralph/lib/transitions/templatetags/transitions_tags.py
```diff
@@ -12,7 +12,11 @@ def available_transitions(context, obj, field):
         obj, 'get_available_transitions_for_{}'.format(field), None
     )
     if get_available_transitions:
+        transitions = []
+        for transition in get_available_transitions(user=context.request.user):
+            transition.has_form = transition.has_form(obj)
```
mkurek: it's rather not good idea to overwrite function with attribute - could you rename it somehow?

mkurek:
:ok:

PR #2067, opened 2015-12-08
temporary fix for used/free licences calculation

mkurek:
during `annotate` in `objects_used_free`, cartesian product was made from licences assigned to users and base objects. This PR fix it (but with additional performance cost necessary to calculate free assignments for each licence separately).


ar4s:
:ok:

szok:
:ok:

PR #2069, opened 2015-12-08
Autocomplete in inlines bugfix

szok:


ar4s:
:ok:

src/ralph/admin/templates/admin/base.html
```diff
@@ -68,6 +68,25 @@
       {{ media.js }}
       <!-- extra media -->
       {% block extra_scripts %}{% endblock %}
+      <script type="text/javascript">
+        (function($) {
+            $(document).ready(function() {
+              $('.add-row a').bind('click', function() {
+                var $inline = $(this).closest('table').find(
+                  '.form-row:not(#baseobjectlicence_set-empty) .autocomplete-widget'
```
mkurek: this should be generic

src/ralph/admin/templates/admin/base.html
```diff
@@ -68,6 +68,25 @@
       {{ media.js }}
       <!-- extra media -->
       {% block extra_scripts %}{% endblock %}
+      <script type="text/javascript">
+        (function($) {
+            $(document).ready(function() {
```
mkurek, line 73: please add some comment what does this piece of code do

mkurek:
:ok:

PR #2074, opened 2015-12-08
fix licences used/free query

mkurek:
Unfortunately, raw sql to be used to get users used quantity and base objects used quantity for licence. Previous solution, with annotate, makes a cartesian product of licence, licenceuser and baseobjectlicence tables and free/used numbers were not valid.


szok:
:ok:

ar4s:
:ok:

PR #2083, opened 2015-12-10
Change hostname using region country instead of owner country

mkurek:
- added country field to region
- BackOfficeAsset country code is based on region instead of owner
- hostname country code is assigned based on region country code
- removed assigning hostname from assign_owner transition (signal with in_progress status chnge handling is enough)


szok:
:ok:

src/ralph/assets/models/assets.py
```diff
@@ -363,6 +366,10 @@ def render_template(template):
             template = Template(template)
             context = Context(template_vars or {})
             return template.render(context)
+
+        logger.warning('Generating new hostname for {} using {}'.format(
```
ar4s, line 370: please add old hostname to the message

ar4s:
:ok:

PR #2084, opened 2015-12-10
`make install` now fetches static files as well

vi4m:
`make install` should run `make install-js` as dependency. 
`make install-js` now fetches all static dependencies.


szok:
:ok:

ar4s:
:ok:

PR #2091, opened 2015-12-14
added support for Raven JS

ar4s:


szok:
:ok:

src/ralph/admin/templatetags/ralph_tags.py
```diff
@@ -14,6 +14,10 @@
 
 register = Library()
 
+SENTRY_ENABLED = getattr(settings, 'SENTRY_ENABLED', False)
```
mkurek: please add default values to settings for this variables and use env. variables to set value in settings

src/ralph/settings/base.py
```diff
@@ -240,6 +244,11 @@
     'PASSWORD': os.environ.get('REDIS_PASSWORD', ''),
 }
 
+
+SENTRY_ENABLED = os_env_true('SENTRY_ENABLED'),
```
mkurek: this will be tuple you know? :)

src/ralph/settings/base.py
```diff
@@ -240,6 +244,11 @@
...
+SENTRY_JS_DSN = os.environ.get('SENTRY_JS_DSN', None),
+SENTRY_JS_CONFIG = os.environ.get('SENTRY_JS_CONFIG', '{}'),
```
mkurek: shouldn't this be loaded to json?

mkurek:
:ok:

PR #2092, opened 2015-12-14
Base object licence clean metgod bugfix

szok:
Bugfix in BaseObjectLicence clean method.
If there was no license raised an exception.


ar4s:
:ok:

mkurek:
:ok:

PR #2093, opened 2015-12-15
LDAP sync fix & accounts views improvements

mkurek:
- parse ldap groups mapping in body of mirror_group function (this module is loaded in settings so settings cannot be readed at module level)
- fix my equipment view permissions
- show ldap group mapping in group view
- show users list in group view


szok:
:ok:

ar4s:
:ok:

PR #2095, opened 2015-12-16
fixed unique validation for NullableCharField in bulk edit

mkurek:
refer to https://github.com/django/django/blob/1.8.7/django/forms/models.py#L666

Django doesn't validate uniqueness when there is `None` in field value(s)


ar4s:
:ok:

szok:
:ok:

PR #2104, opened 2015-12-17
Improved app loading + other minor improvements

mkurek:
- views and urls module from app will be automatically loaded when subclassing `RalphAppConfig`
- replaced `sys.stdout.write` by logger
- added additional logging to permissions check and loading
- console handler used by default with dev settings


szok:
:ok:

ar4s:
:ok:

PR #2109, opened 2015-12-17
Hiding liquidated status in BO and DC Asset

szok:
screen shot 2015-12-18 at 13 04 41

### DIFF
--- a/src/ralph/admin/autocomplete.py
+++ b/src/ralph/admin/autocomplete.py
@@ -146,12 +146,15 @@ class AutocompleteList(SuggestView):
         if not search_fields:
             return []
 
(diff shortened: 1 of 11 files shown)


mkurek:
Couldn't you extend filter for status field?

szok:
But then by default we will not display objects with this status.

mkurek:
And that's how your code is working too. IMO this should work as follows:
- by default all assets except liquidated are displayed (when `All` status filter is chosen or without status filter)
- liquidated assets are shown only when `liquidated` filter value is chosen

mkurek:
:ok:

ar4s:
:ok:

PR #2119, opened 2015-12-22
Add tooltip to support autocomplete

szok:


ar4s:
:ok:

src/ralph/admin/widgets.py
```diff
@@ -156,6 +156,7 @@ def get_related_url(self, info, action, *args):
         )
 
     def render(self, name, value, attrs=None):
+        polymorphic_models = []
```
mkurek: is this used anywhere?

mkurek:
:ok:

PR #2120, opened 2015-12-22
Add is_valid to transition POST

szok:


ar4s:
:ok:

src/ralph/lib/transitions/views.py
```diff
@@ -164,6 +164,10 @@ def get(self, request, *args, **kwargs):
         return self._is_valid() or super().get(request, *args, **kwargs)
 
     def post(self, request, *args, **kwargs):
+        is_valid = self._is_valid()
```
mkurek: rename it to `not_valid`

mkurek:
:ok:

PR #2121, opened 2015-12-22
Fixed polymorphic manager ordering

mkurek:


src/ralph/lib/polymorphic/models.py
```diff
@@ -29,12 +29,19 @@ def iterator(self):
             - For each ContentType generates additional queryset
             - Returns iterator with different models
         """
+        # if this is final-level model, don't check for descendants - just
+        # return original queryset result
+        if not getattr(self.model, '_polymorphic_descendants', []):
+            yield from super().iterator()
```
szok, line 35: :+1:

szok:
:ok:

ar4s:
:ok:

PR #2127, opened 2015-12-23
changed SaveSecurityScanSerializer inheritance to RalphAPISaveSerializer

mkurek:
change inherited serializer to SaveSecurityScanSerializer to properly handle related field in DRF default form (raw input instead of list of choices)


ar4s:
:ok:

szok:
:ok:

PR #2128, opened 2015-12-23
Django 1.9 in travis

szok:


.travis.yml
```diff
@@ -35,6 +40,7 @@ install:
     - gulp
     - make install-test
     - pip install coveralls
+    - if [[ $DJANGO_VERSION == 1.9 ]]; then pip install django==1.9; fi
```
mkurek: just install django in proper version (we could test django dev as well and don't want to if every case)

mkurek:
:ok:

ar4s:
:ok:

PR #2129, opened 2015-12-23
Improved get_free_u method 

ar4s:
Before `get_free_u` might returned negative value (in real life it's not impossible) after this change `get_free_u` work properly. The method respect two side (front, back) and result is a sum of both sides.

### DIFF
--- a/src/ralph/data_center/models/physical.py
+++ b/src/ralph/data_center/models/physical.py
@@ -7,6 +7,7 @@ from django import forms
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.db import models
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.admin.sites import ralph_site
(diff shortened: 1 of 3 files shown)


src/ralph/data_center/models/physical.py
```diff
@@ -190,22 +191,36 @@ def get_orientation_desc(self):
...
+
+        def fill_u_list(objects, height_of_device=lambda obj: 1):
+            for obj in objects:
+                start = obj[0]
+                end = obj[0] + int(height_of_device(obj))
+                for position in range(start, end):
+                    u_list[position - 1] = False
+        fill_u_list(accessories)
+        fill_u_list(dc_assets, lambda obj: obj[1])
+        return Counter(u_list)[True]
```
mkurek: return `sum(u_list)` will be enough :)

requirements/test.txt
```diff
@@ -7,5 +7,6 @@ ddt
 django-plainpasswordhasher
 django_migration_testcase
 factory-boy
+fake-factory
```
mkurek: where are you using it?

src/ralph/data_center/models/physical.py
```diff
@@ -227,6 +242,7 @@ class DataCenterAsset(Asset):
         choices=Orientation(),
         default=Orientation.front.id,
     )
+
```
mkurek: please remove this line

src/ralph/data_center/models/physical.py
```diff
@@ -190,22 +191,36 @@ def get_orientation_desc(self):
...
-        return self.max_u_height - dc_assets_height - len(set(accessories))
+            'position').filter(rack=self)
+        dc_assets = self.get_root_assets().values_list(
+            'position', 'model__height_of_device'
+        )
+
+        def fill_u_list(objects, height_of_device=lambda obj: 1):
+            for obj in objects:
+                start = obj[0]
+                end = obj[0] + int(height_of_device(obj))
```
mkurek: what if someone create device (model) with height of device greater than height of rack (or at least end of device will be greater than end of rack)? IMO this should be `max(self.max_u_height, obj[0] + int(height_of_device(obj)))`

vi4m:
http://chris.beams.io/posts/git-commit/

mkurek:
:ok:

szok:
:ok:

src/ralph/data_center/tests/test_models.py
```diff
@@ -272,3 +272,47 @@ def test_change_rack_in_descendants(self):
...
+        rack = RackFactory()
+        self.assertEqual(rack.get_free_u(), rack.max_u_height)
+
+    @unpack
+    @data(
+        (1, 47, 47, 0),
+        (1, 46, 47, 1),
+        (2, 45, 47, 2),
+        (47, 1, 47, 46),
+        (47, 5, 47, 46),
```
ghost, line 288: Is this valid?

ar4s, line 288: nope :) but is ok - it testing corner case

ghost, line 288: would you add comment for weirdos like me ;P
it could be
`# works ok even when position greater than max_u`

ghost:
:ok: (my comment is minor, I'm not blocking this change)

PR #2130, opened 2015-12-23
fixed host_ip parsing in security scan serializer

mkurek:
try to find first by IPAddress, then by management_ip if not found


szok:
:ok:

ar4s:
:ok:

PR #2133, opened 2015-12-29
Base object support bugfix

szok:
Bugfix for Support BaseObjectSupport through model.
It was not possible to add new elements from gui


mkurek:
:ok:

vi4m:
Please fix commit message.

ar4s:
:ok:

PR #2134, opened 2015-12-29
OpenStack Sync refactoring

mkurek:
- added image_name to CloudHost model
- fixed openstack projects updating
- openstack sync refactorization, added logging, performance improvements etc


szok:
:ok:

PR #2135, opened 2015-12-29
Cloud admin improvements

mkurek:
- added filter for tags and cloud provider
- reduced SQL queries count
- show service env for host on list


szok:
:ok:

PR #2138, opened 2015-12-30
Cloud Flavor view bugfix

mkurek:
Add instances_count to the read only fields list


vi4m:
Commit message should reflect the change, something like: `Add instances_count to the read only fields list`

szok:
:ok:

ar4s:
:ok:

PR #2140, opened 2016-01-04
encoding " using unicode \u2033 character in report failure link

mkurek:
replace quotation by unicode inches sign in report failure url


szok:
:ok:

ar4s:
:ok:

PR #2141, opened 2016-01-05
Add sort to datacenter and backoffice supports.

szok:
Add sort by support__date_to to DataCenterAsset and BackOfficeAsset
supports.


mkurek:
:ok:

ar4s:
:ok:

PR #2143, opened 2016-01-07
Fixed permissions to additional list columns (non-models)

mkurek:
- before this commit, additional columns (created from property on model or function in admin) were visable only to superuser
- now such columns are by default visible to everyone (who has permission to model)
- there is possibility to specify model-field which will be used as permissions to such additional column


szok:
:ok:

ghost:
:ok:

PR #2145, opened 2016-01-07
Fixed transtion assign loan end date field

mkurek:
Previous field (TextField) assigned date as string to BO Asset object, which causes reversion to crash when serializing object (`datetime.date` was expected to call isoformat on it). Changed field to `DateField`


ghost:
:ok:

szok:
:ok:

PR #2147, opened 2016-01-08
None position bugfix

szok:


mkurek:
imo we should provide default for position field (0?) and make it PositiveIntegerField - @ar4s what do you think?

src/ralph/data_center/models/physical.py
```diff
@@ -214,9 +214,11 @@ def get_free_u(self):
 
         def fill_u_list(objects, height_of_device=lambda obj: 1):
             for obj in objects:
-                start = obj[0] - 1
+                position = obj[0] or 0
```
ghost: For me fix should be simply:

```
for obj in objects:
    if obj[0] == None:
        continue  # no position means do not take place in rack
```

ar4s: :+1:, but I think it should be `if obj[0]` for `0` and `None`

ar4s:
@mkurek I agree with you it should be `PositiveIntegerField` and `0` as a default value.

ghost:
Is 0 valid value, like 1, 2 or 3?

src/ralph/data_center/models/physical.py
```diff
@@ -214,6 +214,8 @@ def get_free_u(self):
 
         def fill_u_list(objects, height_of_device=lambda obj: 1):
             for obj in objects:
+                if not obj[0]:
```
mkurek: the question here is if object with position 0 should (or could) have non-zero height: https://github.com/allegro/ralph/blob/ng/src/ralph/data_center/models/physical.py#L333 ? Anyone know something about it?

mkurek: after discussion with @xliiv:
- make this condition more explicit: `if obj[0] == 0 or obj[0] is None`
- add some comment that position is None when objects simply does not have (assigned) position and position 0 is for some accessories (pdu?) with left-right orientation and should not be included in free/filled space.

ghost: +1

mkurek:
:ok:

ghost:
:ok:

PR #2154, opened 2016-01-13
Import kwargs to Transitions History

szok:
- Bugfix import transitions history
- Bugfix display user transitions history
- Added attachment_url template tag


src/ralph/admin/templatetags/ralph_tags.py
```diff
@@ -96,6 +97,28 @@ def get_verbose_name(obj, name):
     return get_field_by_relation_path(obj, name).verbose_name
 
 
+@register.simple_tag
+def attachment_url(history):
```
mkurek: rename it to user_history_attachment_url

ghost:
:ok:

mkurek:
:ok:

PR #2156, opened 2016-01-15
Dc move to bo transition

szok:
- [x] Move DC to BO
- [x] Move BO to DC
- [x] Unit tests

### DIFF
--- a/docs/development/transitions.md
+++ b/docs/development/transitions.md
@@ -55,6 +55,8 @@ Allowed params for field::
     ``field`` - standard form field, e.g. from ``django.forms``,
     ``condition`` - function wich accept one parameter and return boolean, when condition have be met the field will be shown.
 
+Set ``only_one_action`` to ``True`` if the transition is to be only one action requirements.
+
 Set ``return_attachment`` to ``True`` if action return attachment (e.g.: PDF document).
 
 If you want to add additional information to the transition history, you need to add to the dictionary history_kwargs in your action:
--- a/src/ralph/admin/tests/tests_helpers.py
+++ b/src/ralph/admin/tests/tests_helpers.py
@@ -1,11 +1,19 @@
 # -*- coding: utf-8 -*-
+from ddt import data, ddt, unpack
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldDoesNotExist
 from django.test import TestCase
 
-from ralph.admin.helpers import get_field_by_relation_path, getattr_dunder
+from ralph.admin.helpers import (
+    get_content_type_for_model,
+    get_field_by_relation_path,
+    getattr_dunder
+)
 from ralph.assets.models.assets import Asset, Manufacturer
+from ralph.assets.models.base import BaseObject
 
 
+@ddt
 class ModelFieldsTestCase(TestCase):
     def test_return_ok_when_simply_field(self):
         field_name = 'barcode'
@@ -31,3 +39,14 @@ class ModelFieldsTestCase(TestCase):
         a.b = A()
         a.b.name = 'spam'
         self.assertEqual(getattr_dunder(a, 'b__name'), 'spam')
+
+    @unpack
+    @data(
+        (BaseObject, Asset),
+        (Manufacturer, Manufacturer)
+    )
+    def test_get_content_type_for_model(self, expected_model, model):
+        self.assertEqual(
+            ContentType.objects.get_for_model(expected_model),
+            get_content_type_for_model(model)
+        )
--- a/src/ralph/assets/utils.py
+++ b/src/ralph/assets/utils.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from django.db import transaction
+
+
+@transaction.atomic
+def move_parents_models(from_obj, to_obj):
+    """
+    Move parents Django model to diffrent model inheriting from
+    the same parent.
+
+    By default Django does not allow the remove of the model without parents.
+    Clean _meta.parents, we remove object and add back default parents to class
+
+    Args:
+        from_obj: From Django model object
+        to_obj: To Django modelo object
+
+    Example:
+        >> instance = DataCenterAsset.objects.get(pk=1)
+        >> back_office_asset = BackOfficeAsset()
+        >> move_parents_models(instance, back_office_asset)
+    """
+    to_obj.__dict__.update(from_obj.__dict__)
+    to_obj.content_type = None
+    to_obj.save()
+    parents = from_obj._meta.parents
+    from_obj._meta.parents = {}
+    try:
+        from_obj.delete()
+    finally:
+        from_obj._meta.parents = parents
--- a/src/ralph/attachments/migrations/0003_auto_20160121_1346.py
+++ b/src/ralph/attachments/migrations/0003_auto_20160121_1346.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from ralph.admin.helpers import get_content_type_for_model
+
+
+def migrate_attachment_item(apps, schema_editor):
+    AttachmentItem = apps.get_model('attachments', 'AttachmentItem')
+    for item in AttachmentItem.objects.all():
+        model = apps.get_model(
+            item.content_type.app_label, item.content_type.model
+        )
+        item.content_type_id = get_content_type_for_model(model).id
+        item.save()
+
+
+def unload_migrate_attachment_item(apps, schema_editor):
+    AttachmentItem = apps.get_model('attachments', 'AttachmentItem')
+    BaseObject = apps.get_model('assets', 'BaseObject')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    bo_content_type = ContentType.objects.get_for_model(BaseObject)
+    for item in AttachmentItem.objects.filter(content_type=bo_content_type):
+        content_type = BaseObject.objects.get(
+            pk=item.object_id
+        ).content_type
+        item.content_type = content_type
+        item.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('attachments', '0002_auto_20151125_1354'),
+        ('back_office', '0003_auto_20151204_0758'),
+        ('data_center', '0004_auto_20151204_0758'),
+        ('domains', '0002_auto_20151125_1354'),
+        ('licences', '0002_auto_20151204_1325'),
+        ('supports', '0005_auto_20160105_1222'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            migrate_attachment_item,
+            reverse_code=unload_migrate_attachment_item
+        ),
+    ]
--- a/src/ralph/attachments/views.py
+++ b/src/ralph/attachments/views.py
@@ -1,10 +1,15 @@
+import time
+
 from django.forms.models import modelformset_factory
-from django.http import HttpResponseRedirect
+from django.http import FileResponse, HttpResponseRedirect
+from django.shortcuts import get_object_or_404
+from django.utils.http import http_date
+from django.views.generic.base import View
 
 from ralph.admin.views.extra import RalphDetailView
 from ralph.attachments.forms import AttachmentForm
 from ralph.attachments.models import Attachment, AttachmentItem
-from ralph.helpers import add_request_to_form, get_model_view_url_name
+from ralph.helpers import add_request_to_form
 
 
 class AttachmentsView(RalphDetailView):
@@ -38,20 +43,6 @@ class AttachmentsView(RalphDetailView):
         formset = self.get_formset(request)
         return self.render_to_response(self.get_context_data(formset=formset))
 
-    def get_context_data(self, *args, **kwargs):
-        """
-        Extends context by:
-        * url name for attachments for current model
-        """
-        context_data = super().get_context_data(*args, **kwargs)
-        context_data.update({
-            'attachment_url_name': get_model_view_url_name(
-                self.object._meta.model,
-                'attachment',
-            ),
-        })
-        return context_data
-
     def post(self, request, *args, **kwargs):
         """
         Valid and return response from appropriate method depending on
@@ -84,3 +75,24 @@ class AttachmentsView(RalphDetailView):
         return self.render_to_response(
             self.get_context_data(formset=formset)
         )
+
+
+class ServeAttachment(View):
+
+    def get(self, request, id, filename, *args, **kwargs):
+        """
+        All attachments are serving by this view because we need full
+        control (e.g., permissions, rename).
+        """
+        # TODO: respect permissions
+        obj = get_object_or_404(
+            Attachment,
+            id=id,
+            original_filename=filename
+        )
+        fd = open(obj.file.path, 'rb')
+        response = FileResponse(fd, content_type=obj.mime_type)
+        response['Content-Disposition'] = 'attachment; filename="{}"'.format(obj.original_filename)  # noqa
+        http_modified = http_date(time.mktime(obj.modified.timetuple()))
+        response['Last-Modified'] = http_modified
+        return response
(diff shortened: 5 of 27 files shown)


mkurek:
remember about unit tests :)

src/ralph/data_center/models/physical.py
```diff
@@ -435,6 +437,31 @@ def change_rack(cls, instances, request, **kwargs):
...
+    def move_to_bo(cls, instances, request, **kwargs):
+        for instance in instances:
+            back_office_asset = BackOfficeAsset()
+            back_office_asset.__dict__.update(instance.__dict__)
+            back_office_asset.region = Region.objects.get(name='pl')
+            back_office_asset.warehouse = Warehouse.objects.first()
+            back_office_asset.save()
+
+            parents = instance._meta.parents
+            instance._meta.parents = {}
```
mkurek: write some comment what is going on here and why this workaround is necessary.

src/ralph/data_center/models/physical.py
```diff
@@ -435,6 +437,31 @@ def change_rack(cls, instances, request, **kwargs):
...
+    @transition_action(
+        verbose_name=_('Set as BackOffice Asset'),
+        disable_history=True,
+        disable_save_object=True
+    )
+    def move_to_bo(cls, instances, request, **kwargs):
+        for instance in instances:
+            back_office_asset = BackOfficeAsset()
+            back_office_asset.__dict__.update(instance.__dict__)
+            back_office_asset.region = Region.objects.get(name='pl')
```
mkurek: region and warehouse should be transition field

src/ralph/assets/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    Example:
+        >> instance = DataCenterAsset.objects.get(pk=1)
+        >> back_office_asset = BackOfficeAsset()
+        >> move_parents_models(instance, back_office_asset)
+    """
+    to_obj.__dict__.update(from_obj.__dict__)
+    to_obj.save()
+    parents = from_obj._meta.parents
+    from_obj._meta.parents = {}
+    from_obj.delete()
```
mkurek: what if there will be some exception during deletion? parent will be empty for all further objects. this should be wrapped in try-finally imo

src/ralph/back_office/models.py
```diff
@@ -506,6 +507,39 @@ def loan_report(cls, instances, request, **kwargs):
             language=kwargs['report_language']
         )
 
+    @classmethod
+    @transition_action(
+        verbose_name=_('Set as DataCenter Asset'),
+        disable_history=True,
+        disable_save_object=True,
+        form_fields={
+            'rack': {
```
mkurek, line 524: what about position?

src/ralph/back_office/models.py
```diff
@@ -506,6 +507,39 @@ def loan_report(cls, instances, request, **kwargs):
             language=kwargs['report_language']
         )
 
+    @classmethod
+    @transition_action(
+        verbose_name=_('Set as DataCenter Asset'),
+        disable_history=True,
+        disable_save_object=True,
+        form_fields={
```
mkurek, line 523: we should think if model should be changed here as well

src/ralph/assets/utils.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+
+def move_parents_models(from_obj, to_obj):
```
mkurek, line 6: this should be executed in one transaction

ghost:
There should be information about moving in history or something.

src/ralph/assets/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    Args:
+        from_obj: From Django model object
+        to_obj: To Django modelo object
+
+    Example:
+        >> instance = DataCenterAsset.objects.get(pk=1)
+        >> back_office_asset = BackOfficeAsset()
+        >> move_parents_models(instance, back_office_asset)
+    """
+    with transaction.atomic():
```
mkurek: decorator for this function will be enough

src/ralph/assets/utils.py
```diff
@@ -0,0 +1,30 @@
...
+        to_obj.__dict__.update(from_obj.__dict__)
```
mkurek: what about content type of "final" object - is it properly handled here? https://github.com/allegro/ralph/blob/ng/src/ralph/lib/polymorphic/models.py#L176

src/ralph/admin/tests/tests_helpers.py
```diff
@@ -31,3 +37,14 @@ class A():
         a.b = A()
         a.b.name = 'spam'
         self.assertEqual(getattr_dunder(a, 'b__name'), 'spam')
+
```
ar4s: use ``@ddt``

src/ralph/back_office/tests/test_models.py
```diff
@@ -271,6 +275,26 @@ def setUp(self):
         messages = FallbackStorage(self.request)
         setattr(self.request, '_messages', messages)
 
+    def test_move_as_data_center_asset(self):
+        bo_asset = BackOfficeAssetFactory()
```
mkurek, line 280: please add unit test where there is some transition history and attachments attached to asset (we've to make sure that this will be visible from new object too).

src/ralph/attachments/views.py
```diff
@@ -84,3 +75,24 @@ def formset_invalid(self, formset):
...
+
+    def get(self, *args, **kwargs):
+        """
+        All attachments are serving by this view because we need full
+        control (e.g., permissions, rename).
+        """
+        # TODO: respect permissions
+        obj = get_object_or_404(
+            Attachment,
+            id=self.kwargs['id'],
```
mkurek: why not kwargs['id']? you could use another definition too imo: `def get(self, request, id, filename)`

src/ralph/attachments/views.py
```diff
@@ -84,3 +75,24 @@ def formset_invalid(self, formset):
...
+            original_filename=self.kwargs['filename']
+        )
+        fd = open(obj.file.path, 'rb')
```
mkurek: you're not closing file descriptor. use context manager here and wrap it in try-except for open-related exceptions (raise 4XX)

src/ralph/back_office/models.py
```diff
@@ -508,6 +510,53 @@ def loan_report(cls, instances, request, **kwargs):
             language=kwargs['report_language']
         )
 
+    @classmethod
+    @transition_action(
+        verbose_name=_('Set as DataCenter Asset'),
```
mkurek: 'Convert to DataCenter Asset' will fit here better imo

src/ralph/back_office/models.py
```diff
@@ -508,6 +510,53 @@ def loan_report(cls, instances, request, **kwargs):
...
+        }
+    )
+    def move_as_data_center_asset(cls, instances, request, **kwargs):
+        from ralph.data_center.models.physical import DataCenterAsset, Rack  # noqa
+        with transaction.atomic():
+            for i, instance in enumerate(instances):
+                data_center_asset = DataCenterAsset()
+                try:
+                    move_parents_models(instance, data_center_asset)
+                except Exception as e:
```
mkurek: do we know what exceptions expect here?

src/ralph/back_office/models.py
```diff
@@ -517,7 +566,10 @@ def hostname_assigning(sender, instance, raw, using, **kwargs):
     if getattr(settings, 'BACK_OFFICE_ASSET_AUTO_ASSIGN_HOSTNAME', None):
         if instance.status == BackOfficeAssetStatus.in_progress:
             if instance.pk:
-                bo_asset = BackOfficeAsset.objects.get(pk=instance.pk)
+                try:
+                    bo_asset = BackOfficeAsset.objects.get(pk=instance.pk)
+                except BackOfficeAsset.DoesNotExist:
```
mkurek, line 574: add some comment when this exception could occur + add logging here

src/ralph/back_office/models.py
```diff
@@ -517,7 +566,10 @@ def hostname_assigning(sender, instance, raw, using, **kwargs):
...
+                    return
                 if bo_asset.status == BackOfficeAssetStatus.in_progress:
                     return
             logging.info((
```
mkurek: here should be logger used, not logging

src/ralph/lib/transitions/models.py
```diff
@@ -258,6 +261,8 @@ def run_field_transition(
             action, transition
         ))
         func = getattr(first_instance, action.name)
+        if func.disable_save_object:
```
ar4s: ??

src/ralph/data_center/models/physical.py
```diff
@@ -439,6 +443,42 @@ def change_rack(cls, instances, request, **kwargs):
         for instance in instances:
             instance.rack = rack
 
+    @classmethod
+    @transition_action(
```
mkurek, line 480: this action should be the only action in single transition - add additional param to check for that.

src/ralph/lib/transitions/models.py
```diff
@@ -238,10 +241,9 @@ def run_field_transition(
     """
     Execute all actions assigned to the selected transition.
     """
-    if not isinstance(instances, Iterable):
+    if not isinstance(instances, list):
```
mkurek: why did you change that?

mkurek:
update functions (form commit comment) should be moved to migration

src/ralph/attachments/migrations/0003_auto_20160121_1346.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+        item.save()
+
+
+def unload_migrate_attachment_item(apps, schema_editor):
+    AttachmentItem = apps.get_model('attachments', 'AttachmentItem')
+    for item in AttachmentItem.objects.all():
+        content_type = item.content_type.model_class().objects.get(
+            pk=item.object_id
+        ).content_type
```
mkurek, line 26: this will not work for not-polymorphic objects

src/ralph/back_office/models.py
szok, line 574: Done

src/ralph/data_center/models/physical.py
szok, line 480: Done

src/ralph/back_office/models.py
szok: We don't know, unfortunately.

src/ralph/lib/transitions/models.py
szok: Because list it is later modified.

ghost: This is unclear to me.

szok: We expect that here is a list of instances not queryset.
Because that list is modified for bulk transitions.
If this is not a list that is that we operate on one object and we can create a new one.

ghost: Then how it works when `instances` is e.g.: tuple (or any other iterable - is it possible that instances are something else then `list`?)

szok: Tuple can not be modified.
As I wrote, we need an object that can be modified. The only such a list.

If my solution for the transition is not good, ask for suggestions for a better solution move bo to dc and dc to bo. There must be the possibility of redirects to the new objects from transitions.

mkurek: So maybe sth like this:

```
if not isinstance(instances, Iterable):  # it only happens where single model instance is passed to this function (old-style transitions)
    instances = [instances]
instances = list(instances)
```

src/ralph/lib/transitions/decorators.py
```diff
@@ -14,6 +14,8 @@ def decorator(func):
         func.run_after = kwargs.get('run_after', [])
         func.help_text = kwargs.get('help_text', '')
         func.precondition = kwargs.get('precondition', lambda instances: {})
+        func.disable_save_object = kwargs.get('disable_save_object', False)
+        func.only_one_action = kwargs.get('only_one_action', False)
```
mkurek, line 18: add information about these params to transitions documentation

src/ralph/lib/transitions/models.py
szok: This concept is not good.
Because they function inside run_field_transition we can not generate a new list because we lose a reference to it and redirect is will not work.

 I change it to run_field_transition received only a list, then the problem will disappear.

src/ralph/lib/transitions/decorators.py
szok, line 18: Done

src/ralph/lib/transitions/models.py
```diff
@@ -334,10 +338,6 @@ def __new__(cls, name, bases, attrs):
             _transitions_fields[new_class] = fields
             for field in fields:
                 new_class.add_to_class(
-                    'run_{}_transition'.format(field),
```
szok, line 340: @ar4s What do you think about delete this? and use only from run_field_transition()

ar4s, line 340: :ok:

src/ralph/back_office/models.py
mkurek: so IMO we shouldn't catch exception here - instead of it we should catch every exception in transition action call https://github.com/allegro/ralph/blob/ng/src/ralph/lib/transitions/models.py#L270

src/ralph/data_center/tests/test_models.py
```diff
@@ -159,9 +162,22 @@ def setUp(self):
             parent=self.dc_asset,
         )
 
-    # =========================================================================
-    # slot_no
-    # =========================================================================
```
mkurek, line 164: why do you remove this?

src/ralph/lib/transitions/models.py
```diff
@@ -68,8 +71,8 @@ def _generate_transition_history(
 
     return TransitionsHistory(
         transition_name=transition.name,
-        content_type=ContentType.objects.get_for_model(instance._meta.model),
-        object_id=instance.pk,
+        content_type=get_content_type_for_model(instance._meta.model),
+        object_id=instance.id,
```
mkurek: why id, not pk?

src/ralph/lib/transitions/tests/test_models.py
```diff
@@ -14,6 +14,7 @@
     _sort_graph_topologically,
     Action,
     CycleError,
+    run_field_transition,
```
mkurek, line 17: please test case when exception is raised in some transition action

szok, line 17: Done

mkurek:
:ok:

docs/development/transitions.md
```diff
@@ -55,6 +55,8 @@ Allowed params for field::
     ``field`` - standard form field, e.g. from ``django.forms``,
     ``condition`` - function wich accept one parameter and return boolean, when condition have be met the field will be shown.
 
+Set ``only_one_action`` to ``True`` if the transition is to be only one action requirements.
```
xliiv: What does it mean?

src/ralph/back_office/models.py
```diff
@@ -523,10 +569,20 @@ def hostname_assigning(sender, instance, raw, using, **kwargs):
     if getattr(settings, 'BACK_OFFICE_ASSET_AUTO_ASSIGN_HOSTNAME', None):
         if instance.status == BackOfficeAssetStatus.in_progress:
             if instance.pk:
-                bo_asset = BackOfficeAsset.objects.get(pk=instance.pk)
+                try:
+                    bo_asset = BackOfficeAsset.objects.get(pk=instance.pk)
+                except BackOfficeAsset.DoesNotExist:
+                    # Can not assign a new hostname, because there are
+                    # not yet saved the object.
+                    logger.debug(
```
szok: use info

src/ralph/back_office/models.py
```diff
@@ -514,6 +515,51 @@ def loan_report(cls, instances, request, **kwargs):
             language=kwargs['report_language']
         )
 
+    @classmethod
+    @transition_action(
+        verbose_name=_('Convert to DataCenter Assett'),
```
szok: assett

ghost:
:ok:

PR #2157, opened 2016-01-18
Update gitignore.

fossilet:
1. Add `.idea` for Jetbrain IDE's.
2. Remove exclusion for `src/ralph/lib`. Then `__pycache__` under this can be ignored.


mkurek:
:ok:

szok:
:ok:

PR #2160, opened 2016-01-19
Fixed slot_no validation

mkurek:
When model is not assigned, slot_no validation raised `RelatedObjectDoesNotExist`. Now model_id is checked instead of model.


ghost:
:ok:

szok:
:ok:

PR #2161, opened 2016-01-19
Fixed report failure url formatting on my equipment tab

mkurek:
ed8f859 introduced bug with formatting report failure url on my equipment tab - only first-level fields could be used. This commit fixes this bug - nested fields (using __ convention) could be used in report failure url as well.


szok:
:ok:

ghost:
:ok:

PR #2164, opened 2016-01-21
Show buyout date just as it is checked in model category

szok:
Added show_buyout_date boolean field to Model Category


mkurek:
:ok:

ar4s:
:ok:

PR #2169, opened 2016-01-22
Turn on/off position requirement for rack

mkurek:
- added additional param to rack: require position
- when rack requires position, each asset belonging to this rack has to have non-empty position
- when rack does not require position (ex. warehouse-like rack), position could be empty (it's different than Ralph 2.0 behaviour, which required position for every asset)

### DIFF
--- a/src/ralph/data_center/models/physical.py
+++ b/src/ralph/data_center/models/physical.py
@@ -176,6 +176,13 @@ class Rack(AdminAbsoluteUrlMixin, NamedMixin.NonUnique, models.Model):
         default=0,
     )
     accessories = models.ManyToManyField(Accessory, through='RackAccessory')
+    require_position = models.BooleanField(
+        default=True,
+        help_text=_(
+            'Uncheck if position is optional for this rack (ex. when rack '
+            'has warehouse-kind role'
+        )
+    )
 
     class Meta:
         unique_together = ('name', 'server_room')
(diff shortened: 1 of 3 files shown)


ar4s:
:ok:

src/ralph/data_center/models/physical.py
```diff
@@ -375,12 +394,22 @@ def _validate_position_in_rack(self):
                 self.rack.max_u_height,
             )
             raise ValidationError({'position': [msg]})
+        if self.position is not None and self.position < 0:
+            msg = 'Position is lower than 0'
```
ghost: `Position should be 0 or greater`?

src/ralph/data_center/models/physical.py
```diff
@@ -176,6 +176,13 @@ class Rack(AdminAbsoluteUrlMixin, NamedMixin.NonUnique, models.Model):
         default=0,
     )
     accessories = models.ManyToManyField(Accessory, through='RackAccessory')
+    require_position = models.BooleanField(
+        default=True,
+        help_text=_(
+            'Uncheck if position is not required for this rack (ex. when rack '
```
ghost: 'Uncheck if position is optional for this rack (ex. when rack '?
This is only a suggestion, do what you wish :)

ghost:
:ok:

PR #2170, opened 2016-01-22
Enable to clear some fields during multiadd

mkurek:
Some (unique) fields should be cleared for duplicated objects, ex. management ip and hostname for DataCenterAsset. This PR adds the ability to specify what fields should be cleared (and using what value) during duplication.


ar4s:
:ok:

ghost:
:ok:

PR #2171, opened 2016-01-22
Change AssetModel required fields

mkurek:
power_consumption, cores_count and height_of_device cannot be blank anymore (raised IntegrityError, since they cannot be null).


ar4s:
:ok:

ghost:
:ok:

PR #2183, opened 2016-01-28
Update installation.md

cdb-fr:
Adding tips for installation process

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -2,6 +2,7 @@
 
 For production, we provide both deb package and  docker(compose) image.
 We only support Ubuntu 14.04 Trusty distribution.
+Take care that only 64 bits version is available for deb packages.
 
 On the other hand, if you are developer, we strongly suggest using our `Vagrant` inside the `vagrant` directory
 with many development *bells and whistles* included.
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -15,6 +16,9 @@ First, add our official ralph repository:
     sudo apt-key adv --keyserver  hkp://keyserver.ubuntu.com:80 --recv-keys 379CE192D401AB61
     sudo sh -c "echo 'deb https://dl.bintray.com/vi4m/ralph wheezy main' >  /etc/apt/sources.list.d/vi4m_ralph.list"
 
+After, install the secure transport for accessing repository :
```
mkurek, line 19: I've installed Ralph from deb yesterday on clean debian and first of all I have to run `sudo apt-get update && sudo apt-get install apt-transport-https`, then add new apt-key (commands above), and at the end install ralph-core (`sudo apt-get install ralph-core`). So IMO this should be placed before adding new apt-key.

cdb-fr, line 19: I suppose it can be placed before adding the cert key. 
But on my side I downloaded the key before and that created no issue.
The only trouble occurs if you run apt-get update before installing the secure transport.

mkurek, line 19: yes - `apt-get update` definitely shouldn't be executed before installing https transport. @cdb-fr - we've made some changed to the docs recently - could you apply change from this conversation and make it mergable?

cdb-fr, line 19: I'd be happy to do so but don't know how to "apply change" ?

mkurek, line 19: @cdb-fr I mean to move info about secure transport installation before adding new ralph's apt-key - it'd be more straightforward (and will not crash if someone will execute `sudo apt-get update` in a meanwhile)

vi4m:
@mkurek please commit it by yourself. Closing.

PR #2184, opened 2016-01-28
Remove permissions from demo data command and add provisioning_service

szok:
Bugfix permission codename in demodata command


ar4s:
:ok:

mkurek:
:ok:

PR #2191, opened 2016-02-01
Fixed DataCenter Asset export

szok:
Connected to #2189

PR #2192, opened 2016-02-01
Fixed ralph import/export

szok:
By default, ralph import/export tries to match ralph 2.0 import models as well - it should be only done by clear user intent, not default behavior


src/ralph/assets/models/assets.py
```diff
@@ -358,7 +358,7 @@ class Asset(AdminAbsoluteUrlMixin, BaseObject):
     )
 
     def __str__(self):
-        return self.hostname
+        return self.hostname if self.hostname else ''
```
mkurek: `self.hostname or ''`?

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -71,6 +71,12 @@ def add_arguments(self, parser):
             default='utf-8',
             help="Output encoding",
         )
+        parser.add_argument(
+            '-f', '--from_ralph',
```
mkurek: i'd rather name this argument `--use-imported-id` or `--match-fk-by-imported-id` and use this as a boolean flag. It doesn't have to be imported from Ralph 2 - someone could make exporter from he's own system as well and use his ids during imports.

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -170,6 +176,9 @@ def from_file(self, options):
             self.stdout.write('Done\n')
 
     def handle(self, *args, **options):
+        if options.get('from_ralph') == '2':
+            settings.IMPORT_DATA_FROM_OLD_RALPH = True
```
mkurek: look above

src/ralph/data_importer/widgets.py
```diff
@@ -135,11 +136,16 @@ class ImportedForeignKeyWidget(widgets.ForeignKeyWidget):
...
-                result = self.model.objects.filter(
-                    pk=int(imported_obj.object_pk)
-                ).first()
+            if settings.IMPORT_DATA_FROM_OLD_RALPH:
+                content_type, imported_obj = get_imported_obj(
+                    self.model, value
+                )
+                if imported_obj:
+                    result = self.model.objects.filter(
+                        pk=int(imported_obj.object_pk)
```
mkurek: don't know if mapping it always to int is a good idea - what if we'll use for example uid somewhere?

src/ralph/assets/models/assets.py
mkurek: btw this should be merged in #2191 am i right?

szok: Yes

src/ralph/data_importer/widgets.py
szok: Yes, I know, but this is a bigger problem that we need to discuss.

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -71,6 +71,13 @@ def add_arguments(self, parser):
             default='utf-8',
             help="Output encoding",
         )
+        parser.add_argument(
+            '--use-imported-id',
```
mkurek: add info about this flag to migration documentation.

src/ralph/data_importer/mixins.py
```diff
@@ -1,16 +1,60 @@
...
 
 
 class ImportForeignKeyMixin(object):
 
     """ImportForeignKeyMixin class for django import-export resources."""
 
+    def __init__(self, *args, **kwargs):
+        if not settings.USE_IMPORTED_ID:
+            update_fields = []
+            for name, field in self.fields.items():
```
mkurek: some comment what is going on here will be appreciated

src/ralph/data_importer/mixins.py
```diff
@@ -1,16 +1,60 @@
...
 from ralph.data_importer.models import ImportedObjects
+from ralph.data_importer.widgets import ForeignKeyIdWidget
 
 
 class ImportForeignKeyMixin(object):
 
     """ImportForeignKeyMixin class for django import-export resources."""
 
+    def __init__(self, *args, **kwargs):
+        if not settings.USE_IMPORTED_ID:
```
mkurek: don't we need this logic (str field) only during export?

src/ralph/data_importer/management/commands/importer.py
mkurek: and i know that i was suggesting `use-imported-id` before, but what do you think about `map-imported-id` (or even `map-imported-id-to-new-id`) - wouldn't it be cleaner and more self-descriptive?

src/ralph/data_importer/mixins.py
```diff
@@ -44,5 +88,20 @@ def export_field(self, field, obj):
...
+            value = getattr(obj, self.get_field_name(field), None)
+            return str(value) if value else ''
         return super().export_field(field, obj)
+
+    def get_field_name(self, field):
+        """
+        Returns field name for given field.
+        """
+        for field_name, f in self.fields.items():
+            if field_name.endswith('_str'):
```
mkurek: are you sure this is working correctly? in this first if any information about `field` passed to this function is used

src/ralph/data_importer/widgets.py
```diff
@@ -70,6 +71,20 @@ def clean(self, value):
...
+    def __init__(self, model):
+        self.model = model
+
+    def clean(self, value):
+        if value:
+            return self.model.objects.filter(pk=value).first()
+        return None
+
+    def render(self, value):
+        return value.pk
```
mkurek: what if value is None?

src/ralph/data_importer/widgets.py
mkurek: At least on mysql this is working:

```
In [2]: Warehouse.objects.filter(pk=1)
Out[2]: <Warehouse: DC1>

In [3]: Warehouse.objects.filter(pk="1")
Out[3]: <Warehouse: DC1>
```

We should check how django handles str and int here (maybe all ints are mapped to str?)

src/ralph/admin/mixins.py
```diff
@@ -17,6 +17,8 @@
 from mptt.admin import MPTTAdminForm, MPTTModelAdmin
 from reversion import VersionAdmin
 
+from import_export.admin import ImportExportModelAdmin
```
mkurek: why do you just "import" this?

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -11,6 +11,7 @@
 from django.apps import apps
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+
```
mkurek: flake reject it

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -71,6 +72,13 @@ def add_arguments(self, parser):
             default='utf-8',
             help="Output encoding",
         )
+        parser.add_argument(
+            '--imported-id-to-new-id',
```
mkurek: maybe `--map-imported-id-to-new-id`? :)

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -71,6 +72,13 @@ def add_arguments(self, parser):
...
+            dest='imported_id_to_new_id',
+            default=False,
+            action='store_true',
+            help="Import data from ralph version.",
```
mkurek: Use it when importing data from Ralph 2.

src/ralph/data_importer/resources.py
```diff
@@ -28,12 +32,26 @@
...
+
 
 class DefaultResource(ImportForeignKeyMixin, resources.ModelResource):
     pass
 
 
-class AssetModelResource(ImportForeignKeyMixin, resources.ModelResource):
+class AssetModelResource(
+    ImportForeignKeyMixin,
+    resources.ModelResource,
```
mkurek, line 47: instead of using metaclass here, what do you think about creating new class RalphModelResource that will use this metaclass and inherit from it here?

src/ralph/data_importer/mixins.py
```diff
@@ -1,10 +1,41 @@
...
+
+
+class ImportForeignKeyMeta(type):
+    def __new__(cls, name, bases, attrs):
+        """
+        Add additional fields to the new_class.fields with *_str
+        for display ForeignKey fields.
+        """
+        new_class = super().__new__(cls, name, bases, attrs)
+        export_class = super().__new__(cls, name, bases, attrs)
```
mkurek: export name could be for example `name + 'Exporter'`

src/ralph/data_importer/mixins.py
```diff
@@ -1,10 +1,41 @@
...
+        update_fields = []
+        for name, field in new_class.fields.items():
+            if isinstance(field.widget, widgets.ForeignKeyWidget):
+                update_fields.append((name, deepcopy(field)))
```
mkurek: why do we need deepcopy here?

src/ralph/data_importer/mixins.py
```diff
@@ -1,10 +1,42 @@
...
+
+
+class ImportForeignKeyMeta(type):
+    def __new__(cls, name, bases, attrs):
+        """
+        Add additional fields to the export_class.fields with *_str
+        for display ForeignKey fields.
+        """
+        new_class = super().__new__(cls, name, bases, attrs)
+        export_class = super().__new__(
```
mkurek, line 24: add some comment why do we call super second time here.

src/ralph/data_importer/tests/test_commands.py
```diff
@@ -26,6 +27,7 @@ class DataImporterTestCase(TestCase):
     """TestCase data importer command."""
 
     def setUp(self):  # noqa
+        settings.USE_NEW_IMPORTED_ID = True
```
mkurek: use `overwrite_settings` decorator

src/ralph/data_importer/management/commands/importer.py
```diff
@@ -170,6 +177,8 @@ def from_file(self, options):
             self.stdout.write('Done\n')
 
     def handle(self, *args, **options):
+        if options.get('map_imported_id_to_new_id'):
+            settings.USE_NEW_IMPORTED_ID = True
```
mkurek: name it the same as the parameter: `MAP_IMPORTED_ID_TO_NEW_ID`

src/ralph/data_importer/widgets.py
```diff
@@ -135,11 +156,16 @@ class ImportedForeignKeyWidget(widgets.ForeignKeyWidget):
...
-            if imported_obj:
-                result = self.model.objects.filter(
-                    pk=int(imported_obj.object_pk)
-                ).first()
+            if settings.USE_NEW_IMPORTED_ID:
+                content_type, imported_obj = get_imported_obj(
+                    self.model, value
+                )
+                if imported_obj:
+                    result = self.model.objects.filter(
```
mkurek: just overwrite value by `imported_obj.object_pk` to not duplicate this code

src/ralph/data_importer/mixins.py
```diff
@@ -1,10 +1,60 @@
...
+        new_class = super().__new__(cls, name, bases, attrs)
+        # Generate second class only for export which has added
+        # additional *_str fields
+        export_class = super().__new__(
+            cls, '{}Exporter'.format(name), bases, attrs
+        )
+        update_fields = []
+        for name, field in new_class.fields.items():
+            if isinstance(field.widget, widgets.ForeignKeyWidget):
+                update_fields.append((name, field))
```
mkurek: this line is the same for every condition here - move it before `if`

mkurek:
@szok fix this minor issue and it's :ok: for me

ghost:
:ok:

src/ralph/settings/base.py
```diff
@@ -267,3 +267,5 @@ def os_env_true(var, default=''):
 MY_EQUIPMENT_LINKS = json.loads(os.environ.get('MY_EQUIPMENT_LINKS', '[]'))
 MY_EQUIPMENT_REPORT_FAILURE_URL = os.environ.get('MY_EQUIPMENT_REPORT_FAILURE_URL', '')  # noqa
 MY_EQUIPMENT_SHOW_BUYOUT_DATE = os_env_true('MY_EQUIPMENT_SHOW_BUYOUT_DATE')
+
+MAP_IMPORTED_ID_TO_NEW_ID = False
```
vi4m, line 271: Really easy to understand for new developer who came across this line. Really supportive.
In addition, we should avoid making comments everywhere, thus reducing network costs.

PR #2199, opened 2016-02-05
WIP Fixed generated transitions history

szok:


src/ralph/lib/transitions/models.py
```diff
@@ -427,6 +458,7 @@ class TransitionsHistory(TimeStampMixin):
     attachment = models.ForeignKey(Attachment, blank=True, null=True)
     kwargs = JSONField()
     actions = JSONField()
+    diff = JSONField(blank=True, null=True)
```
mkurek, line 461: why do we need another field? couldn't we store it in one, ex. kwargs?

PR #2200, opened 2016-02-08
Fixed generated transitions history

szok:
Fixed generated transitions history


mkurek:
what about documentation? :)

ar4s:
:ok:

mkurek:
:ok:

PR #2206, opened 2016-02-10
Revert tabular inline first column

mkurek:
Revert c0a006e change which breaks inline saving (removes hidden field with id)


ar4s:
:ok:

szok:
:ok:

PR #2210, opened 2016-02-15
many2many in filters

ghost:


src/ralph/admin/filters.py
```diff
@@ -277,21 +277,28 @@ class RelatedAutocompleteFieldListFilter(RelatedFieldListFilter):
     """Filter for Foregin key field."""
 
     template = "admin/filters/related_filter.html"
-    empty_value = '##@_empty_@##'
+    empty_value = 0
```
mkurek: move this to some "global" constant in some module and reuse it in others

ar4s: :+1:

ghost:
@mkurek @ar4s done

ar4s:
:ok:

PR #2211, opened 2016-02-15
Startswith and icontains in polymorphic API

szok:
Startswith and icontains in polymorphic API

### DIFF
--- a/src/ralph/api/viewsets.py
+++ b/src/ralph/api/viewsets.py
@@ -146,6 +146,47 @@ class RalphAPIViewSet(
             self.extend_filter_fields = {}
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


ghost:
:ok:

mkurek:
we need unit tests for that :)

src/ralph/api/viewsets.py
```diff
@@ -124,6 +124,7 @@ class RalphAPIViewSet(
     metaclass=RalphAPIViewSetMetaclass
 ):
     extend_filter_fields = None
+    polymorphic_limit = 50
```
mkurek: why do we need limit here? isn't it handled by api pagination?

mkurek:
:ok:

PR #2212, opened 2016-02-15
[M2M] Small fixes 2

ghost:
fixes
- cancel submit when enter key is pressed on filtering view
- cancel scrolling on up or down key

### DIFF
--- a/src/ralph/admin/autocomplete.py
+++ b/src/ralph/admin/autocomplete.py
@@ -211,7 +211,7 @@ class AutocompleteList(SuggestView):
                 'tooltip': getattr(obj, 'autocomplete_tooltip', None)
             } for obj in self.get_queryset(user)
         ]
-        if self.request.GET.get('prependEmpty', 'false') == 'true':
+        if self.request.GET.get('prepend-empty', 'false') == 'true':
             results.insert(0, {
                 'pk': self.empty_value,
                 '__str__': '<empty>',
--- a/src/ralph/admin/autocomplete.py
+++ b/src/ralph/admin/autocomplete.py
@@ -13,6 +13,7 @@ from ralph.admin.helpers import get_admin_url
 from ralph.admin.sites import ralph_site
 from ralph.lib.permissions.models import PermissionsForObjectMixin
 
+AUTOCOMPLETE_EMPTY_VALUE = 0
 QUERY_PARAM = 'q'
 DETAIL_PARAM = 'pk'
 QUERY_REGEX = re.compile(r'[.| ]')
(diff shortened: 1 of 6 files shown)


ar4s:
:ok:

src/ralph/admin/autocomplete.py
```diff
@@ -66,17 +67,7 @@ def get(self, request, *args, **kwargs):
...
-                'pk': obj.pk,
-                '__str__': str(obj),
-                'label': getattr(obj, 'autocomplete_str', None),
-                'edit_url': '{}?_popup=1'.format(
-                    get_admin_url(obj, 'change')
-                ) if can_edit else None,
-                'tooltip': getattr(obj, 'autocomplete_tooltip', None)
-            } for obj in self.get_queryset(request.user)
-        ]
+        results = self.get_results(request.user, can_edit)
```
mkurek, line 83: shouldn't this method be implemented here as well (or `abc.abstractmethod`)?

src/ralph/admin/autocomplete.py
```diff
@@ -189,6 +201,25 @@ def get_base_ids(self, model, value):
         queryset = self.get_query_filters(queryset, value, search_fields)
         return queryset[:self.limit].values_list('pk', flat=True)
 
+    def get_results(self, user, can_edit):
+        results = [
```
mkurek: this list comprehension could be at least separated to another function (this is the same code as above).

src/ralph/admin/filters.py
```diff
@@ -277,21 +279,28 @@ class RelatedAutocompleteFieldListFilter(RelatedFieldListFilter):
...
-                queryset = queryset.filter(**{self.field_path: value})
+        if not value:
+            ids = []
+        else:
+            ids = map(int, value.split(','))
+        q_param = models.Q()
+        for id_ in ids:
+            if id_ == self.empty_value:
+                q_param |= Q(**{'{}__isnull'.format(self.field_path): True})
+            q_param |= Q(**{self.field_path: id_})
```
mkurek: shouldn't this be in else block?

src/ralph/admin/autocomplete.py
ghost, line 83:
```
    def get_results(self, user, can_edit):                                                                                                                                                                                          
        raise NotImplementedError    
```

Would it be enough?

mkurek, line 83: sure

src/ralph/admin/autocomplete.py
```diff
@@ -92,25 +96,35 @@ def get_urls(self):
...
         class Detail(SuggestView):
+            empty_value = AUTOCOMPLETE_EMPTY_VALUE
             model = outer_model
 
             def dispatch(self, request, *args, **kwargs):
                 self.pk = request.GET.get(DETAIL_PARAM, None)
                 if not self.pk:
                     return HttpResponseBadRequest()
-                self.pks = self.pk.split(',')
+                self.values = list(map(int, self.pk.split(',')))
```
mkurek: i'm not sure if mapping to int is good idea here - what if we'll have uid somewhere in the future?

ghost: Then much more should be changed, I'd rather not do it now.

mkurek:
:ok:

szok:
:ok:

PR #2213, opened 2016-02-15
Add filter to management ip and hostname

szok:
Add filter to data center asset:
- management IP
- management hostname


mkurek:
:ok:

ar4s:
:ok:

PR #2216, opened 2016-02-16
[M2M] Fixed comments

ghost:


src/ralph/admin/static/auto-complete-helpers.js
```diff
@@ -0,0 +1,73 @@
+function getCallerNode(win) {
+    var name = windowname_to_id(win.name);
+    var elem = document.querySelector('#' + name);
+    return elem;
+}
+
+function dismissRelatedLookupPopup(win, chosenId) {
+    debugger;
```
mkurek: remove it. and every other debugger here :)

src/ralph/admin/static/auto-complete-helpers.js
```diff
@@ -0,0 +1,73 @@
...
+    if (elem.nodeName.toUpperCase() === 'AUTO-COMPLETE') {
+        elem.updateById(newId);
+    } else if (elem) {
+        var elemName = elem.nodeName.toUpperCase();
+        if (elemName == 'SELECT') {
+            o = new Option(newRepr, newId);
+            elem.options[elem.options.length] = o;
+            o.selected = true;
+        } else if (elemName == 'INPUT') {
+            if (elem.className.indexOf('vManyToManyRawIdAdminField') != -1 && elem.value) {
```
mkurek, line 43: this code seems very similar to the one in `dismissRelatedLookupPopup` - could you extract it to one function?

src/ralph/admin/widgets.py
```diff
@@ -122,10 +127,11 @@ def render_options(self, choices, selected_choices):
 
 
 class AutocompleteWidget(forms.TextInput):
+    separator = ','
```
mkurek: maybe `multivalue_separator`?

src/ralph/admin/widgets.py
```diff
@@ -156,14 +161,63 @@ def get_related_url(self, info, action, *args):
...
+        polymorphic_descendants = getattr(
+            self.rel_to, '_polymorphic_descendants', []
+        )
+
+        if polymorphic_descendants:
+            # Check if model after which we are looking for is polymorphic
+            # if they are also looking for the models of its dependencies
+            # or by limit_models defined in model
+            polymorphic_models = polymorphic_descendants
+            limit_models = getattr(self.field, 'limit_models', [])
```
mkurek, line 198: could you add some comment, that usually all polymorphic descendants are processed here, but they could be limited only to particular models using `limit_models` attribute of field?

src/ralph/admin/widgets.py
```diff
@@ -156,14 +161,63 @@ def get_related_url(self, info, action, *args):
...
-            print('name', name)
+        if self.multi:
             if value:
-                value = value.split(',')
+                value = value.split(self.separator)
             else:
                 value = []
         return value
 
+    def get_search_fields(self):
```
mkurek, line 188: we should think about caching value of this function

src/ralph/admin/widgets.py
```diff
@@ -172,14 +226,13 @@ def render(self, name, value, attrs=None):
...
-            value = ','.join(force_text(v) for v in value)
+            attrs['multi'] = 'true'
+            value = self.separator.join(force_text(v) for v in value)
         else:
             value = value or ""
 
-        searched_fields = self.admin_site._registry[self.rel_to].search_fields
-        searched_fields_text = ', '.join(sorted(searched_fields))
-        searched_fields_info = "Search by: {}".format(searched_fields_text)
+        searched_fields = self.get_search_fields()
```
mkurek: `searched_fields` -> `search_fields`

src/ralph/admin/widgets.py
```diff
@@ -172,14 +226,13 @@ def render(self, name, value, attrs=None):
...
+        searched_fields_info = self.render_search_fields_info(dict(searched_fields))
```
mkurek: ditto

src/ralph/admin/widgets.py
```diff
@@ -156,14 +161,63 @@ def get_related_url(self, info, action, *args):
...
+                        related_model, field_name
+                    )
+                    key = str(related_model._meta.verbose_name)
+                    searched_fields_tooltip[key].append(
+                        str(field.verbose_name)
+                    )
+        else:
+            for field_name in self.admin_site._registry[self.rel_to].search_fields:
+                try:
+                    field = get_field_by_relation_path(
```
mkurek: this block is the same as above - you could extract it to some function and call it here and above (to process search fields of single model)

src/ralph/admin/static/auto-complete-helpers.js
ghost: done

src/ralph/admin/widgets.py
ghost: done

src/ralph/admin/widgets.py
ghost, line 198: I think such explanations should be included here https://github.com/allegro/ralph/blob/ng/src/ralph/lib/mixins/fields.py#L69 (where feature is defined) rather then every its use.
Can I add your comment there?

src/ralph/admin/widgets.py
ghost: done

src/ralph/admin/widgets.py
ghost: done

src/ralph/admin/widgets.py
ghost, line 198: done

src/ralph/admin/widgets.py
ghost: done

src/ralph/admin/static/auto-complete-helpers.js
ghost, line 43: I can't see the pattern, could you be more precise? :)

src/ralph/admin/widgets.py
ghost, line 188: Should I decorate it with `@cached_property`?

szok:
:ok:

mkurek:
:ok:

PR #2217, opened 2016-02-16
LOG_FILEPATH introduced to not write into sources location

vi4m:
Could be redefined from: 
- docker 
- deb package
- generic installation

### DIFF
--- a/src/ralph/settings/base.py
+++ b/src/ralph/settings/base.py
@@ -11,6 +11,7 @@ def os_env_true(var, default=''):
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 SECRET_KEY = os.environ.get('SECRET_KEY', 'CHANGE_ME')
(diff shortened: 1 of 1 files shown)


src/ralph/settings/base.py
```diff
@@ -11,6 +11,7 @@ def os_env_true(var, default=''):
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 SECRET_KEY = os.environ.get('SECRET_KEY', 'CHANGE_ME')
+LOG_FILEPATH = os.environ.get('LOG_FILEPATH', '/tmp/runtime.log')
```
mkurek, line 15: What do you think about `/tmp/ralph.log`?

vi4m, line 15: good idea.

mkurek:
:ok:

ghost:
:ok:

PR #2219, opened 2016-02-17
Add new Tags filter

szok:


szok:
Q object does not work.

src/ralph/admin/tests/tests_filters.py
```diff
@@ -162,6 +166,32 @@ def test_text_filter_contains(self):
 
         self.assertEqual(self.dca_1.pk, queryset.first().pk)
 
+    def test_tags_filter(self):
+        tags_filter = TagsListFilter(
+            field=DataCenterAsset._meta.get_field('tags'),
+            request=None,
+            params={'tags': 'tag1 | tag2'},
```
ghost: Pipe separator as logic `and` is misleading IMO, but it was introduced before.

ghost:
:ok:

src/ralph/admin/tests/tests_filters.py
mkurek: +1

src/ralph/admin/filters.py
```diff
@@ -253,10 +255,21 @@ def queryset(self, request, queryset):
...
         return ({
             'current_value': self.value(),
-            'parameter_name': self.field_path
+            'parameter_name': self.field_path,
+            'separators': self.separators
         },)
 
 
+class TagsListFilter(TextListFilter):
+    separators = '"," or ";"'
```
mkurek: according to your convention above, only comma is valid here

src/ralph/admin/filters.py
```diff
@@ -20,7 +20,8 @@
 from ralph.admin.autocomplete import DETAIL_PARAM, QUERY_PARAM
 from ralph.admin.helpers import get_field_by_relation_path
 
-SEARCH_SEPARATORS_REGEX = re.compile(r'[;|]')
+SEARCH_OR_SEPARATORS_REGEX = re.compile(r'[;|]')
+SEARCH_AND_SEPARATORS_REGEX = re.compile(r'[,]')
```
mkurek: i'm afraid that this could be a little bit confusing (especially for users). Ex. using comma as AND separator, but semicolon as OR separator.

szok: What do you propose to use for OR and AND?

mkurek:
:ok:

PR #2227, opened 2016-02-19
Fixed 500 errors in transitions

mkurek:
- When object does not have `get_available_transitions_for_<field>` method, `None` was used before and later called as a function - i've replaced it with lambda
- When model does not have `transition_models`, None was returned which was iterated by later - now empty list is returned

These bugs appears on demo instance during preparation to FOSDEM.

### DIFF
--- a/src/ralph/lib/transitions/templatetags/transitions_tags.py
+++ b/src/ralph/lib/transitions/templatetags/transitions_tags.py
@@ -9,7 +9,7 @@ register = template.Library()
 def available_transitions(context, obj, field):
     """Render available transitions for instance."""
     get_available_transitions = getattr(
-        obj, 'get_available_transitions_for_{}'.format(field), None
+        obj, 'get_available_transitions_for_{}'.format(field), lambda user: []
     )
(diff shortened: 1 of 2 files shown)


src/ralph/lib/transitions/templatetags/transitions_tags.py
```diff
@@ -9,7 +9,7 @@
 def available_transitions(context, obj, field):
     """Render available transitions for instance."""
     get_available_transitions = getattr(
-        obj, 'get_available_transitions_for_{}'.format(field), None
+        obj, 'get_available_transitions_for_{}'.format(field), lambda user: []
```
ghost, line 12: `lambda: []`?

mkurek, line 12: see line 16

ghost, line 12: :+1:

ghost:
:ok:

szok:
:ok:

PR #2229, opened 2016-02-19
Add domains import resources

szok:


src/ralph/data_importer/widgets.py
```diff
@@ -182,11 +182,14 @@ def clean(self, value):
...
         try:
-            value = value.split("|")  # service, enviroment
-            value = ServiceEnvironment.objects.get(
-                service__name=value[0],
-                environment__name=value[1]
-            )
+            if value.isdigit():
+                value = ServiceEnvironment.objects.get(pk=value)
+            else:
+                service, enviroment = value.split("|")
```
mkurek, line 188: we should catch `ValueError` here as well

src/ralph/data_importer/widgets.py
```diff
@@ -182,12 +182,15 @@ def clean(self, value):
...
+                value = ServiceEnvironment.objects.get(
+                    service__name=service,
+                    environment__name=enviroment
+                )
+        except (ValueError, ServiceEnvironment.DoesNotExist):
             value = None
```
mkurek, line 194: shouldn't we raise exception here? Is this row going to be imported - in my opinion it shouldn't

mkurek:
:ok:

src/ralph/data_importer/widgets.py
ghost, line 194: :+1:

ghost:
:ok:

PR #2231, opened 2016-02-19
Change get_queryset in django popup view.

szok:
In method get_queryset in RalphChangeList checked whether the request is a django POPUP, then checked whether the model has a method get_queryset_autocomplete and we turned back.
Then we overwrite queryset in POPUP and we return queryset from autocomplete and not from Django ChangeList.

### DIFF
--- a/src/ralph/admin/tests/tests_functional.py
+++ b/src/ralph/admin/tests/tests_functional.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 from ddt import data, ddt, unpack
+from django.contrib.admin.options import IS_POPUP_VAR
+from django.contrib.admin.views.main import SEARCH_VAR
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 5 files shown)


src/ralph/licences/models.py
```diff
@@ -231,10 +231,12 @@ def free(self):
     free._permission_field = 'number_bought'
 
     @classmethod
-    def get_autocomplete_queryset(cls):
+    def get_autocomplete_queryset(cls, queryset=None):
         # filter by ids of licences which could be assigned (are not fully
         # used)
-        return cls.objects_used_free.filter(
+        if queryset is None:
+            queryset = cls.objects_used_free
```
szok: wrong!

mkurek:
Add unit tests for that (ex. for searching in popup)

ar4s:
:ok:

ghost:
:ok:

mkurek:
:ok:

PR #2232, opened 2016-02-19
Add input focus in search bar

szok:
After open site with search bar lost the focus on the input search.
- Add focus to first input in add form
- Add focus to first input in change form
- Add focus to searchbar in change list


mkurek:
:ok:

ar4s:
:ok:

PR #2233, opened 2016-02-19
Limit scan ganeti plugin calls

mkurek:
For ganeti system http_family is Unspecified, so there is no need to try call ip's which http family is different.


ar4s:
:ok:

szok:
:ok:

PR #2235, opened 2016-02-19
Fixed export IPs and linked assets

ar4s:
Default manager doesn't work properly cause we need fetch all objects of
asset's region.


mkurek:
:ok:

szok:
:ok:

PR #2250, opened 2016-02-26
Added m2m for licence in `Assign licence` transition.

ghost:
Before:
User could pick only one licence in `Assign licence` tranision.
Now:
User can pick many licences in `Assign licence` tranision.

### DIFF
--- a/src/ralph/back_office/models.py
+++ b/src/ralph/back_office/models.py
@@ -220,19 +220,22 @@ class BackOfficeAsset(Regionalizable, Asset):
     @classmethod
     @transition_action(
         form_fields={
-            'licence': {
-                'field': forms.CharField(label=_('Licence')),
+            'licences': {
+                'field': forms.ModelMultipleChoiceField(
+                    queryset=Licence.objects.all(), label=_('Licence')
+                ),
                 'autocomplete_field': 'licence',
-                'autocomplete_model': 'licences.BaseObjectLicence'
+                'autocomplete_model': 'licences.BaseObjectLicence',
+                'widget_options': {'multi': True},
             }
         }
     )
     def assign_licence(cls, instances, request, **kwargs):
         for instance in instances:
-            BaseObjectLicence.objects.get_or_create(
-                base_object=instance,
-                licence=Licence.objects.get(pk=kwargs['licence'])
-            )
+            for obj in kwargs['licences']:
+                BaseObjectLicence.objects.get_or_create(
+                    base_object=instance, licence_id=obj.id,
+                )
 
     @classmethod
     @transition_action(
--- a/src/ralph/back_office/tests/test_models.py
+++ b/src/ralph/back_office/tests/test_models.py
@@ -27,6 +27,7 @@ from ralph.lib.transitions.models import (
     TransitionNotAllowedError
 )
 from ralph.lib.transitions.tests import TransitionTestCase
+from ralph.licences.tests.factories import LicenceFactory
 from ralph.reports.factories import ReportTemplateFactory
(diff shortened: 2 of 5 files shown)


src/ralph/back_office/models.py
```diff
@@ -229,10 +231,10 @@ def assign_user(cls, instances, request, **kwargs):
     )
     def assign_licence(cls, instances, request, **kwargs):
         for instance in instances:
-            BaseObjectLicence.objects.get_or_create(
-                base_object=instance,
-                licence=Licence.objects.get(pk=kwargs['licence'])
-            )
+            for pk in kwargs['licence']:
+                BaseObjectLicence.objects.get_or_create(
+                    base_object=instance, licence=Licence.objects.get(pk=pk)
```
mkurek: why not `base_object=instance, licence_id=pk`?

src/ralph/lib/transitions/models.py
```diff
@@ -101,8 +101,11 @@ def _get_history_dict(data, instance, runned_funcs):
                     value = str(field.rel.to.objects.get(pk=v))
             except FieldDoesNotExist:
                 field = func.form_fields[k]['field']
-                if isinstance(field, forms.ChoiceField):
+                if type(field) == forms.ChoiceField:
```
mkurek: i'd prefer not to use `type` comparison. you could swap these if-else to first check `if isinstance(field, forms.ModelMultipleChoiceField):`, then `elif isinstance(field, forms.ChoiceField):`

src/ralph/lib/transitions/views.py
```diff
@@ -62,10 +62,15 @@ def form_fields_from_actions(self):
                     field = model._meta.get_field(
                         options['autocomplete_field']
                     )
+                    widget_options = {}
+                    autocomplete_name = options.get('autocomplete_field', None)
+                    if autocomplete_name in ['licence']:
```
mkurek: http://www.nooooooooooooooo.com/
please add this as another parameter to transitions fields definition or deduct if from field type

src/ralph/back_office/models.py
```diff
@@ -221,18 +221,21 @@ def assign_user(cls, instances, request, **kwargs):
     @transition_action(
         form_fields={
             'licence': {
```
mkurek: licences?

src/ralph/back_office/tests/test_models.py
```diff
@@ -296,6 +297,20 @@ def test_convert_to_data_center_asset(self):
...
+    def test_assign_many_licences(self):
+        asset = BackOfficeAssetFactory()
+        licences = [LicenceFactory() for i in range(2)]
+        self.assertFalse(asset.licences.all())
+
+        BackOfficeAsset.assign_licence(
+            instances=[asset], request=None, licence=licences,
+        )
+
+        self.assertEqual(
```
mkurek: I'd use https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.assertCountEqual here

src/ralph/lib/transitions/views.py
```diff
@@ -62,10 +62,12 @@ def form_fields_from_actions(self):
                     field = model._meta.get_field(
                         options['autocomplete_field']
                     )
+                    widget_options = options.get('widget_options', {})
                     options['field'].widget = AutocompleteWidget(
                         field=field,
                         admin_site=ralph_site,
                         request=self.request,
+                        **widget_options
```
mkurek: why not just `**options.get('widget_options', {})`?

szok:
:ok:

mkurek:
:ok:

PR #2253, opened 2016-02-29
Operations fixes

mkurek:
- limit possible type choices by Operation main type (Failure, Change etc)
- fixed doubled popup when changing Operation from DCA view
- changed base objects widget to M2M
- fixed Django Debug Toolbar to work with iron-flex-layout
- add popup to Operation change link in DCA tab

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,5 +1,5 @@
 -r test.txt
-django-debug-toolbar
+git+http://github.com/django-debug-toolbar/django-debug-toolbar.git@0ed1538374d3dc41d51fc562b4e8be6ce8208552
 Werkzeug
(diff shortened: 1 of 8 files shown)


src/ralph/back_office/templates/backofficeasset/add_form.html
```diff
@@ -5,6 +5,7 @@
     $(function () {
       "use strict";
       $('#id_model').change(function() {
+        console.log( $(this).val());
```
mkurek: remove

ghost:
:ok:

ar4s:
:ok:

szok:
:ok:

Issue #2255, opened 2016-02-29
TypeError at /assets/service/add/


renki81:
Hi
I become this error when i will add a new Service:

```
TypeError at /assets/service/add/
'NoneType' object is not iterable
Request Method: GET
Request URL:    http://myip/assets/service/add/?_popup=1&_changelist_filters=_to_field%3Did%26_popup%3D1&_to_field=id
Django Version: 1.8.9
Exception Type: TypeError
Exception Value:    
'NoneType' object is not iterable
Exception Location: /opt/ralph/ralph-core/lib/python3.4/site-packages/ralph/admin/widgets.py in render, line 234
Python Executable:  /opt/ralph/ralph-core/bin/python
Python Version: 3.4.3
Python Path:    
['/opt/ralph/ralph-core/bin',
 '/opt/ralph/ralph-core/lib/python3.4',
 '/opt/ralph/ralph-core/lib/python3.4/plat-x86_64-linux-gnu',
 '/opt/ralph/ralph-core/lib/python3.4/lib-dynload',
 '/usr/lib/python3.4',
 '/usr/lib/python3.4/plat-x86_64-linux-gnu',
 '/opt/ralph/ralph-core/lib/python3.4/site-packages']
Server time:    Mon, 29 Feb 2016 18:25:33 +0100
Error during template rendering

In template /opt/ralph/ralph-core/lib/python3.4/site-packages/ralph/lib/foundation/templates/foundation_form/form_field.html, error at line 14


'NoneType' object is not iterable
4     <div class="row">
5       <div class="small-3 columns">
6         {% if show_label %}
7           {% label field %}
8         {% endif %}
9       </div>
10      <div class="small-9 end columns {% if field.errors %} error{% endif %}">
11        {% if field.is_readonly %}
12          <span class='read-only'>{{ field.contents|default:"-" }}</span>
13        {% else %}
14  

      {{ admin_field }}



15        {% endif %}
16        {% if admin_field.errors %}
17          {% for error in admin_field.errors %}
18            <small class="error">{{ error }}</small>
19          {% endfor %}
20        {% endif %}
21        {% if admin_field.help_text %}
22          <span class="help-text">
23            <i
24              class="fa fa-lg fa-info-circle has-tip tip-top"


Environment:


Request Method: GET
Request URL: http://myip/assets/service/add/?_popup=1&_changelist_filters=_to_field%3Did%26_popup%3D1&_to_field=id

Django Version: 1.8.9
Python Version: 3.4.3
[...]

PR #2256, opened 2016-02-29
Prevent `object is not iterable` exception

ar4s:
#2255


mkurek:
@xliiv's PR #2250 (https://github.com/allegro/ralph/pull/2250/files#diff-bb7c7123035c5c7c8acda2f278fcf898R232) fixes it to - this is :ok: for me but since his PR will be probably merged tomorrow, maybe we'll stay with his version?

PR #2257, opened 2016-03-01
Handle ValueError in admin related filter

szok:
Unfortunately django admin does not support for filter validation.
TODO:
- [x] tests
- [x] message

### DIFF
--- a/src/ralph/admin/filters.py
+++ b/src/ralph/admin/filters.py
@@ -3,8 +3,10 @@ import re
 from datetime import datetime
 from functools import lru_cache
 
+from django.contrib import messages
 from django.contrib.admin import SimpleListFilter
 from django.contrib.admin.filters import FieldListFilter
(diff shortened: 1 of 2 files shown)


src/ralph/admin/filters.py
```diff
@@ -317,7 +318,11 @@ def queryset(self, request, queryset):
                 q_param |= Q(**{'{}__isnull'.format(self.field_path): True})
             else:
                 q_param |= Q(**{self.field_path: id_})
-        queryset = queryset.filter(q_param)
+        try:
+            queryset = queryset.filter(q_param)
+        except ValueError as e:
+            raise IncorrectLookupParameters(e)
```
mkurek: make test for that

ar4s:
:ok:

mkurek:
:ok:

PR #2260, opened 2016-03-02
Fixed taggit in reversion history

mkurek:
Original taggit returns (not-evaluated) QuerySet from `value_from_object`, which was evaluated in has_changed method (this fetched current data from DB, not initial state before). After changes, Ralph's `value_from_object` returns list of tags, which could be easily compared with current data from form.

### DIFF
--- a/src/ralph/lib/mixins/fields.py
+++ b/src/ralph/lib/mixins/fields.py
@@ -6,8 +6,10 @@ from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.db.models.loading import get_model
 from django.forms.utils import flatatt
+from django.utils import six
 from django.utils.html import format_html, smart_urlquote
 from django.utils.translation import ugettext_lazy as _
+from taggit.forms import TagField
 
 
 class NullableFormFieldMixin(object):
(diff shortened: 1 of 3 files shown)


szok:
:ok:

src/ralph/lib/mixins/models.py
```diff
@@ -71,6 +73,21 @@ def get_absolute_url(self):
         )
 
 
+class TaggableManager(TaggableManagerOriginal):
+    def value_from_object(self, instance):
+        """
+        Fix to work properly with reversion (or more generally, with
+        `django.forms.models.model_to_dict` - TaggableManager is not
+        ManyToManyField instance, so this result of this method is dumped
+        directly to resulting dict.
```
xliiv: dict or list?

src/ralph/lib/mixins/fields.py
```diff
@@ -169,3 +171,38 @@ def get_limit_models(self):
         Returns Model class list from limit_models.
         """
         return [get_model(model) for model in self.limit_models]
+
+
+class TagWidget(forms.TextInput):
+    def render(self, name, value, attrs=None):
+        if value is not None and not isinstance(value, six.string_types):
+            value = ', '.join(sorted([
+                (t if ',' not in t else '"%s"' % t) for t in value
```
ghost, line 180: what are possible values for `value` here?

mkurek, line 180: type of `value` in this method could be either `None`, `str` or `list (iterable)`

mkurek, line 180: https://github.com/alex/django-taggit/blob/develop/taggit/forms.py#L13
https://github.com/alex/django-taggit/blob/develop/taggit/utils.py#L108

ghost, line 180: :ok:

ghost:
:ok:

PR #2261, opened 2016-03-02
Add Permission Admin Mixin to model admins

szok:
Add permission admin mixin to model admins
Change code style.


mkurek:
is there any admin inheriting from `RalphAdmin` that does not inherit from `PermissionAdminMixin`?

mkurek:
:ok:

ar4s:
:ok:

PR #2263, opened 2016-03-03
USE_TZ = True.

fossilet:
We are using the following settings regarding timezone:

```
TIME_ZONE = 'Europe/Warsaw'
USE_TZ = False
```

When `USE_TZ` is False, Django would store timestamps in the timezone of `TIME_ZONE`, and this is how ralph is storing time. After I realised I should change `TIME_ZONE`, even if I changed `TIME_ZONE to` `'Asia/Beijing'`, the stored time are still in the old timezone (GMT+1). How about it that we change `USE_TZ` to `True`, to make sue time is always saved as UTC? This is also [recommended by Django](https://docs.djangoproject.com/es/stable/topics/i18n/timezones/).

### DIFF
--- a/src/ralph/settings/base.py
+++ b/src/ralph/settings/base.py
@@ -124,7 +124,7 @@ LOCALE_PATHS = (os.path.join(BASE_DIR, 'locale'), )
 TIME_ZONE = 'Europe/Warsaw'
 USE_I18N = True
 USE_L10N = True
-USE_TZ = False
+USE_TZ = True
 
 STATIC_URL = '/static/'
 STATICFILES_DIRS = (
--- a/src/ralph/settings/base.py
+++ b/src/ralph/settings/base.py
@@ -121,7 +121,7 @@ LOGIN_URL = '/login/'
 
 LANGUAGE_CODE = 'en-us'
 LOCALE_PATHS = (os.path.join(BASE_DIR, 'locale'), )
-TIME_ZONE = 'Europe/Warsaw'
+TIME_ZONE = os.environ.get('TIME_ZONE', 'Europe/Warsaw')
 USE_I18N = True
 USE_L10N = True
 USE_TZ = True
--- a/src/ralph/virtual/management/commands/openstack_sync.py
+++ b/src/ralph/virtual/management/commands/openstack_sync.py
@@ -2,7 +2,6 @@
 import logging
 import re
 from collections import defaultdict
-from datetime import datetime
 from functools import lru_cache
 
 import reversion as revisions
--- a/src/ralph/virtual/management/commands/openstack_sync.py
+++ b/src/ralph/virtual/management/commands/openstack_sync.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from functools import lru_cache
 
 import reversion as revisions
+from dateutil import parser
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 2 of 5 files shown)


mkurek:
Hi @fossilet. We cannot simply enable `USE_TZ` right now because it breaks existing code. Take a look at travis unit tests - some of them are broken. Its probably caused by using `datetime` core module of Python instead of `django.utils.timezone` (https://docs.djangoproject.com/es/1.9/topics/i18n/timezones/#naive-and-aware-datetime-objects) - will you be able to fix these broken places in our code?

fossilet:
Oops, let me see.

fossilet:
The difference of the objects in the failed test is two hours. I cannot find why. @mkurek can you have a look at it?

ghost:
@fossilet it looks like openstack importer has issue which blocks this pull request.
I'm gonna fix that and we handle your case, ok?

fossilet:
@xliiv great.

ghost:
The bug which I mentioned earlier is because timezone is skipped here:
https://github.com/allegro/ralph/blob/ng/src/ralph/virtual/management/commands/openstack_sync.py#L309
It turned out also that it'd be easier when you merge my changes (then the opposite)
So here's the changes:
https://github.com/allegro/ralph/compare/ng...xliiv:open-stack-tz
A few tests are broken in my change but I think you already fixed them.
If not give some feedback.

fossilet:
@xliiv, there is still a flake error even if I have fixed one.

ar4s:
@fossilet remove this line https://github.com/allegro/ralph/pull/2263/files#diff-4fa30bd6416c1bceb147b0bf9dbf6999R5

fossilet:
@ar4s, I saw the error, but it is in fact used by the newly added code by @xliiv. That's why I asked him to have a look. Maybe its an error of flake8 itself.

mkurek:
@fossilet no it's not used - he replaced `datetime.strptime` by `parser.parse`.

fossilet:
I have fixed it.

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -306,8 +306,7 @@ def _add_server(self, openstack_server, server_id, project):
 
         # workaround - created field has auto_now_add attribute
         new_server.save()
-        new_server.created = datetime.strptime(openstack_server['created'],
-                                               self.DATETIME_FORMAT)
```
mkurek, line 310: `self.DATETIME_FORMAT` is not used anymore - could you delete it in line 49?

src/ralph/settings/base.py
```diff
@@ -124,7 +124,7 @@ def os_env_true(var, default=''):
 TIME_ZONE = 'Europe/Warsaw'
```
mkurek: if you want to use this when installing from Docker or deb you have to pass it as env variable. Could you change it to `TIME_ZONE = os.environ.get('TIME_ZONE', 'Europe/Warsaw')`?

fossilet: This would entail changes to test fixtures that use Warsaw timezones. Should they be changed to using dynamic timezones accordingly?

fossilet:
@mkurek I've added it.

mkurek:
@fossilet thanks. we'll pull and test your code if it works well with our current setup.

mkurek:
Continued in #2391

PR #2265, opened 2016-03-03
Operations improvements

mkurek:
- filter Failures by manufacturer
- improved ticket id field: it accepts and parse url now and display url below field
- fixed Operation import resource
- added autocomplete tooltip with additional info for DataCenterAsset


szok:
:ok:

PR #2266, opened 2016-03-03
Operations type filter fix

mkurek:
Filter by whole subtree of types, not only the selected one (without descendants).


szok:
:ok:

ghost:
:ok:

PR #2270, opened 2016-03-07
Added missing migrations

mkurek:
After change in `tags` field in 5706bd5 there were missing migrations for models using tags.


ghost:
:ok:

PR #2273, opened 2016-03-09
Value repeater in bulk form for autocomplete.

ghost:
Autocomplete now is able to copy values in row by user click.
Added copying empty values.


ar4s:
:ok:

szok:
:ok:

mkurek:
:ok:

PR #2278, opened 2016-03-10
[WIP] DNS records from DNSAAS

ghost:
- [ ] unittests


src/ralph/data_center/admin.py
```diff
@@ -41,6 +41,8 @@
 from ralph.licences.models import BaseObjectLicence
 from ralph.operations.views import OperationViewReadOnlyForExisiting
 from ralph.supports.models import BaseObjectsSupport
+if 'ralph.dns' in settings.INSTALLED_APPS:
```
mkurek: imo we should use dedicated variable for that (ex. 'ENABLE_DNSAAS_INTEGRATION')

PR #2280, opened 2016-03-11
Added additional data to supports export

mkurek:
- added additional data to DC/BO Asset str (SN and barcode)
- added assigned objects count to support export


szok:
:ok:

ghost:
:ok:

PR #2281, opened 2016-03-14
Export and import rack from __str__.

fossilet:
One feature request from our team is that they want to import from human readable representation of racks in CSV file rather than filing them with int ID's. After I implemented this, I find this does not break the models of ralph, since **str** of rack is unique for each one.

### DIFF
--- a/src/ralph/data_importer/resources.py
+++ b/src/ralph/data_importer/resources.py
@@ -21,6 +21,7 @@ from ralph.data_importer.widgets import (
     BaseObjectWidget,
     ImportedForeignKeyWidget,
     NullStringWidget,
+    RackWidget,
     UserManyToManyWidget,
     UserWidget
 )
@@ -233,7 +234,7 @@ class DataCenterAssetResource(RalphModelResource):
     rack = fields.Field(
         column_name='rack',
         attribute='rack',
-        widget=ImportedForeignKeyWidget(physical.Rack),
+        widget=RackWidget(physical.Rack),
     )
     budget_info = fields.Field(
         column_name='budget_info',
--- a/src/ralph/data_importer/widgets.py
+++ b/src/ralph/data_importer/widgets.py
@@ -8,7 +8,7 @@ from import_export import widgets
 from ralph.assets.models.assets import ServiceEnvironment
 from ralph.assets.models.base import BaseObject
 from ralph.back_office.models import BackOfficeAsset
-from ralph.data_center.models.physical import DataCenterAsset
+from ralph.data_center.models.physical import DataCenterAsset, Rack
 from ralph.data_importer.models import ImportedObjects
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


src/ralph/data_importer/resources.py
```diff
@@ -233,7 +234,7 @@ class DataCenterAssetResource(RalphModelResource):
     rack = fields.Field(
         column_name='rack',
         attribute='rack',
-        widget=ImportedForeignKeyWidget(physical.Rack),
+        widget=RackWidget(physical.Rack),
```
szok, line 237: Does it work with CLI? with the flag MAP_IMPORTED_ID_TO_NEW_ID ? (import from old ralph)

src/ralph/data_importer/widgets.py
```diff
@@ -217,3 +217,29 @@ def render(self, value):
...
+            return None
+        else:
+            try:
+                racks = Rack.objects.all()
+                data = ({'id': rack.id, '__str__': str(rack)} for rack in racks)
+                for datum in data:
+                    if datum['__str__'] == value:
+                        value = Rack.objects.get(id=datum['id'])
+                        break
+            except (ValueError, ServiceEnvironment.DoesNotExist):
```
szok: Rack.DoesNotExists

src/ralph/data_importer/resources.py
fossilet, line 237: No, it prints `[08:52:53] WARNING Record with pk 1 not found for model assetmodel of 'model'` and `Some of records were not imported`. It fails too without that flag.

vi4m:
@szok is it ready to merge?

src/ralph/data_importer/widgets.py
```diff
@@ -217,3 +217,29 @@ def render(self, value):
...
+    """Widget for Rack Foreign Key field.
+
+    CSV field format Rack.name
+    """
+    def clean(self, value):
+        if not value:
+            return None
+        else:
+            try:
+                racks = Rack.objects.all()
```
szok, line 232: Clean method is run for each records in CSV file. 
You can do this more performance optimal.

Pleas add some test for:
1. Import from old ralph (which Rack ID and **str**)
2. Numbers of SQL queries. (assertNumQueries)

src/ralph/data_importer/widgets.py
```diff
@@ -217,3 +217,29 @@ def render(self, value):
...
+                data = ({'id': rack.id, '__str__': str(rack)} for rack in racks)
+                for datum in data:
+                    if datum['__str__'] == value:
+                        value = Rack.objects.get(id=datum['id'])
+                        break
+            except (ValueError, Rack.DoesNotExist):
+                value = None
+            return value
+
+    def render(self, value):
```
szok, line 242: Render method must be return ID for export from GUI. Please test this
1. Import/export from GUI
2. Import/export from CLI

szok:
@vi4m  Unfortunately it is not ready to merge. I have added new comments.

PR #2288, opened 2016-03-15
Update and delete DNS API

szok:


src/ralph/dns/views.py
```diff
@@ -33,31 +44,53 @@ def get_dns_records(ipaddresses):
...
     api_results = get_api_result(url)
     if api_results:
         for item in api_results:
             dns_records.append({
-                'pk': 1,
+                'pk': item['id'],
                 'name': item['name'],
                 'type': RecordType.from_name(item['type'].lower()).id,
                 'content': item['content'],
+                'ptr': item['auto_ptr'] == 2
```
mkurek: magic number

src/ralph/dns/views.py
```diff
@@ -33,31 +44,53 @@ def get_dns_records(ipaddresses):
             ('type', 'A'),
```
mkurek: why only A?

src/ralph/dns/views.py
```diff
@@ -33,31 +44,53 @@ def get_dns_records(ipaddresses):
...
             ('limit', 100),
```
mkurek: what if there is more than 100 records for single host? :)

src/ralph/dns/views.py
```diff
@@ -33,31 +44,53 @@ def get_dns_records(ipaddresses):
...
             ('offset', 0)
-        ])  # TODO ?
+        ] + ipaddresses)
     )
     api_results = get_api_result(url)
     if api_results:
```
mkurek: this if isn't necessary imo

src/ralph/dns/views.py
mkurek: another thing: PTR record should be matched to A record when iterating through it, not by `auto_ptr` flag of A record

src/ralph/dns/views.py
```diff
@@ -33,31 +44,53 @@ def get_dns_records(ipaddresses):
...
 
 
 def update_dns_records(records):
-    pass
+    error = False
+    for item in records:
+        url = '/'.join([settings.DNSAAS_URL, 'api/records', str(item['pk'])])
+        data = {
+            'name': item['name'],
+            'type': RecordType.from_id(int(item['type'])).raw,
```
mkurek: you could use `name_from_id`

PR #2289, opened 2016-03-15
DNS records from DNSAAS.

szok:
Added optional DNS feature to data center assets.
See: https://github.com/allegro/django-powerdns-dnssec

### DIFF
--- a/docs/user/dns.md
+++ b/docs/user/dns.md
@@ -1,0 +1,17 @@
+# DNS
+
+## Introduction
+
+Ralph integrated with DNSAAS [PowerDNS](https://github.com/allegro/django-powerdns-dnssec)
+
+
+## Configuration
+
+- ``ENABLE_DNSAAS_INTEGRATION`` - set to True if you want to enable DNSaaS integration
+- ``DNSAAS_URL`` - Url to DNSAAS
+- ``DNSAAS_TOKEN`` - API Token to DNSAAS
+- ``DNSAAS_AUTO_PTR_ALWAYS`` - DNSAAS auto_ptr value, default is 2
+- ``DNSAAS_AUTO_PTR_NEVER`` - DNSAAS auto_ptr value, default is 1
+
+On the edit page of DataCenterAsset will appear a new tab DNS Edit.
+DNS records are matched using DataCenterAssets IP
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -97,7 +97,7 @@ class DNSaaS:
         if request.status_code != 200:
             return request.json()
 
-    @cache()
+    @cache(skip_first=True)
     def get_domain(self, domain_name):
         """
         Return domain URL base on record name.
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -69,7 +69,7 @@ class DNSaaS:
                 })
         return sorted(dns_records, key=lambda x: x['type'])
 
-    def update_dns_records(self, record):
+    def update_dns_record(self, record):
         """
         Update DNS Record in DNSAAS
 
@@ -117,7 +117,7 @@ class DNSaaS:
         if result:
             return result[0]['url']
 
-    def create_dns_records(self, record):
+    def create_dns_record(self, record):
         """
         Create new DNS record.
 
@@ -153,9 +153,9 @@ class DNSaaS:
         if request.status_code != 201:
             return request.json()
 
-    def delete_dns_records(self, record_id):
+    def delete_dns_record(self, record_id):
         """
-        Delete rcords in DNSAAS
+        Delete record in DNSAAS
 
         Args:
             record_ids: ID's to delete
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+import logging
+from urllib.parse import urlencode, urljoin
+
+import requests
+from django.conf import settings
+from django.utils.translation import ugettext_lazy as _
+
+from ralph.dns.forms import RecordType
+from ralph.helpers import cache
+
+logger = logging.getLogger(__name__)
+
+
+class DNSaaS:
+
+    def __init__(self, headers=None):
+        self.session = requests.Session()
+        _headers = {
+            'Authorization': 'Token {}'.format(settings.DNSAAS_TOKEN)
+        }
+        if headers is not None:
+            _headers.update(headers)
+        self.session.headers.update(_headers)
+
+    def get_api_result(self, url):
+        """
+        Returns 'results' from DNSAAS API.
+
+        Args:
+            url: Url to API
+
+        Returns:
+            list of records
+        """
+        request = self.session.get(url)
+        json_data = request.json()
+        api_results = json_data.get('results', [])
+        if json_data.get('next', None):
+            _api_results = self.get_api_result(json_data['next'])
+            api_results.extend(_api_results)
+        return api_results
+
+    def get_dns_records(self, ipaddresses):
+        """Gets DNS Records for `ipaddresses` by API call"""
+        dns_records = []
+        ipaddresses = [('ip', i) for i in ipaddresses]
+        url = urljoin(
+            settings.DNSAAS_URL,
+            'api/records/?{}'.format(
+                urlencode([
+                    ('limit', 100),
+                    ('offset', 0)
+                ] + ipaddresses)
+            )
+        )
+        api_results = self.get_api_result(url)
+        ptrs = set([i['content'] for i in api_results if i['type'] == 'PTR'])
+
+        for item in api_results:
+            if item['type'] in {'A', 'CNAME', 'TXT'}:
+                dns_records.append({
+                    'pk': item['id'],
+                    'name': item['name'],
+                    'type': RecordType.from_name(item['type'].lower()).id,
+                    'content': item['content'],
+                    'ptr': item['name'] in ptrs and item['type'] == 'A',
+                    'owner': settings.DNSAAS_OWNER
+                })
+        return sorted(dns_records, key=lambda x: x['type'])
+
+    def update_dns_records(self, record):
+        """
+        Update DNS Record in DNSAAS
+
+        Args:
+            record: record cleaned data
+
+        Returns:
+            Validation error from API or None if update correct
+        """
+        url = urljoin(
+            settings.DNSAAS_URL, 'api/records/{}/'.format(record['pk'])
+        )
+        data = {
+            'name': record['name'],
+            'type': RecordType.raw_from_id(int(record['type'])),
+            'content': record['content'],
+            'auto_ptr': (
+                settings.DNSAAS_AUTO_PTR_ALWAYS if record['ptr'] and
+                record['type'] == str(RecordType.a.id)
+                else settings.DNSAAS_AUTO_PTR_NEVER
+            ),
+            'owner': settings.DNSAAS_OWNER
+        }
+        request = self.session.patch(url, data=data)
+        if request.status_code != 200:
+            return request.json()
+
+    @cache()
+    def get_domain(self, domain_name):
+        """
+        Return domain URL base on record name.
+
+        Args:
+            domain_name: Domain name
+
+        Return:
+            Domain URL from API or False if not exists
+        """
+        url = urljoin(
+            settings.DNSAAS_URL, 'api/domains/?'.format(
+                urlencode([('name', domain_name)])
+            )
+        )
+        result = self.get_api_result(url)
+        if result:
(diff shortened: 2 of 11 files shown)


src/ralph/dns/views.py
```diff
@@ -0,0 +1,153 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from ralph.admin.views.extra import RalphDetailView
+from ralph.dns.forms import DNSRecordForm, RecordType
+
+logger = logging.getLogger(__name__)
+
+
+DNS_AUTO_PTR_ALWAYS = 2
+DNS_AUTO_PTR_NEVER = 1
```
ghost: Django settings?

src/ralph/dns/templates/dns/dns_edit.html
```diff
@@ -0,0 +1,30 @@
...
+            {% for form in formset %}
+            <tr>
+              {% for field in form %}
+                <td>{{ field }} {{ field.errors }}</td>
+              {% endfor %}
+            </tr>
+            {% endfor %}
+        </table>
+        <br />
+        <input type="submit" class="button" value="{% trans "Save" %}" />
```
mkurek: is it possible to "wrap" it using standard ralph view to have buttons bar on the bottom?

src/ralph/dns/views.py
```diff
@@ -0,0 +1,150 @@
...
+def update_dns_records(records):
+    error = False
+    for item in records:
+        url = '/'.join([settings.DNSAAS_URL, 'api/records', str(item['pk'])])
+        data = {
+            'name': item['name'],
+            'type': RecordType.raw_from_id(int(item['type'])),
+            'content': item['content'],
+            'auto_ptr': (
+                settings.DNS_AUTO_PTR_ALWAYS if item['ptr'] and
```
mkurek: what will happen if:
- there is already attached PTR record to A record
- user will uncheck PTR checkbox

Will PTR record then be deleted from DNSaaS? How passing this `auto_ptr` as False will work for existing A record?

src/ralph/dns/views.py
```diff
@@ -0,0 +1,150 @@
...
+                'content': item['content'],
+                'ptr': item['name'] in ptr_list
+            })
+    return dns_records
+
+
+def update_dns_records(records):
+    error = False
+    for item in records:
+        url = '/'.join([settings.DNSAAS_URL, 'api/records', str(item['pk'])])
```
mkurek: wouldn't it return 302 redirect first to url with `/` at the end?

mkurek:
please add basic documentation (docs) for dnsaas integration - possible settings. Put a note here, that this is not transaction-safe, since it's communicating with external system through the api.

src/ralph/dns/templates/dns/dns_edit.html
szok: Unfortunately, because we don't have django change form

src/ralph/dns/views.py
szok: No, but requests it has  history of "response.history" and supports it.

src/ralph/dns/views.py
szok: This is supported by PowerDNS

```
@receiver(post_save, sender=Record, dispatch_uid='record_create_ptr')
def create_ptr(sender, instance, **kwargs):
    if instance.auto_ptr == AutoPtrOptions.NEVER or instance.type != 'A':
        instance.delete_ptr()
        return
    instance.create_ptr()
```

create_ptr method automatically created ptr.

If user will uncheck PTR, powerdns delete ptr records, 
If already PTR powerdns refresh ptr records. 

@xliiv What do you think? 

https://github.com/allegro/django-powerdns-dnssec/blob/develop/powerdns/models/powerdns.py#L556

src/ralph/dns/views.py
mkurek: IMO it will (i see 301, not 302, but it doesn't make any difference here):

```
curl -v https://<DNSAAS_URL>/api/records/1
...
< HTTP/1.1 301 MOVED PERMANENTLY
...
< Location: https://<DNSAAS_URL>/api/records/1/
```

I'd add `/` at the end and we'll make twice less requests.

src/ralph/dns/views.py
```diff
@@ -0,0 +1,150 @@
...
+                'type': RecordType.from_name(item['type'].lower()).id,
+                'content': item['content'],
+                'ptr': item['name'] in ptr_list
+            })
+    return dns_records
+
+
+def update_dns_records(records):
+    error = False
+    for item in records:
```
mkurek: what will you say about rewriting it using asyncio? We make here many (independent) requests, so the main blocking factor will be network (io) - we could "parallelize" it.

mkurek:
:ok:

docs/user/dns.md
```diff
@@ -0,0 +1,20 @@
...
+
+Ralph integrated with DNSAAS [PowerDNS](https://github.com/allegro/django-powerdns-dnssec)
+
+
+## Configuration
+
+- ``ENABLE_DNSAAS_INTEGRATION`` - set to True if you want to enable DNSaaS integration
+- ``DNSAAS_URL`` - Url to DNSAAS
+- ``DNSAAS_TOKEN`` - API Token to DNSAAS
+- ``DNS_AUTO_PTR_ALWAYS`` - DNSAAS auto_ptr value, default is 2
```
ghost: `DNS_AUTO_PTR_ALWAYS` -> `DNSAAS_AUTO_PTR_ALWAYS`?

docs/user/dns.md
```diff
@@ -0,0 +1,20 @@
...
+- ``DNS_AUTO_PTR_NEVER`` - DNSAAS auto_ptr value, default is 1
```
ghost: `DNS_AUTO_PTR_NEVER` -> `DNSAAS_AUTO_PTR_NEVER`?

docs/user/dns.md
```diff
@@ -0,0 +1,20 @@
...
+
+On the edit page of DataCenterAsset will apper a new tab DNS Edit.
```
ghost: `apper` -> `appear`?

src/ralph/dns/views.py
```diff
@@ -0,0 +1,154 @@
...
+from django.http import HttpResponseRedirect
+from django.utils.translation import ugettext_lazy as _
+
+from ralph.admin.views.extra import RalphDetailView
+from ralph.dns.forms import DNSRecordForm, RecordType
+
+logger = logging.getLogger(__name__)
+
+
+def get_api_kwargs(url, data=None, headers=None):
```
ar4s: please move all `*_api_*` and `*_dns_*` method to class and separated module like `dnsass.py`

src/ralph/dns/dnsaas.py
```diff
@@ -0,0 +1,178 @@
...
+                else settings.DNSAAS_AUTO_PTR_NEVER
+            ),
+            'owner': settings.DNSAAS_OWNER
+        }
+        request = self.session.patch(url, data=data)
+        if request.status_code != 200:
+            return False
+        return True
+
+    @cache()
```
mkurek: this cache here is useless since almost all record will be different. The only cache which makes sense here is caching based on `domain_name` param.

docs/user/dns.md
```diff
@@ -0,0 +1,20 @@
...
+- ``DNSAAS_URL`` - Url to DNSAAS
+- ``DNSAAS_TOKEN`` - API Token to DNSAAS
+- ``DNSAAS_AUTO_PTR_ALWAYS`` - DNSAAS auto_ptr value, default is 2
+- ``DNSAAS_AUTO_PTR_NEVER`` - DNSAAS auto_ptr value, default is 1
+
+On the edit page of DataCenterAsset will appear a new tab DNS Edit.
+DNS records are matched using DataCenterAssets IP
+
+
+> "DNS edit is not transaction-safe, since it's communicating with external system through the api."
```
mkurek: this is obsolete now, since there is separated save button for every form?

src/ralph/dns/dnsaas.py
```diff
@@ -0,0 +1,172 @@
...
+            settings.DNSAAS_URL,
+            'api/records/?{}'.format(
+                urlencode([
+                    ('limit', 100),
+                    ('offset', 0)
+                ] + ipaddresses)
+            )
+        )
+        api_results = self.get_api_result(url)
+        ptr_list = set(
```
mkurek: list suggests that this is a list - maybe `ptrs`?

src/ralph/dns/dnsaas.py
```diff
@@ -0,0 +1,172 @@
...
+        return sorted(dns_records, key=lambda x: x['type'])
+
+    def update_dns_records(self, record):
+        """
+        Update DNS Record in DNSAAS
+
+        Args:
+            record: record cleaned data
+
+        Returns:
```
mkurek: is this valid right now?

src/ralph/dns/views.py
```diff
@@ -0,0 +1,72 @@
...
+        return forms
+
+    def get(self, request, *args, **kwargs):
+        kwargs['forms'] = self.get_forms()
+        return super().get(request, *kwargs, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        forms = self.get_forms()
+        posted_form = DNSRecordForm(request.POST)
+        for i, form in enumerate(forms):
```
mkurek, line 53: could you add some comment what is going on here?

src/ralph/dns/views.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+        )
+        for item in initial:
+            forms.append(DNSRecordForm(item))
+
+        forms.append(DNSRecordForm())
+        return forms
+
+    def get(self, request, *args, **kwargs):
+        kwargs['forms'] = self.get_forms()
```
mkurek: shouldn't we check if forms are already in kwargs? What about the case with errors during post? Do we need to call this API second time?

src/ralph/dns/templates/dns/dns_edit.html
```diff
@@ -0,0 +1,49 @@
+{% extends BASE_TEMPLATE %}
+{% load i18n %}
+
+{% block view_content %}
+  <div class="row">
+    <div class="large-12 columns">
+      <h1>DNS edit</h1>
```
mkurek: translate

src/ralph/dns/forms.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    content = forms.CharField(
+        label=_('Content'),
+        max_length=255,
+        help_text=_(
+            'The \'right hand side\' of a DNS record. For an A'
+            ' record, this is the IP address'
+        ),
+    )
+    ptr = forms.BooleanField(
```
mkurek, line 43: shouldn't we validate if PTR is checked only for 'A' records?

src/ralph/dns/dnsaas.py
```diff
@@ -0,0 +1,172 @@
...
+                record['type'] == str(RecordType.a.id)
+                else settings.DNSAAS_AUTO_PTR_NEVER
+            ),
+            'owner': settings.DNSAAS_OWNER
+        }
+        request = self.session.patch(url, data=data)
+        if request.status_code != 200:
+            return request.json()
+
+    @cache()
```
mkurek: i'm not sure if this cache will work in class (think about first param)

src/ralph/dns/forms.py
ghost, line 43: done

src/ralph/dns/dnsaas.py
```diff
@@ -0,0 +1,171 @@
...
+        url = urljoin(
+            settings.DNSAAS_URL, 'api/domains/?'.format(
+                urlencode([('name', domain_name)])
+            )
+        )
+        result = self.get_api_result(url)
+        if result:
+            return result[0]['url']
+
+    def create_dns_records(self, record):
```
vi4m: create_dns_records or create_dns_record ?

src/ralph/dns/dnsaas.py
mkurek: imo we should implement sth similar to `skip_first` here: https://github.com/ambv/kitpy/blob/master/src/lck/cache/memoization.py#L47

src/ralph/dns/views.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class DNSView(RalphDetailView):
+    icon = 'chain-broken'
+    name = 'dns_edit'
+    label = 'DNS'
+    url_name = 'dns_edit'
+    template_name = 'dns/dns_edit.html'
+
+    def __init__(self, *args, **kwargs):
```
mkurek: could you check here if `settings.ENABLE_DNSAAS_INTEGRATION` is set to `True`?

mkurek:
:ok:

ar4s:
:ok:

Issue #2298, opened 2016-03-16
LDAP Sync not working


mholttech:
I am trying to set up LDAP sync with our 2008 AD server and am running into some issues.

Code inserted into src/ralph/settings/prod.py:

```
AUTHENTICATION_BACKENDS = (
    'django_auth_ldap.backend.LDAPBackend',
    'django.contrib.auth.backends.ModelBackend',
)

import ldap
from django_auth_ldap.config import LDAPSearch, GroupOfNamesType
AUTH_LDAP_SERVER_URI = "ldap://ladbsix.somenet.local:389"
AUTH_LDAP_BIND_DN = "CN=ralph,OU=Ralph,DC=somenet,DC=local"
AUTH_LDAP_BIND_PASSWORD = "somepass"
AUTH_LDAP_PROTOCOL_VERSION = 3
AUTH_LDAP_USER_USERNAME_ATTR = "sAMAccountName"
AUTH_LDAP_USER_SEARCH_BASE = "DC=allegrogroup,DC=internal"
AUTH_LDAP_USER_SEARCH_FILTER = '(&(objectClass=*)({0}=%(user)s))'.format(
  AUTH_LDAP_USER_USERNAME_ATTR)
AUTH_LDAP_USER_SEARCH = LDAPSearch(AUTH_LDAP_USER_SEARCH_BASE,
  ldap.SCOPE_SUBTREE, AUTH_LDAP_USER_SEARCH_FILTER)
AUTH_LDAP_USER_ATTR_MAP = {
  "first_name": "givenName",
  "last_name": "sn",
  "email": "mail",
}

from ralph.accounts.ldap import MappedGroupOfNamesType

AUTH_LDAP_GROUP_MAPPING = {
  'CN=_gr_ralph_admin,OU=Ralph,DC=somenet,DC=local': "staff",
  'CN=_gr_ralph_admin,OU=Ralph,DC=somenet,DC=local': "superuser",
  'CN=_gr_ralph_admin,OU=Ralph,DC=somenet,DC=local': "ladb",
}
AUTH_LDAP_MIRROR_GROUPS = True
AUTH_LDAP_GROUP_TYPE = MappedGroupOfNamesType(name_attr="cn")
AUTH_LDAP_GROUP_SEARCH = LDAPSearch("DC=somenet,DC=local",
    ldap.SCOPE_SUBTREE, '(objectClass=group)')

AUTH_LDAP_NESTED_GROUPS = {
[...]

PR #2299, opened 2016-03-17
Fixed TypeError during running ``ralph ldap_sync``

ar4s:
Related to #2298


src/ralph/accounts/management/commands/ldap_sync.py
```diff
@@ -66,7 +66,7 @@ def get_nested_groups():
     users_groups = defaultdict(set)
     nested_groups = getattr(settings, 'AUTH_LDAP_NESTED_GROUPS', None)
     if not nested_groups:
-        return
+        return None, None
```
mkurek: maybe it'll be better to return here `return group_users, users_groups`? Then it'll have types of the valid result

mkurek:
:ok:

ghost:
:ok:

szok:
:ok:

PR #2302, opened 2016-03-17
Create DNS records bugfix

szok:
- Change lru_cache to ralph cache
- Correct display PTR checkbox


ghost:
:ok:

src/ralph/dns/dnsaas.py
```diff
@@ -98,7 +98,7 @@ def update_dns_records(self, record):
             return False
         return True
 
-    @lru_cache()
+    @cache()
```
mkurek, line 101: this cache here is useless since almost all record will be different. The only cache which makes sense here is caching based on `domain_name` param.

PR #2305, opened 2016-03-17
BaseObject API improvements

mkurek:
- added serializer for each descendant model of BaseObject
- added `__str__` field to BaseObject serializer - this field provides simple way to get string representation of BaseObject (not matter what descendant type is)
- added tests to check if every descendant model of BaseObject has `__str__` field attached

### DIFF
--- a/src/ralph/api/utils.py
+++ b/src/ralph/api/utils.py
@@ -1,7 +1,11 @@
+import logging
+
 from django.db import models
 from rest_framework import serializers
 
 from ralph.admin.sites import ralph_site
+
+logger = logging.getLogger('__name__')
 
 
 class QuerysetRelatedMixin(object):
@@ -69,7 +73,9 @@ class PolymorphicViewSetMixin(QuerysetRelatedMixin):
                     args[0].__class__
                 ]
             except KeyError:
-                pass
+                logger.warning('Dedicated serializer not found for {}'.format(
+                    args[0].__class__
+                ))
         return serializer_class(*args, **kwargs)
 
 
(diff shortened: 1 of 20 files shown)


src/ralph/api/utils.py
```diff
@@ -69,7 +73,9 @@ def get_serializer(self, *args, **kwargs):
                     args[0].__class__
                 ]
             except KeyError:
-                pass
```
ghost, line 72: On `KeyError` this `serializer_class = self.get_serializer_class()` will be used.
Is this ok?

src/ralph/assets/tests/test_api.py
```diff
@@ -489,3 +531,21 @@ def test_tags(self):
         )
         response = self.client.get(url, format='json')
         self.assertEqual(len(response.data['results']), 1)
+
+    def test_str_field(self):
+        count = 0
+        for descendant in BaseObject._polymorphic_descendants:
+            if not descendant._polymorphic_descendants:
```
ghost, line 538: Why this check?

mkurek, line 538: There is `Asset` model, which inherits from `BaseObject`, but is not "final" type - `DataCenterAsset` and `BackOfficeAsset` are inheriting from it - so we're not exposing `assets` API endpoint and don't want to check it

ghost, line 538: :+1:

ghost:
:ok:

src/ralph/api/utils.py
mkurek, line 72: imo yes. If something has no serializer we don't know how to serialize it (on dedicated view), so only base object serializer (or any other polymorphic in the future) will be used and only fields of base object will be returned.

ghost, line 72: :ok:

szok:
:ok:

ar4s:
:ok:

PR #2315, opened 2016-03-22
Network trees

szok:
TODO
- [x] breadcrumps
- [x] sql performance

zrzut ekranu 2016-03-23 o 09 56 11

### DIFF
--- a/src/ralph/networks/admin.py
+++ b/src/ralph/networks/admin.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlencode
+
 from django import forms
 from django.conf import settings
 from django.core.urlresolvers import reverse
@@ -6,7 +8,8 @@ from django.forms.models import ModelForm
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.admin import RalphAdmin, register
-from ralph.admin.mixins import RalphAdminFormMixin, RalphMPTTAdmin
+from ralph.admin.filters import RelatedAutocompleteFieldListFilter
+from ralph.admin.mixins import RalphAdminFormMixin
 from ralph.assets.models import BaseObject
 from ralph.data_importer import resources
 from ralph.lib.mixins.admin import ParentChangeMixin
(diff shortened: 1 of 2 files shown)


src/ralph/networks/admin.py
```diff
@@ -45,11 +48,42 @@ class Meta:
 
 
 @register(Network)
-class NetworkAdmin(RalphMPTTAdmin):
+class NetworkAdmin(RalphAdmin):
+
+    def subnetwork(self, obj):
```
mkurek: please move these fields under attributes definition

src/ralph/networks/admin.py
```diff
@@ -45,11 +48,42 @@ class Meta:
...
+                urlencode({'parent': obj.pk}),
+                _('Show'),
+                count
+            )
+        return ''
+
+    subnetwork.short_description = 'Subnetwork'
+    subnetwork.allow_tags = True
+
+    def name(self, obj):
```
mkurek: do we really need this field declaration? why default-way is not working (just declaring `'name'` on `list_display`)?

szok: to delete this.

src/ralph/networks/templates/networks/network_change_list.html
```diff
@@ -0,0 +1,13 @@
+{% extends 'admin/change_list.html' %}
+{% block breadcrumbs %}
+<ul class="breadcrumbs">
+  <li><a href="{% url 'admin:networks_network_changelist' %}">Network</a></li>
```
mkurek, line 4: this is not the original behaviour of https://github.com/allegro/ralph/blob/ng/src/ralph/admin/templates/admin/breadcrumbs_foundation.html - couldn't you wrap the for loop here in some block and overwrite it with calling super and adding parent networks after default content?

src/ralph/networks/admin.py
```diff
@@ -90,6 +124,18 @@ def changeform_view(
...
         )
 
+    def changelist_view(self, request, extra_context=None):
+        if extra_context is None:
+            extra_context = {}
+        parent = request.GET.get('parent', None)
+        if parent:
+            parent = Network.objects.get(pk=parent)
+            extra_context.update(
+                {'parents_network': parent.get_ancestors(include_self=True)}
```
mkurek, line 109: are there ancestors returned in specified order?

mkurek:
:ok:

ghost:
:ok:

PR #2317, opened 2016-03-23
Network improvements

szok:
- Remove 'This is a' from IP Address admin
- Add sorting by IP Address in IP address page
- Change display parent network with new line
- Add url_field config to TableWithUrl
- Add linked_object to Addresses table
- Filter by ip
- Add new filters: IP Range Filter and Network Range Filter


src/ralph/networks/models/networks.py
```diff
@@ -365,16 +371,16 @@ class IPAddress(LastSeenMixin, TimeStampMixin, NetworkMixin, models.Model):
         default=None,
     )
     is_management = models.BooleanField(
-        verbose_name=_('This is a management address'),
+        verbose_name=_('Management address'),
```
mkurek: Maybe `is management address?`? Btw you'll need migrations for that

src/ralph/networks/filters.py
```diff
@@ -0,0 +1,37 @@
...
+
+class IPRangeFilter(TextListFilter):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.title = _('IP range')
+
+    def queryset(self, request, queryset):
+        if self.value():
+            network = ipaddress.ip_network(self.value())
```
mkurek: shouldn't we validate the value? or at least catch some `ValueError` here?

src/ralph/networks/filters.py
```diff
@@ -0,0 +1,37 @@
...
+        super().__init__(*args, **kwargs)
+        self.title = _('Network range')
+
+    def queryset(self, request, queryset):
+        if self.value():
+            network = ipaddress.ip_network(self.value())
+            min_ip = int(network.network_address)
+            max_ip = int(network.broadcast_address)
+
+            queryset = queryset.filter(min_ip__gte=min_ip, max_ip__lte=max_ip)
```
mkurek, line 46: how will this work with this whole 'parent' filter? IMO when I filter by this, I want to see all networks which falls into this range, not only top-level ones. 

Maybe parent filter should be replaced by that + another (boolean) filter "show only top level networks"?

src/ralph/networks/admin.py
```diff
@@ -168,26 +194,32 @@ def show_subnetworks(self, network):
...
+            url_field='address'
         ).render()
     show_addresses.allow_tags = True
     show_addresses.short_description = _('Addresses')
 
     def get_queryset(self, request):
         queryset = super().get_queryset(request)
-        parent = request.GET.get('parent', None)
-        queryset = queryset.filter(parent=parent)
+        parent = request.GET.get('parent', 0)
```
mkurek: maybe `parent = request.GET.get('parent') or None` ?

src/ralph/admin/views/main.py
```diff
@@ -77,4 +80,17 @@ def get_queryset(self, request):
...
+            range_filter = [
+                i for i in filters if isinstance(i, NetworkRangeFilter)
+            ]
+            if range_filter:
+                filter_params = self.get_filters_params()
+                if range_filter[0].lookup_kwarg in filter_params:
+                    self.model_admin.mptt_indent_field = 10
+                else:
+                    self.model_admin.mptt_indent_field = 0
+        self.list_per_page = queryset.count()
```
ar4s: to delete

src/ralph/admin/views/main.py
```diff
@@ -77,4 +80,17 @@ def get_queryset(self, request):
                 ):
                     queryset = queryset.distinct()
                 queryset = queryset & autocomplete_queryset
+
+        if issubclass(self.model_admin.__class__, RalphMPTTAdmin):
```
mkurek: this (admin) is definitely not a place for such a network-specific code

src/ralph/networks/admin.py
```diff
@@ -47,17 +47,53 @@ class Meta:
...
+
+    def linked_object(self, item):
+        return ip_address_base_object_link(item)
+
+    linked_object.title = _('Linked object')
+
+
+class NetworkRalphChangeList(RalphChangeList):
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request)
```
mkurek, line 83: missing comment why are we doing this

src/ralph/networks/admin.py
```diff
@@ -47,17 +47,64 @@ class Meta:
...
+        filters, *rest = self.get_filters(request)
+        range_filter = [
+            i for i in filters if isinstance(i, NetworkRangeFilter)
+        ]
+        if range_filter:
+            filter_params = self.get_filters_params()
+            if range_filter[0].lookup_kwarg in filter_params:
+                self.model_admin.mptt_indent_field = 10
+            else:
+                self.model_admin.mptt_indent_field = 0
```
mkurek: IMO it will not work in following case:
- user 1 visits networks admin (see mptt tree)
- user 1 applies range filter (see flat list)
- user 2 visits networks admin (see flat list, since mptt_indent_field is still 0 on model admin)

szok: This works properly.

src/ralph/networks/admin.py
```diff
@@ -168,26 +189,32 @@ def show_subnetworks(self, network):
...
-    def get_queryset(self, request):
-        queryset = super().get_queryset(request)
-        parent = request.GET.get('parent', None)
-        queryset = queryset.filter(parent=parent)
-        return queryset
+    def get_paginator(
+        self, request, queryset, per_page, orphans=0,
+        allow_empty_first_page=True
+    ):
+        per_page = queryset.count()
```
mkurek, line 207: some comment will be usefull here

src/ralph/networks/admin.py
```diff
@@ -47,17 +47,59 @@ class Meta:
...
+
+        Args:
+            request: Django Request object
+
+        Returns:
+            Django queryset
+        """
+        queryset = super().get_queryset(request)
+        filter_params = self.get_filters_params()
+        if filter_params:
```
mkurek, line 85: although this code is pretty ugly, we don't have any idea how to handle it better right now. We'll back to this in the next week.

mkurek:
@szok add this one comment and it's :ok: for me

ar4s:
:ok:

ghost:
:ok:

PR #2322, opened 2016-03-24
Asynchronous transitions - part 1

mkurek:
- added possibility to run transition on rq worker
- added views with preview of job status

Notice that async method (method executed on worker) requires DB access to update Job status and progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,5 @@ src/ralph/var/
 debian/ralph-core*
 
 .idea
+
+!src/ralph/lib
--- a/src/ralph/lib/external_services/base.py
+++ b/src/ralph/lib/external_services/base.py
@@ -1,15 +1,7 @@
 import time
 
+import django_rq
 from django.conf import settings
-from redis import Redis
-from rq import Queue
-
-from .conf import get_redis_connection_params
-
-redis_conn = Redis(**get_redis_connection_params())
-
-
-RALPH_EXTERNAL_SERVICES = getattr(settings, 'RALPH_EXTERNAL_SERVICES', {})
 
 
 class QueuedServiceError(Exception):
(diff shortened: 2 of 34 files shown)


src/ralph/lib/transitions/async.py
```diff
@@ -0,0 +1,132 @@
...
+            with transaction.atomic():
+                func(instances=[obj], **defaults)
+        except RescheduleAsyncTransitionActionLater as e:
+            # action is not ready - reschedule this job later and continue
+            # when you left off
+            transition_job.reschedule()
+            return
+        except Exception as e:
+            logger.exception(e)
+            tja.status = TransitionJobActionStatus.FAILED
```
ar4s, line 124: if action failed then job fail

src/ralph/lib/transitions/async.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    started_actions = 0
+    for action in executed_actions:
+        if action.status == TransitionJobActionStatus.FAILED:
+            logger.warning(
+                'Action {} has failed. Marked job {} as failed too'.format(
+                    action, job
+                )
+            )
+            job.status = JobStatus.FAILED
```
szok: job.fail() ?

src/ralph/lib/transitions/async.py
```diff
@@ -0,0 +1,132 @@
...
+    try:
+        _check_previous_actions(transition_job, executed_actions)
+    except AsyncTransitionError:
+        return
+
+    completed_actions_names = set([
+        tja.action_name for tja in executed_actions
+        if tja.status != TransitionJobActionStatus.STARTED
+    ])
+    for action in _order_actions_by_requirements(transition.actions.all(), obj):
```
mkurek: TODO: make sure that first executed action if the one with STARTED status (if there is already exactly one with this status)

src/ralph/lib/external_services/base.py
```diff
@@ -50,3 +42,14 @@ def run(self, **kwargs):
         while job and not any([job.is_finished, job.is_failed]):
             time.sleep(0.1)
         return job.result
+
+    def run_async(self, **kwargs):
+        job = self.queue.enqueue(self.method, kwargs=kwargs)
+        return job
+
+
+class InternalService(ExternalService):
```
mkurek, line 51: we should discuss if this is proper place for it (and if it should stay named `InternalService`)

.gitignore
```diff
@@ -75,3 +75,5 @@ src/ralph/var/
 debian/ralph-core*
 
 .idea
+
+!src/ralph/lib
```
szok, line 79: :+1:

PR #2326, opened 2016-03-25
Added --exclude-networks switch.

ghost:
This switch allows to exclude network from exported networks.
In addition networking resources was updated.

### DIFF
--- a/src/ralph/export_to_ng/management/commands/export_model.py
+++ b/src/ralph/export_to_ng/management/commands/export_model.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import os
+import sys
 import tempfile
 from optparse import make_option
 from zipfile import ZipFile
@@ -14,6 +15,7 @@ from zipfile import ZipFile
 from django.core.management.base import BaseCommand
 from django.db.models import get_models
 from import_export import resources
+from ipaddr import AddressValueError, IPv4Network
 
 import ralph_assets
 from ralph.discovery import models_device
@@ -33,6 +35,7 @@ SIMPLE_MODELS = [
     # mainly trash data: 'ComponentModel',
     'Warehouse',
     'DataCenter',
+    'DiscoveryDataCenter',
     'Accessory',
     # TODO: 'CloudProject',
     'LicenceType',
@@ -43,6 +46,7 @@ SIMPLE_MODELS = [
     # TODO: VIP
     # TODO: VirtualServer
     'NetworkKind',
+    'NetworkEnvironment',
     'SupportType',
     'Region',
 ]
@@ -125,7 +129,31 @@ class Command(BaseCommand):
             action="append",
             type="str"
         ),
+        make_option(
+            '--exclude-network',
+            action="append",
+            help='Excludes network from address, eg.: --exclude-network=10.20.30.00/24',  # noqa
+            type="str"
+        ),
     )
+
+    def set_options_on_resource(self, model, options, model_resource):
+        """Injects command line options to resources"""
+        if model == 'Network' and options.get('exclude_network', None):
+            excluded_networks = []
+            for network in options['exclude_network']:
+                try:
+                    IPv4Network(network)
+                except AddressValueError:
+                    self.stdout.write(
+                        "Value '{}' for '--exclude-network' is not CIDR format".format(  # noqa
+                            network
+                        )
+                    )
+                    sys.exit()
+                excluded_networks.append(network)
+            model_resource.excluded_networks = excluded_networks
+        return model_resource
 
     def handle(self, *args, **options):
         models = args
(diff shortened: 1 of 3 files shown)


src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -125,6 +141,11 @@ class Command(BaseCommand):
             action="append",
             type="str"
         ),
+        make_option(
+            '--exclude-networks',
+            help='Excludes networks from passed range, eg.: --exclude-networks=<min-ip>,<max-ip>', # noqa
```
mkurek: min and max ips are "ok", but IMO it'd be better to pass network with mask here (ex. 10.20.30.00/24).
Btw what if I want to exclude many networks? Consider `action="append"` here

src/ralph/export_to_ng/resources.py
```diff
@@ -430,6 +430,12 @@ def get_queryset(self):
 class NetworkResource(resources.ModelResource):
     kind = fields.Field('kind', column_name='kind')
 
+    def get_queryset(self):
+        queryset = models_network.Network.objects.exclude(
+            min_ip__gte=self.min_ip, max_ip__lte=self.max_ip,
```
mkurek: what if `min_ip` and `max_ip` are not defined?

src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -109,6 +110,21 @@ def get_resource(model_name):
     return resource()
 
 
+def set_options_on_resource(model, options, model_resource):
+    """Injects command line options to resources"""
+    if model == 'Network' and 'exclude_networks' in options:
+        try:
+            min_ip, max_ip = options['exclude_networks'].split(',')
+        except ValueError:
+            print(
```
vi4m: Print errors not on stdout, but stderr.

ghost:
TODO: test if data are importable in NG

szok:
:ok:

src/ralph/export_to_ng/management/commands/export_model.py
```diff
@@ -125,8 +131,33 @@ class Command(BaseCommand):
...
+                    IPv4Network(network)
+                except AddressValueError:
+                    self.stdout.write(
+                        "Value '{}' for '--exclude-network' is not CIDR format".format( # noqa
+                            network
+                        )
+                    )
+                    sys.exit()
+                excluded_networks.append(network)
+        model_resource.excluded_networks = excluded_networks
```
mkurek: shouldn't it be set only on Network model?

ghost: of course :) thanks

src/ralph/export_to_ng/resources.py
```diff
@@ -429,6 +430,16 @@ def get_queryset(self):
 
 class NetworkResource(resources.ModelResource):
```
mkurek, line 477: I'd add here `excluded_networks = []` property if somehow this resource is called outside of mgmt command

src/ralph/export_to_ng/resources.py
```diff
@@ -30,11 +30,29 @@
...
 logger = logging.getLogger(__name__)
 
 
+def get_data_center_id(network_datacenter):
+    """
+    Gets data center id for `network_datacenter`.
+    """
+    try:
+        data_center_id = DataCenter.objects.get(
+            name__iexact=network_datacenter.name
```
mkurek, line 46: couldn't you use `deprecated_ralph_dc` field (https://github.com/allegro/ralph_assets/blob/develop/src/ralph_assets/models_dc_assets.py#L128)?

src/ralph/export_to_ng/resources.py
```diff
@@ -427,8 +445,61 @@ def get_queryset(self):
         return list(chain(parents, children))
 
 
+class NetworkKindResource(resources.ModelResource):
+    class Meta:
+        fields = ('id', 'name',)
+        model = models_network.NetworkKind
+
+
+class DiscoveryDataCenterResource(resources.ModelResource):
```
mkurek, line 454: shouldn't we skip here DCs which have equivalent in assets? comment is suggesting that, but i don't see any code that is doing that

src/ralph/export_to_ng/resources.py
ghost, line 46: It ended up, it points to different datacenter (ralph core device with type='datacenter') which means. It's impossible.

src/ralph/export_to_ng/resources.py
xliiv: I think this docsting is obsolete.
Something more like this
```
This exports only data centers from:
models_network.DataCenter
```
is correct but in such case there is no point to put it at all, imho.

ghost, line 454: Done

szok:
:ok:

mkurek:
:ok:

PR #2330, opened 2016-03-30
Fixed missing iron-a11y-announce.

ghost:
Changes:
- fixed elements' versions
- all static files copied in gulp task 'js' (in case another element is added)


mkurek:
:ok:

ar4s:
:ok:

PR #2331, opened 2016-03-30
WIP Transition API

szok:


src/ralph/lib/transitions/api/routes.py
```diff
@@ -0,0 +1,19 @@
...
+    TransitionActionViewSet,
+    TransitionViewSet,
+    TransitionView
+)
+
+
+router.register(r'transitions', TransitionViewSet)
+router.register(r'transitions-action', TransitionActionViewSet)
+
+urlpatterns = [url(
```
mkurek: why not in `urls.py`? btw in other modules it's `routers`, not `routes`

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,106 @@
...
+)
+from ralph.lib.transitions.models import (
+    Action,
+    Transition,
+    run_field_transition
+)
+from ralph.lib.transitions.views import collect_actions
+
+
+FIELD_MAP = {
```
mkurek, line 21: what about `ChoiceField` and `ModelChoiceField` (see `BackOfficeAsset` actions)?

src/ralph/lib/transitions/api/routes.py
szok: because it is part of the API, and urlpatterns is loaded from a file routers.

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,106 @@
...
+                field_attr = field[1]
+                attrs = {
+                    name: getattr(
+                        options['field'], name, None
+                    ) for name in field_attr
+                }
+                fields[name] = field_class(**attrs)
+        return fields
+
+    def get_serializer(self, only_class=False):
```
ghost:
```
def get_serializer_class(self):                                               
    class_name = 'TransitionSerializer{}'.format(                             
        self.obj.__class__.__name__                                           
    )                                                                         
    class_attrs = self.get_fields()                                           
    serializer_class = type(                                                  
        class_name, (serializers.Serializer,), class_attrs                    
    )                                                                                                                                                 
    return serializer_class                                                   

def get_serializer(self):                                                     
    return get_serializer_class()()
```

?

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,106 @@
...
+            self.obj, self.transition
+        )
+        return super().dispatch(*args, **kwargs)
+
+    def get_fields(self):
+        fields = {}
+        for action in self.actions:
+            action_fields = getattr(action, 'form_fields', {})
+            for name, options in action_fields.items():
+                field = FIELD_MAP.get(options['field'].__class__, None)
```
ghost: `field_class, field_attr = FIELD_MAP.get(options['field'].__class__, None)`?
Btw, None is not indexable :)

PR #2332, opened 2016-03-31
Added `skip_first` to cache and check if dnsaas integration is enabled in view

mkurek:
- Now you could use `skip_first` param in cache decorator - when set to True, first param (usually class instance a.k.a self) will not be used to calculate hash of params
- dnsview not checks if dnsaas integration is enabled

### DIFF
--- a/src/ralph/helpers.py
+++ b/src/ralph/helpers.py
@@ -48,16 +48,23 @@ def _cache_key_hash(func, *args, **kwargs):
     return pickle.dumps((func.__module__, func.__name__, args, kwargs))
 
 
-def cache(seconds=300, cache_name=DEFAULT_CACHE_ALIAS):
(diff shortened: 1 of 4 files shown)


szok:
:ok:

src/ralph/helpers.py
```diff
@@ -48,16 +48,23 @@ def _cache_key_hash(func, *args, **kwargs):
...
 
 
-def cache(seconds=300, cache_name=DEFAULT_CACHE_ALIAS):
+def cache(seconds=300, cache_name=DEFAULT_CACHE_ALIAS, skip_first=False):
     """
     Cache the result of a function call with particular parameters for specified
     number of seconds.
+
+    Args:
+        * skip_first - set to True if first argument should not be considered
```
xliiv: Is this a big change to make seperate fns, one for skipping self (the new one) and old one?

ghost:
:ok:

PR #2334, opened 2016-03-31
Transition API

szok:
Possibility to run transition through the API and tests for synchronous and asynchronous transition.

### DIFF
--- a/src/ralph/lib/external_services/models.py
+++ b/src/ralph/lib/external_services/models.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 import logging
 import uuid
+from datetime import date
 
+from dateutil.parser import parse
 from dj.choices import Choices
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from django.db.models.query import QuerySet
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.fields.json import JSONField
 
@@ -137,8 +140,21 @@ class Job(TimeStampMixin):
         Dump obj to JSON-acceptable format
         """
         result = obj
-        if isinstance(obj, (list, tuple)):
+        if isinstance(obj, (list, tuple, set)):
             result = [cls.dump_obj_to_jsonable(p) for p in obj]
+        elif isinstance(obj, QuerySet):
+            result = {
+                '__django_queryset': True,
+                'value': [i.pk for i in obj],
+                'content_type_id': ContentType.objects.get_for_model(
+                    obj.model
+                ).pk
+            }
+        elif isinstance(obj, date):
+            result = {
+                '__date': True,
+                'value': str(obj)
+            }
         elif isinstance(obj, dict):
             result = {}
             for k, v in obj.items():
@@ -161,7 +177,14 @@ class Job(TimeStampMixin):
         if isinstance(obj, (list, tuple)):
             result = [cls._restore_django_models(p) for p in obj]
         elif isinstance(obj, dict):
-            if obj.get('__django_model') is True:
+            if obj.get('__date') is True:
+                result = parse(obj.get('value')).date()
+            elif obj.get('__django_queryset') is True:
+                ct = ContentType.objects.get_for_id(obj['content_type_id'])
+                result = ct.model_class().objects.filter(
+                    pk__in=obj.get('value')
+                )
+            elif obj.get('__django_model') is True:
                 ct = ContentType.objects.get_for_id(obj['content_type_id'])
                 result = ct.get_object_for_this_type(pk=obj['object_pk'])
             else:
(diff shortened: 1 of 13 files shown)


src/ralph/lib/transitions/api/routers.py
```diff
@@ -0,0 +1,21 @@
...
+    TransitionViewSet
+)
+
+router.register(r'transitions', TransitionViewSet)
+router.register(r'transitions-action', TransitionActionViewSet)
+router.register(r'transitions-model', TransitionModelViewSet)
+
+
+urlpatterns = [url(
+    r'^transition/(?P<transition_id>[0-9]+)/(?P<obj_pk>[0-9]+)$',
```
mkurek: `obj_pk` is not necessary int (ex. uid)

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,121 @@
...
+        if serializer.is_valid():
+            transition_result = run_transition(
+                [self.obj],
+                self.transition,
+                self.transition.model.field_name,
+                serializer.data,
+                request=request
+            )
+            if self.transition.is_async:
+                result['job_ids'] = transition_result
```
mkurek: shouldn't we expose `TransitionJob` through API and put there some links to these resources?

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,135 @@
...
+
+class TransitionViewSet(RalphReadOnlyAPIViewSet):
+    queryset = Transition.objects.all()
+    serializer_class = TransitionSerializer
+    prefetch_related = ['actions']
+
+
+class TransitionView(APIView):
+
+    def dispatch(self, *args, **kwargs):
```
mkurek: why `obj_pk` and `transition_id` (btw - pk and id - can't we do this uniformly?) are not explicit params here?

szok: How to unify?
We need transition ID because need to know what transitions run and object (obj_pk) for this transition.

src/ralph/lib/transitions/async.py
```diff
@@ -128,6 +128,6 @@ def _perform_async_transition(transition_job):
         finally:
             tja.save()
         completed_actions_names.add(action.name)
-
+    obj.save()
```
mkurek: this is fixed in my PR: #2338

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,135 @@
...
+                self.transition.model.field_name,
+                serializer.validated_data,
+                request=request
+            )
+            if self.transition.is_async:
+                result['job_ids'] = [
+                    reverse('transitionjob-detail', args=(i,))
+                    for i in transition_result
+                ]
+                result['status'] = True
```
mkurek: true?

szok: yes because this is async and this view run task correct.

src/ralph/lib/external_services/models.py
```diff
@@ -137,8 +140,13 @@ def dump_obj_to_jsonable(cls, obj):
         Dump obj to JSON-acceptable format
         """
         result = obj
-        if isinstance(obj, (list, tuple)):
+        if isinstance(obj, (list, tuple, set, QuerySet)):
```
mkurek: dump queryset to dict with content type and list od ids and restore this queryset in `_restore_django_models`

src/ralph/lib/external_services/models.py
```diff
@@ -137,8 +140,21 @@ def dump_obj_to_jsonable(cls, obj):
         Dump obj to JSON-acceptable format
         """
         result = obj
-        if isinstance(obj, (list, tuple)):
+        if isinstance(obj, (list, tuple, set)):
             result = [cls.dump_obj_to_jsonable(p) for p in obj]
+        elif isinstance(obj, QuerySet):
+            result = {
+                '__django_queryset': True,
+                'value': ','.join([str(i.pk) for i in obj]),
```
mkurek: why not `[i.pk for i in obj]`?

src/ralph/lib/transitions/api/views.py
```diff
@@ -0,0 +1,136 @@
...
+            transition_result = run_transition(
+                [self.obj],
+                self.transition,
+                self.transition.model.field_name,
+                serializer.validated_data,
+                request=request
+            )
+            if self.transition.is_async:
+                result['job_ids'] = [
+                    reverse('transitionjob-detail', args=(i,))
```
mkurek: isn't it "gui" url? shouldn't we expose it through api as well?

mkurek:
:ok: - lets merge it and check it in action :)

ar4s:
:ok:

ghost:
:ok:

PR #2335, opened 2016-04-01
Added async fields to transition model

mkurek:
Allow to explicitly specify if transition should be executed asynchronously


szok:
:ok:

ar4s:
:ok:

PR #2338, opened 2016-04-05
Async transitions unification - part 1

mkurek:
- unification of synchronous and asynchoronous transitions
- fixed saving obj after async transition
- history entries are now properly created after async transition
- job field user renamed to username


ar4s:
:ok:

szok:
:ok:

PR #2339, opened 2016-04-05
Add position filter if is blade to localization column

szok:


src/ralph/data_center/admin.py
```diff
@@ -219,25 +220,47 @@ def get_multiadd_fields(self, obj=None):
...
+            position = generate_html_url(
+                base_url,
+                {
+                    'position__start': obj.position,
+                    'position__end': obj.position
+                },
+                '<strong>{}</strong>'.format(obj.position)
+            )
+        else:
+            position = '<strong>{}</strong>'.format(obj.position or '')
```
mkurek: you could put this before this if (and then use it as a label)

src/ralph/data_center/admin.py
```diff
@@ -219,25 +220,47 @@ def get_multiadd_fields(self, obj=None):
...
     def localization(self, obj):
+        """
+        Additional column 'localization' display filter by:
+        data center, server_room, rack, position (if is blade)
+        """
         base_url = reverse('admin:data_center_datacenterasset_changelist')
+        if obj.is_blade:
+            position = generate_html_url(
+                base_url,
+                {
```
mkurek, line 232: this filter will not work :)

ar4s:
:ok:

mkurek:
:ok:

PR #2352, opened 2016-04-08
Add cache to mptt network tree

szok:


src/ralph/admin/templates/admin/mptt_change_list.html
```diff
@@ -1,6 +1,12 @@
 {% extends "admin/change_list.html" %}
-{% load admin_list i18n mptt_admin %}
+{% load admin_list i18n mptt_admin cache %}
 
 {% block result_list %}
+  {% if cl.is_cacheable %}
+    {% cache 300 mptt_change_list_result request.get_full_path request.user.username %}
```
mkurek, line 6: what will happen if I add new network? :)

PR #2353, opened 2016-04-08
Fixed skip_row in ip addresse import.

ghost:
MAP_IMPORTED_ID_TO_NEW_ID got value before importer which is False.
When importer runs it changes MAP_IMPORTED_ID_TO_NEW_ID to True but this is
not reflected in resource.py since the value was loaded before
importer run.


mkurek:
:ok:

szok:
:ok:

PR #2354, opened 2016-04-08
Generate demodata bugfix

szok:


xliiv:
Don't we generate this data anymore?

ghost:
:ok:

mkurek:
Couldn't we generate demo data in a single test?

src/ralph/data_importer/management/commands/demodata.py
```diff
@@ -217,703 +221,699 @@ def add_user_and_group(username, password, group_name, permissions):
             user.save()
 
         dc_permissions = [
```
mkurek: do we have to list every possible permission? couldn't we just grant all in particular app or app.model?

mkurek:
:ok:

ghost:
:ok:

PR #2355, opened 2016-04-08
register transition

ghost:


src/ralph/lib/transitions/tests/test_models.py
```diff
@@ -220,3 +220,14 @@ def test_topological_sort_cycle(self):
         }
         with self.assertRaises(CycleError):
             [a for a in _sort_graph_topologically(graph)]
+
+
+
+class APITest(TransitionTestCase):
+    def test_xxx(self):
+        @transition_action(model=OrderStatus)
+        def example_action(*args, **kwargs):
```
mkurek: this action should take `cls` as a first argument IMO

szok:
:ok:

src/ralph/lib/transitions/tests/test_models.py
```diff
@@ -103,6 +103,31 @@ def test_run_action_during_transition(self):
...
         self.assertTrue(order.go_to_post_office.runned)
 
+    def test_action_is_added_to_model_when_registered_on_model(self):
+        @transition_action(model=OrderStatus)
+        def example_action(cls, *args, **kwargs):
+            pass
+        self.assertTrue(hasattr(OrderStatus, 'example_action'))
+
+    def test_transition_runs_action_registered_on_model(self):
+        @transition_action(model=Order)
```
mkurek: move this function to models.py

src/ralph/lib/transitions/decorators.py
```diff
@@ -22,6 +25,13 @@ def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
+
+        if 'model' in kwargs:
+            setattr(kwargs['model'], func.__name__, wrapper)
+            action, _ = Action.objects.get_or_create(name=func.__name__)
```
mkurek: this shouldn't be necessary if transition is registered before post_migrate signal

mkurek:
:ok:

PR #2276, opened 2016-03-09
Added DHCP config generated from Ralph

ar4s:

PR #2357, opened 2016-04-08
Added DHCP config generated from Ralph

ar4s:
Related to #2276


szok:
:ok:

vi4m:
When are we going to separate it into separate package(inside the same repo but still different directory) ?

src/ralph/assets/models/components.py
```diff
@@ -77,3 +84,25 @@ class GenericComponent(Component):
     class Meta:
         verbose_name = _('generic component')
         verbose_name_plural = _('generic components')
+
+
+class Ethernet(Component):
+    label = NullableCharField(
+        verbose_name=_('name'), max_length=255, blank=True, null=True
+    )
+    mac = models.CharField(
```
szok, line 93: Why not used: NullableCharFieldMixin ?

mkurek:
:ok:

Issue #2359, opened 2016-04-08
DC Visualization


MB0404:
As you see, the menu Layer is under the layer for the height-units and that grey bar for the back-side of a rack.
Tested in Chrome and ff. 

![ralph](https://cloud.githubusercontent.com/assets/8530436/14384843/9dbb0f10-fd9d-11e5-807a-49ea683a4082.JPG)

vi4m:
Thank you for bug report. Confirmed locally - looks like z-order issue in css.

PR #2360, opened 2016-04-08
Changed z-index of listing and pdu on DC View.

ar4s:
50 instead of 1000 (lower than dropdown z-index: 99).
Related to #2359.


mkurek:
:ok:

szok:
:ok:

PR #2361, opened 2016-04-08
Added invoice_no to DC asset admin

ar4s:
Commit 82dad2908a7c added localization and removed invoice_no.


ar4s:
- [ ] remove localization from bulk edit

mkurek:
:ok:

szok:
:ok:

PR #2364, opened 2016-04-11
localization -> location

vi4m:
Localization is used in translation sense. The author probably meant "location" word.


mkurek:
:ok:

szok:
:ok:

PR #2367, opened 2016-04-11
DNS: Change content field, add initial value to name

szok:
- Add Textinput widget to content field
- Add initial value to name field from first A record


mkurek:
:ok:

ar4s:
:ok:

PR #2368, opened 2016-04-12
Network improvements

szok:
- Add ordering in IPAddress and Network page
- Add ordering to network columns
- Add update network parents after save

### DIFF
--- a/src/ralph/networks/admin.py
+++ b/src/ralph/networks/admin.py
@@ -1,8 +1,8 @@
 from django import forms
 from django.conf import settings
-from django.contrib.admin.views.main import SEARCH_VAR
+from django.contrib.admin.views.main import ORDER_VAR, SEARCH_VAR
 from django.core.urlresolvers import reverse
-from django.db.models import Prefetch
+from django.db.models import Count, Prefetch
 from django.forms.models import ModelForm
 from django.utils.translation import ugettext_lazy as _
 
@@ -82,7 +82,11 @@ class NetworkRalphChangeList(RalphChangeList):
             Django queryset
         """
         queryset = super().get_queryset(request)
-        any_params = self.get_filters_params() or self.params.get(SEARCH_VAR)
+        any_params = (
+            self.get_filters_params() or
+            self.params.get(SEARCH_VAR) or
+            self.params.get(ORDER_VAR)
+        )
         if any_params:
             self.model_admin.mptt_indent_field = 10
         else:
@@ -92,10 +96,12 @@ class NetworkRalphChangeList(RalphChangeList):
 
 @register(Network)
 class NetworkAdmin(RalphMPTTAdmin):
+    ordering = ['min_ip', '-max_ip']
     change_form_template = 'admin/data_center/network/change_form.html'
     search_fields = ['name', 'address', 'remarks']
     list_display = [
-        'name', 'address', 'kind', 'vlan', 'network_environment'
+        'name', 'address', 'kind', 'vlan', 'network_environment',
+        'subnetworks_count', 'ipaddresses_count'
     ]
     list_filter = [
         'network_environment', 'kind', 'dhcp_broadcast', 'racks', 'terminators',
@@ -160,7 +166,17 @@ class NetworkAdmin(RalphMPTTAdmin):
     def address(self, obj):
         return obj.address
     address.short_description = _('Network address')
-    address.admin_order_field = ['min_ip']
+    address.admin_order_field = ['min_ip', '-max_ip']
+
+    def subnetworks_count(self, obj):
+        return obj.get_descendant_count()
+    subnetworks_count.short_description = _('Subnetworks count')
+    subnetworks_count.admin_order_field = 'subnetworks_count'
+
+    def ipaddresses_count(self, obj):
+        return obj.ipaddress_count
+    ipaddresses_count.short_description = _('IPAddress count')
+    ipaddresses_count.admin_order_field = 'ipaddress_count'
 
     def show_parent_networks(self, network):
         if not network or not network.pk:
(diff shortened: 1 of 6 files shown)


src/ralph/networks/models/networks.py
```diff
@@ -242,6 +242,16 @@ def save(self, *args, **kwargs):
...
         self._assign_ips_to_network()
+        self._update_subnetworks_parent()
+
+    def _update_subnetworks_parent(self):
+        parent = self.get_network()
+        networks = self.__class__.objects.filter(
+            parent=parent
+        ).exclude(pk=self.pk)
+        for network in networks:
+            network.parent = network.get_network()
```
mkurek: `save` already take care of this

src/ralph/networks/models/networks.py
```diff
@@ -242,6 +242,16 @@ def save(self, *args, **kwargs):
         self._assign_parent()
         super(Network, self).save(*args, **kwargs)
         self._assign_ips_to_network()
+        self._update_subnetworks_parent()
+
+    def _update_subnetworks_parent(self):
```
mkurek, line 275: i'd appreciate some tests for it

src/ralph/networks/admin.py
```diff
@@ -162,6 +168,16 @@ def address(self, obj):
     address.short_description = _('Network address')
     address.admin_order_field = ['min_ip']
```
mkurek: min and max

src/ralph/networks/admin.py
```diff
@@ -208,6 +224,12 @@ def show_addresses(self, network):
     show_addresses.allow_tags = True
     show_addresses.short_description = _('Addresses')
 
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        qs = qs.annotate(ipaddress_count=Count('ips'))
+        qs = qs.extra(select={'subnetwork_count': 'rght - lft'})
```
mkurek: add a comment that this is not exact number of subnetworks but it's fine estimate for ordering

mkurek: btw you could paste a link to the mptt here (https://github.com/django-mptt/django-mptt/blob/master/mptt/models.py#L594)

src/ralph/networks/admin.py
```diff
@@ -270,7 +293,7 @@ def get_queryset(self, request):
         )
 
     def ip_address(self, obj):
```
mkurek, line 297: do we still need this function? if you'll remove it, remember that sorting by ip should be sorting by number underneath

src/ralph/networks/tests/test_models.py
```diff
@@ -197,6 +197,25 @@ def test_sub_network_should_assign_automatically(self):
...
+
+        net3 = Network.objects.create(
+            name='net3', address='10.20.30.128/25'
+        )
+
+        net2 = Network.objects.get(pk=net2.pk)
+        self.assertEqual(net2.parent, net3)
+
+        self.assertEqual(net3.parent, net1)
+
```
mkurek, line 218: additional tests:
- add network `10.20.40.128/28` as net4
- change net3 address now to `10.20.40.128/25`
- now net4 should be child of net3, and net2 should be child of net1
- remove net3 - now net4 should be parentless

src/ralph/networks/admin.py
```diff
@@ -162,6 +168,16 @@ def address(self, obj):
     address.short_description = _('Network address')
     address.admin_order_field = ['min_ip']
 
+    def subnetwork_count(self, obj):
```
mkurek: subnetworks_count

src/ralph/networks/admin.py
szok, line 297: Yes, only for column ordering.

mkurek, line 297: i really don't like wrapping it in `<a ...>` - what about this (on model):

```
@property
def _address_sortable(self):
    return self.address
_address_sortable.admin_order_field = 'number'
```

src/ralph/networks/models/networks.py
```diff
@@ -82,9 +82,7 @@ class NetworkMixin(object):
     _parent_attr = None
 
     def _assign_parent(self):
-        parent = getattr(self, self._parent_attr)
-        if not parent or self.pk:
-            setattr(self, self._parent_attr, self.get_network())
+        setattr(self, self._parent_attr, self.get_network())
```
mkurek, line 88: why?

mkurek:
:ok:

src/ralph/networks/tests/test_models.py
```diff
@@ -197,6 +197,39 @@ def test_sub_network_should_assign_automatically(self):
...
+        net3 = Network.objects.create(
+            name='net3', address='10.20.30.128/25'
+        )
+
+        self.refresh_objects_from_db(net2)
+        self.assertEqual(net2.parent, net3)
+
+        self.assertEqual(net3.parent, net1)
+
+        net4 = Network.objects.create(
```
mkurek, line 219: please add here `net5` with address `10.20.40.0/24` and check if net3 after changing address will be descendant of net5

src/ralph/networks/tests/test_models.py
```diff
@@ -197,6 +197,39 @@ def test_sub_network_should_assign_automatically(self):
...
+        net3.address = '10.20.40.128/25'
+        net3.save()
+
+        self.refresh_objects_from_db(net1, net2, net3, net4)
+
+        self.assertEqual(net4.parent, net3)
+        self.assertEqual(net2.parent, net1)
+
+        net3.delete()
+        self.assertIsNone(Network.objects.filter(pk=net4.pk).first())
```
mkurek: is this desired behavior that deleting parent network will delete all subnetworks as well? imo `net4` should stay as it were before, but without parent now

ar4s:
:ok:

src/ralph/networks/tests/test_models.py
szok, line 219: Done

src/ralph/networks/models/networks.py
```diff
@@ -257,11 +255,24 @@ def save(self, *args, **kwargs):
...
         self.max_ip = int(self.broadcast_address)
         self._assign_parent()
         super(Network, self).save(*args, **kwargs)
         self._assign_ips_to_network()
+        if enable_save_descendants:
+            self._update_subnetworks_parent()
+
+    def delete(self):
+        with transaction.atomic():
+            self.address = '0.0.0.0/32'
```
mkurek, line 271: add some comment why this fake address is used

PR #2370, opened 2016-04-12
Added information about empty list and added asterisk for required fields

ar4s:
spectacle 123
spectacle 222

### DIFF
--- a/src/ralph/admin/templates/admin/change_list_results.html
+++ b/src/ralph/admin/templates/admin/change_list_results.html
@@ -1,4 +1,4 @@
-{% load i18n admin_static %}
+{% load i18n admin_static admin_urls %}
 
 {% if result_hidden_fields %}
   <div class="hiddenfields">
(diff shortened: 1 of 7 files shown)


src/ralph/admin/templates/admin/change_list_results.html
```diff
@@ -56,4 +56,11 @@
       </tbody>
     </table>
   </div>
+{% else %}
+  <div data-alert class="alert-box info">
+    {% url cl.opts|admin_urlname:'add' as add_url %}
+    {% blocktrans with cl.opts.verbose_name as name %}
+      Nothing to display. You can <a href="{{ add_url }}">add first {{ name }}</a>.
```
mkurek: I'm not sure if "first" is the best word here - what if empty list is result of filtering/searching?

szok:
:ok:

mkurek:
:ok:

PR #2373, opened 2016-04-14
Add tags to BO/DC Asset bulk edit

szok:


src/ralph/lib/mixins/models.py
```diff
@@ -88,8 +89,15 @@ def formfield(self, form_class=TaggitTagField, **kwargs):
         return super().formfield(form_class, **kwargs)
 
 
+class RalphTaggableManager(_TaggableManager):
```
mkurek: why is this necessary?

src/ralph/lib/mixins/fields.py
```diff
@@ -205,4 +208,8 @@ def has_changed(self, initial, data):
             data = [i.strip() for i in data.split(',') if i.strip()]
             changed = len(initial) != len(data) or set(initial) - set(data)
             return changed
+        elif not initial and data:
```
mkurek: you could simplify these conditions into one

src/ralph/lib/mixins/models.py
szok: Yes, because Django Model manager need have self.name, default TaggitManager has not this variable. 

https://github.com/alex/django-taggit/blob/develop/taggit/managers.py#L111
https://github.com/django/django/blob/1.8.12/django/db/models/manager.py#L71

If we not add self.name we get exception:

```
AttributeError at /back_office/backofficeasset/
'_TaggableManager' object has no attribute 'name'
```

src/ralph/lib/mixins/fields.py
```diff
@@ -205,4 +208,4 @@ def has_changed(self, initial, data):
             data = [i.strip() for i in data.split(',') if i.strip()]
             changed = len(initial) != len(data) or set(initial) - set(data)
             return changed
-        return False
+        return initial or data
```
mkurek, line 211: :+1:

mkurek:
:ok:

ar4s:
:ok:

PR #2384, opened 2016-04-18
Add Cluster model

szok:


src/ralph/data_center/models/virtual.py
```diff
@@ -20,3 +23,31 @@ class Meta:
 
     def __str__(self):
         return 'VIP: {}'.format(self.service_env)
+
+
+class ClusterType(NamedMixin):
```
ar4s: please add `models.Model` then model will be explicit

src/ralph/data_center/models/virtual.py
```diff
@@ -20,3 +23,31 @@ class Meta:
...
+
+    content_types = models.ManyToManyField(ContentType)
+
+
+class Cluster(NamedMixin):
```
ar4s: ditto

mkurek: `Cluster` should inherit from `BaseObject`

mkurek:
please expose it through API as well

src/ralph/data_center/api/serializers.py
```diff
@@ -15,6 +17,23 @@
...
 
 
+class ClusterTypeSerializer(RalphAPISerializer):
+    class Meta:
+        model = ClusterType
+        depth = 1
+
+
+class ClusterSerializer(BaseObjectSerializer):
+    base_objects = BaseObjectSerializer(
```
mkurek: we could use urls here instead of full serialized object

src/ralph/data_center/models/virtual.py
```diff
@@ -20,3 +23,36 @@ class Meta:
...
+
+
+class Cluster(BaseObject, NamedMixin, models.Model):
+
+    type = models.ForeignKey(ClusterType)
+    base_objects = models.ManyToManyField(
+        BaseObject,
+        verbose_name=_('Assigned base objects'),
+        through='BaseObjectCluster',
+        related_name='+',
```
mkurek: why not use defined name (ex. `clusters`)?

src/ralph/data_center/admin.py
```diff
@@ -54,6 +60,37 @@ class AccessoryAdmin(RalphAdmin):
...
+    )
+    raw_id_fields = ['parent', 'service_env']
+    raw_id_override_parent = {'parent': Cluster}
+    list_display = ['name', 'type']
+    list_select_related = ['type']
+    list_filter = ['name', 'type', 'parent', 'service_env']
+
+    class ClusterBaseObjectInline(RalphTabularInline):
+        model = BaseObjectCluster
+        fk_name = 'cluster'
```
mkurek, line 86: wouldn't django admin handle it for us?

ghost:
:ok:

src/ralph/data_center/admin.py
szok, line 86: Unfortunately Django is not handle this.

```
<class 'ralph.data_center.admin.ClusterAdmin.ClusterBaseObjectInline'>: (admin.E202) 'data_center.BaseObjectCluster' has more than one ForeignKey to 'data_center.Cluster'.
```

mkurek:
:ok:

PR #2388, opened 2016-04-19
Optional licence on assign_licence action (backoffice)

ghost:
Transition action: assign_licence (on backoffice asset) doesn't require
licence any more.


ar4s:
:ok:

szok:
:ok:

PR #2390, opened 2016-04-20
Fixed bulk-edit fill fields

mkurek:
f724344855ba8b0bae6022522a10a6c3bb828993 added bulk-edit css class to body, which breaks lookup of inputs for fill-fields (ex. search input was included to fill-fields). This fix narrows down possible inputs to be searched only in div with result list.


ghost:
:ok:

ar4s:
:ok:

PR #2391, opened 2016-04-20
[WIP] Timezone support

mkurek:
Successor of #2263 (i needed to add one more commit/fix)

We are using the following settings regarding timezone:

TIME_ZONE = 'Europe/Warsaw'
USE_TZ = False
When USE_TZ is False, Django would store timestamps in the timezone of TIME_ZONE, and this is how ralph is storing time. After I realised I should change TIME_ZONE, even if I changed TIME_ZONE to 'Asia/Beijing', the stored time are still in the old timezone (GMT+1). How about it that we change USE_TZ to True, to make sue time is always saved as UTC? This is also recommended by Django.

### DIFF
--- a/src/ralph/security/tests/factories.py
+++ b/src/ralph/security/tests/factories.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import factory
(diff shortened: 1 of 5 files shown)


romcheg:
I think we should actually revive this.

romcheg:
I have realized it's not possible for me to push commits to the same pull request. I will create a new one

PR #2392, opened 2016-04-20
Add reversion history to API

szok:
Added generate reversion history to API action

### DIFF
--- a/src/ralph/api/serializers.py
+++ b/src/ralph/api/serializers.py
@@ -3,7 +3,9 @@ import logging
 import operator
 from functools import reduce
 
+import reversion
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import transaction
 from django.db.models import Q
 from django.db.models.fields import exceptions
 from rest_framework import permissions, relations, serializers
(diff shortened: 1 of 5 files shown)


ghost:
:ok:

src/ralph/api/serializers.py
```diff
@@ -188,6 +190,13 @@ def build_field(self, *args, **kwargs):
             field_kwargs.setdefault('style', {})['base_template'] = 'input.html'
         return field_class, field_kwargs
 
+    def save(self):
```
mkurek: please make sure that every class that is used in `save_serializer_class` is inheriting from `RalphAPISaveSerializer`

szok: Yes, because all viewset inherit from RalphAPIViewSetMixin (RalphAPIViewSet)

mkurek: No, ex. `BaseObjectLicenceViewSet` and `BaseObjectLicenceSerializer` used in `save_serializer_class` - i assume that there is more cases like that.

szok: Done.

ar4s:
:ok:

PR #2402, opened 2016-04-27
force download files from /media/*

mkurek:
Added following header to nginx config when `/media` url is called:

```
add_header Content-disposition "attachment";
```

Not-forcing downloading of the media file could cause running dangerous script in client browser.

### DIFF
--- a/contrib/ralph.conf.nginx
+++ b/contrib/ralph.conf.nginx
@@ -14,6 +14,7 @@ server {
 
     location /media {
        alias /opt/media;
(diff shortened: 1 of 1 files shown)


ghost:
:ok:

ar4s:
:ok:

Chewie9999:
The only problem with this, is that it even forces the browser to download jpgs etc, even when it could display them natively. Is there any chance this could be optional, or allow a whitelist of permitted extensions, for example?

PR #2406, opened 2016-04-28
Improved Virtual Server model

mkurek:
Virtual model improvements (added hostname, sn, cluster). 

TODO (in another PR): components, IPAM, operations and other tabs from DCA


szok:
:ok:

ar4s:
:ok:

PR #2407, opened 2016-04-29
Configuration path (a.k.a venture)

mkurek:
- models for configuration path (a.k.a. venture/role in Ralph2)
- with API and importer (ex. from Ralph2)

### DIFF
--- a/docs/user/dcim.md
+++ b/docs/user/dcim.md
@@ -1,0 +1,24 @@
+# DCIM
+
+## Introduction
+
+TODO
+
+## Configuration path
+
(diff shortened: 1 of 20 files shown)


vi4m:
@andrzej-jankowski what about ansible ? could we reuse module, class in this case?

mkurek:
@vi4m i've described briefly in docs how this could be used for ansible but someone with better ansible-skill should look at it.

szok:
:ok:

ar4s:
:ok:

PR #2410, opened 2016-04-29
API performance improvements

mkurek:
- don't present (or even evaluate) choices for related field in OPTIONS request
- added fields to select_related and prefetch_related
- added some simple serializers for nested relations

### DIFF
--- a/src/ralph/lib/api/utils.py
+++ b/src/ralph/lib/api/utils.py
@@ -1,5 +1,31 @@
+import logging
+from collections import OrderedDict
+
+from django.core.urlresolvers import NoReverseMatch
+from django.utils.encoding import force_text
+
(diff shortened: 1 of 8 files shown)


src/ralph/lib/api/utils.py
```diff
@@ -1,5 +1,26 @@
...
+
+def get_list_view_name(model):
+    return '{}-list'.format(
+        model._meta.object_name.lower()
+    )
+
+
+def get_list_view_url_for_model(model):
+    view_name = get_list_view_name(model)
+    print(view_name)
```
szok: Remove print :)

szok:
:ok:

ar4s:
:ok:

PR #2415, opened 2016-05-09
Change column order in asset admin

ar4s:
Now, hostname is first column in DC and BO.


ghost:
:ok:

src/ralph/data_center/forms/data_center_asset.py
```diff
@@ -0,0 +1,14 @@
+from ralph.admin.mixins import RalphAdminForm
+from ralph.data_center.models.physical import DataCenterAsset
+
+
+class DataCenterAssetForm(RalphAdminForm):
+    def clean_hostname(self):
+        hostname = self.cleaned_data['hostname']
```
mkurek, line 8: what about `NullableCharField` for hostname?

PR #2417, opened 2016-05-10
Configuration path improvements

mkurek:
- removed confusing field "path" from configuration module
- configuration class "**str**" is now only last module + class name
- renamed "name" field in ConfigurationClass to "class_name"
- added filtering by configuration module and all its descendants
- added filtering by configuration_path in API

### DIFF
--- a/src/ralph/admin/filters.py
+++ b/src/ralph/admin/filters.py
@@ -414,6 +414,54 @@ class TreeRelatedFieldListFilter(RelatedFieldListFilter):
         return queryset
 
 
+class TreeRelatedAutocompleteFilterWithDescendants(
+    RelatedAutocompleteFieldListFilter
+):
+    """
+    Autocomplete filter for ForeignKeys to `mptt.models.MPTTModel` with
+    filtering by object and all its descendants.
+    """
+    def _get_descendants(self, request, root_id):
+        """
+        Return descendants of root object.
+        """
+        try:
+            root = self.field.rel.to.objects.get(pk=root_id)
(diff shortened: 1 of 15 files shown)


src/ralph/admin/filters.py
```diff
@@ -414,6 +414,54 @@ def queryset(self, request, queryset):
...
+class TreeRelatedAutocompleteFilterWithDescendants(
+    RelatedAutocompleteFieldListFilter
+):
+    """
+    Autocomplete filter for ForeignKeys to `mptt.models.MPTTModel` with
+    filtering by object and all its descendants.
+    """
+    def _get_descendants(self, request, root_id):
+        """
+        Return descendants of root object.
```
vi4m: Method descriptor already says that. Any more details? If not - remove.

src/ralph/admin/filters.py
```diff
@@ -414,6 +414,54 @@ def queryset(self, request, queryset):
...
+    def queryset(self, request, queryset):
+        value = self.value()
+        if not value:
+            ids = []
+        else:
+            ids = value.split(',')
+        q_param = Q()
+        for id_ in ids:
+            if id_ == self.empty_value:
+                q_param |= Q(**{'{}__isnull'.format(self.field_path): True})
```
ar4s, line 445: Change this to `+` notation or change below to `.format`

mkurek, line 445: done

ar4s:
:ok:

szok:
:ok:

PR #2419, opened 2016-05-11
Changed search box behaviour in DC View

ar4s:
- automatically redirect to first item after press enter in search box
- searching only by name of rack


szok:
:ok:

mkurek:
:ok: (what about jquery version)?

PR #2421, opened 2016-05-11
Cluster improvements

mkurek:
- added table with master information to cluster
- new cluster fields: status, hostname
- better handling of cluster base objects in API

### DIFF
--- a/src/ralph/api/tests/_base.py
+++ b/src/ralph/api/tests/_base.py
@@ -69,3 +69,12 @@ class RalphAPITestCase(APITestCase):
     def setUp(self):
         super().setUp()
         self.client.force_authenticate(self.superuser)
+
+    @classmethod
+    def get_full_url(self, url):
+        """
+        Convert URL to fully qualified URL (with protocol and domain).
+        """
+        # testserver is default name of server using by Django test client
+        # see django/test/client.py for details
+        return 'http://testserver{}'.format(url)
--- a/src/ralph/data_center/admin.py
+++ b/src/ralph/data_center/admin.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 from django.conf import settings
-from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.admin import RalphAdmin, RalphTabularInline, register
@@ -10,7 +9,6 @@ from ralph.admin.filters import (
     TagsListFilter,
     TreeRelatedAutocompleteFilterWithDescendants
 )
-from ralph.admin.helpers import generate_html_link
 from ralph.admin.m2m import RalphTabularM2MInline
 from ralph.admin.mixins import BulkEditChangeListMixin
 from ralph.admin.views.extra import RalphDetailViewAdmin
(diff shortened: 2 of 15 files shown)


src/ralph/api/tests/_base.py
```diff
@@ -69,3 +69,10 @@ def setUpClass(cls):
     def setUp(self):
         super().setUp()
         self.client.force_authenticate(self.superuser)
+
+    @classmethod
+    def get_full_url(self, url):
+        """
+        Convert URL to fully qualified URL (with protocol and domain).
```
mkurek: add note about https://github.com/django/django/blob/master/django/test/client.py#L283

src/ralph/data_center/admin.py
```diff
@@ -84,7 +99,35 @@ class ClusterBaseObjectInline(RalphTabularInline):
         extra = 1
         verbose_name = _('Base Object')
 
-    inlines = [ClusterBaseObjectInline]
+    inlines = [ClusterBaseObjectInline]  # TODO: add ClusterNetworkInline
```
szok: Add ClusterNetworkInline  to inlines :)

src/ralph/data_center/models/virtual.py
```diff
@@ -26,18 +30,73 @@ def __str__(self):
...
+
+class ClusterStatus(Choices):
+    _ = Choices.Choice
 
+    in_use = _('in use')
+    for_deploy = _('for deploy')
 
-class Cluster(BaseObject, NamedMixin, models.Model):
 
+class Cluster(BaseObject, models.Model):
```
szok, line 49: Why not use NamedMixin?

src/ralph/lib/table/table.py
```diff
@@ -24,10 +24,10 @@ class Table(object):
     Generating contents for table based on predefined columns and queryset.
 
     Example:
-        >>> table = Table(queryset, ['id', 'name'])
+        >>> table = Table(queryset, ['id', ('name', 'My field name')])
```
ghost, line 26: Could you mention about `transpose` option?
What's the meaning of tuple `('name', 'My field name')`?

src/ralph/data_center/models/virtual.py
mkurek, line 49: because in our case this field could be blank/null

src/ralph/lib/table/tests.py
```diff
@@ -13,17 +13,34 @@ def setUpClass(cls):
...
+        self.assertEqual(len(table1), len(table2))
+        for row1, row2 in zip(table1, table2):
+            self.assertSequenceEqual(row1, row2)
+
     def test_queryset_render(self):
         """Test get_perm_key function."""
         table = Table(Foo.objects.all(), ['id', 'bar'])
         result = [
-            ['ID', 'bar'],
+            [{'value': 'ID'}, {'value': 'bar'}],
```
ghost, line 28: Why ID here is capsed?

src/ralph/data_center/models/virtual.py
szok, line 49: ok :)

src/ralph/lib/table/table.py
xor-xor, line 26: +1 re: meaning of this tuple.

mkurek, line 26: i've added some description for **init** params. check it out :)

src/ralph/lib/table/tests.py
mkurek, line 28: this is verbose name of this field. https://github.com/django/django/blob/master/django/db/models/options.py#L254

src/ralph/lib/table/table.py
xor-xor, line 26: That description is fine, but I'd also add here a remark: "see **init**'s docstring for additional info" (or something like that).

mkurek, line 26: done

szok:
:ok:

ghost:
:ok:

PR #2429, opened 2016-05-17
Additional base object filters in API

mkurek:
- Filter by ip address using `ip` param
- Filter by service name or uid using `service` param


ghost:
:ok:

xor-xor:
:ok:

PR #2435, opened 2016-05-19
Fixed admin permissions for non-model fields

mkurek:
Before, only superuser users could access non-model fields (ex. property, admin method, model method) in admin change view. This commit fixes it - by default every such field is accessible by user who has view/change access to the model (this field is read-only either way). Additional permission could be set on this field using `_permission_field` property if necessary.

### DIFF
--- a/src/ralph/lib/permissions/models.py
+++ b/src/ralph/lib/permissions/models.py
@@ -241,7 +241,7 @@ class PermByFieldMixin(models.Model, metaclass=PermissionsBase):
         :return: List of field names
         :rtype: list
         """
-        result = []
+        result = set()
         blacklist = cls._permissions.blacklist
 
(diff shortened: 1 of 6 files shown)


ar4s:
:ok:

src/ralph/lib/permissions/models.py
```diff
@@ -241,17 +241,20 @@ def allowed_fields(cls, user, action='change'):
...
                 cls.has_access_to_field(field.name, user, action)
             ):
-                result.append(field.name)
-
-        return set(result)
+                result.add(field.name)
+        # If the user does not have rights to view,
+        # but has the right to change he can view the field
+        if action == 'view':
+            result |= cls.allowed_fields(user, 'change')
```
szok, line 256: 👍

ghost:
:ok:

szok:
:ok:

PR #2437, opened 2016-05-20
Deployment

mkurek:
Deployment for physical and virtual servers.

Features:
- configurable preboots (ipxe, kickstart, initrd, kernel)
- DHCP entries support
- deployment actions for transition

### DIFF
--- a/src/ralph/static/src/js/widgets.js
+++ b/src/ralph/static/src/js/widgets.js
@@ -1,0 +1,19 @@
+(function($){
+    $(function () {
+        "use strict";
+        $("div.select-other > select").change(function(){
+            var $this = $(this), val = $this.val(), name = $this.attr('name');
+            if(val === '__other__'){
+                $this.after(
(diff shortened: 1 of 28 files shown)


mkurek:
Preview:
ralph_deployment_form

src/ralph/static/src/js/widgets.js
```diff
@@ -0,0 +1,19 @@
+(function($){
+    $(function () {
+        "use strict";
+        $("div.select-other > select").change(function(){
+            var $this = $(this), val = $this.val(), name = $this.attr('name');
+            if(val === '__other__'){
+                $this.after(
+                    '<input class="vTextField" name="' + name + '__other__" type="text" autocomplete="off">'
```
ar4s, line 8: try hidden field and handle it in `ChoiceFieldWithOtherOption` then in JS you can toggle type of field (text, hidden)

PR #2439, opened 2016-05-24
Cloud host FQDN instead of name

mkurek:
This commmit changes the way cloud host (from OpenStack) is saved - before it was name of the instance from OpenStack. Now it's FQDN of this host (resolved by ip address of the host).


ar4s:
:ok:

szok:
:ok:

PR #2444, opened 2016-05-27
Deploy views

ar4s:


src/ralph/dhcp/models.py
```diff
@@ -1,32 +1,51 @@
...
+        )
+        deployments_mapper = {
+            obj.object_id: obj
+            for obj in Deployment.objects.filter(
+                id__in=deployment_ids
+            ).order_by('modified')
+        }
+        for obj in queryset:
+            obj.deployment = deployments_mapper.get(
+                str(obj.ethernet.base_object_id), False
```
mkurek: use None instead of False

src/ralph/dhcp/models.py
```diff
@@ -1,32 +1,51 @@
...
+        dca_content_type = ContentType.objects.get_for_model(DataCenterAsset)
+        deployment_ids = deployment_queryset.filter(
+            object_id__in=[str(i) for i in entries_ids],
+            content_type=dca_content_type
+        )
+        deployments_mapper = {
+            obj.object_id: obj
+            for obj in Deployment.objects.filter(
+                id__in=deployment_ids
+            ).order_by('modified')
```
mkurek: we don't need ordering here

src/ralph/deployment/forms.py
```diff
@@ -0,0 +1,16 @@
...
+
+from ralph.admin.mixins import RalphAdminFormMixin
+from ralph.deployment.models import PrebootItemType
+
+
+class PrebootConfigurationForm(RalphAdminFormMixin, forms.ModelForm):
+    def clean_configuration(self):
+        configuration_type = self.cleaned_data['type']
+        configuration = self.cleaned_data['configuration']
+        if configuration_type == PrebootItemType.kickstart.id:
```
mkurek, line 11: move it to model

src/ralph/deployment/models.py
```diff
@@ -1,33 +1,65 @@
...
 
-class PrebootFile(NamedMixin):
+def get_choices_from_group(choices_group):
+    return (
+        (choice.id, choice.name) for choice in choices_group.choices
+    )
+
+
+def preboot_file_name(instance, filename):
+    return os.sep.join(('pxe', instance.get_type_display(), instance.name))
```
mkurek: slugify the name

src/ralph/deployment/models.py
```diff
@@ -80,6 +159,28 @@ class Deployment(TransitionJob):
...
+    @classmethod
+    def get_deployment_for_ip(cls, ip):
+        base_object = Ethernet.objects.get(ipaddress__address=ip).base_object
+        deployment = None
+        try:
+            deployment = cls.objects.active().get(
+                content_type_id=base_object.content_type_id,
+                object_id=base_object.id
+            )
+        except cls.DoesNotExist:
```
mkurek: add TODO that there shoulb be validation for existing deployment for the same IP

src/ralph/deployment/views.py
```diff
@@ -0,0 +1,141 @@
...
+from django.http import Http404, HttpResponse, HttpResponseRedirect
+from django.template import Context, Template
+
+from ralph.admin.helpers import get_client_ip
+from ralph.deployment.models import Deployment
+
+logger = logging.getLogger(__name__)
+
+# TODO: find better solution
+RALPH_INSTANCE = getattr(settings, 'RALPH_INSTANCE', 'http://127.0.0.1:8000')
```
mkurek: just add default value to the settings and use later `settings.RALPH_INSTANCE`

src/ralph/deployment/views.py
```diff
@@ -0,0 +1,141 @@
...
+        raise Http404
+    return preboot
+
+
+def _render_configuration(configuration, deployment):
+    template = Template(configuration)
+    context = Context({
+        'ralph_instance': RALPH_INSTANCE,
+        'deployment_id': deployment.id,
+        'kickstart': RALPH_INSTANCE + reverse('deployment_kickstart', kwargs={
```
mkurek: couldn't we use `RALPH_INSTANCE` directly in template?

src/ralph/deployment/urls.py
```diff
@@ -0,0 +1,26 @@
+from django.conf.urls import url
+
+from ralph.deployment.views import done_ping, files, ipxe, kickstart
+
+urlpatterns = [
+    url(
+        r'^boot.ipxe$',
```
mkurek, line 7: could we expose pxe file for each deployment too if we later manage this to work in DHCP server? I mean `'^(?P<deployment_id>[-\w]+)/ipxe$`

src/ralph/deployment/views.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Raises:
+        Http404: if deployment with specified UUID doesn't exist
+    """
+    preboot = _get_preboot(deployment_id)
+    configuration = preboot.get_configuration('kickstart')
+    if configuration is None:
+        raise Http404
+    configuration = _render_configuration(
+        configuration, Deployment.objects.get(id=deployment_id)
```
mkurek: what if deployment does not exist? maybe use `get_object_or_404`?

src/ralph/deployment/views.py
```diff
@@ -0,0 +1,141 @@
+import logging
```
mkurek, line 1: some logging (info) will be useful in every view in this file

src/ralph/dhcp/models.py
```diff
@@ -1,32 +1,50 @@
...
+        return super().get_queryset().select_related(
+            'ethernet', 'ethernet__base_object'
+        ).exclude(
             hostname=None,
             ethernet__base_object=None,
             ethernet=None,
             status=IPAddressStatus.reserved.id
         )
 
+    # TODO: delete this
```
mkurek: so remove it :)

src/ralph/dhcp/templates/dhcp/entries.conf
```diff
@@ -1,4 +1,4 @@
 # DHCP config generated by Ralph last modified at {{ last_modified }}
 {% for entry in entries %}{% if entry.mac %}
```
mkurek: this if is not necessary IMO - `DHCPEntryManager` should handle it

src/ralph/dhcp/views.py
```diff
@@ -135,6 +136,11 @@ def get_last_modified(self, networks):
         IP (DHCP entry), ethernet.
         """
         last_items = []
+
+        try:
+            last_items.append(Deployment.objects.latest('modified').modified)
```
mkurek: shouldn't we filter deployments for this particular network(s) only? :) It'll probably be not so easy with current models

src/ralph/lib/external_services/base.py
```diff
@@ -44,7 +44,7 @@ def run(self, **kwargs):
         return job.result
 
     def run_async(self, **kwargs):
-        job = self.queue.enqueue(self.method, kwargs=kwargs)
+        job = self.queue.enqueue(self.method, kwargs=kwargs, timeout=6000)
```
mkurek: move the timeout to settings

src/ralph/deployment/tests/test_deployment.py
```diff
@@ -105,3 +109,20 @@ def _prepare_rack(self):
...
         self.instance.parent.rack = self.rack
         self.instance.parent.save()
+
+
+@ddt
+class AutocompleteFunctionsTestCase(TestCase):
+    @unpack
+    @data(
+        ([],),
+        ([DataCenterAssetFactory(), DataCenterAssetFactory()],)
```
mkurek: move it to the body of the function (or make it somehow lazy) - now these objects are created when this module is importing and not every table is created then.

src/ralph/deployment/deployment.py
```diff
@@ -266,9 +270,7 @@ def clean_dhcp(cls, instances, **kwargs):
     Clean DHCP entries for each instance.
     """
     for instance in instances:
-        mac_addresses = _get_non_mgmt_ethernets(instance).values_list(
-            'mac', flat=True
-        )
+        _get_non_mgmt_ethernets(instance).values_list('mac', flat=True)
```
szok, line 273: It is used somewhere?

mkurek, line 273: It'll be used below

src/ralph/deployment/views.py
ar4s: We could but now you can add simple `{{ kickstart }}` to template instead of `{{ ralph_instance }}{{ kickstart }}`. IMHO if user insert one variable to template then probably do it correctly.

mkurek: ok, 👍

src/ralph/lib/external_services/base.py
ar4s: done

src/ralph/deployment/urls.py
ar4s, line 7: done

src/ralph/deployment/views.py
ar4s, line 1: done

src/ralph/deployment/models.py
```diff
@@ -1,33 +1,68 @@
...
+        choices=get_choices_from_group(PrebootItemType.CONFIGURATION),
+        default=PrebootItemType.ipxe.id,
     )
-    raw_config = models.TextField(
-        _('raw config'),
+    configuration = models.TextField(
+        _('configuration'),
         blank=True,
         help_text=_(
             'All newline characters will be converted to Unix \\n '
```
mkurek, line 68: is this up to date?

src/ralph/deployment/models.py
```diff
@@ -59,13 +111,43 @@ class Preboot(NamedMixin):
...
+        )
+        if item is not None:
+            return item.configuration
+
 
-class DeploymentManager(models.Manager):
+class DeploymentManager(JobManager):
     def get_queryset(self):
         from ralph.deployment.deployment import deploy
         # TODO: test it
```
mkurek, line 153: add another TODO that this should return only Jobs for DataCenterAsset or VirtualServer

src/ralph/settings/base.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import json
 import os
+from collections import ChainMap
```
mkurek, line 4: 👍 👏

mkurek:
:ok:

PR #2445, opened 2016-05-27
Network form & API fixes and improvements

mkurek:
Added additional field to IP address: `dhcp_expose`. When it's checked, trio (hostname, mac, ip) will be exposed in DHCP config. There are some constraints for this (when `dhcp_expose` is selected) in GUI and API:
- mac, hostname or ip could not be changed
- `dhcp_expose` could be changed only by transition

So it's easy to add entry to DHCP, but more complicated to remove it from DHCP (requires executing transition).

### DIFF
--- a/src/ralph/networks/forms.py
+++ b/src/ralph/networks/forms.py
@@ -2,7 +2,7 @@
 from django import forms
 from django.conf import settings
 from django.forms import ValidationError
-# from django.forms.formsets import DELETION_FIELD_NAME
+from django.forms.formsets import DELETION_FIELD_NAME
 from django.forms.models import BaseInlineFormSet
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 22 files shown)


ar4s:
:ok:

src/ralph/networks/forms.py
```diff
@@ -30,13 +35,39 @@ def validate_is_management(forms):
...
 
 class SimpleNetworkForm(forms.ModelForm):
-    hostname = forms.CharField(label='Hostname')
-    address = forms.IPAddressField(label='IP address')
+    """
+    This form handles both Ethernet and IPAddress models.
+
+    Validations and checks:
+    * at least one of MAC and IP address is required
+    * when address is empty, none of ip fields (is_management_, hostname etc)
```
szok: should be is_managmened

szok:
🆗

PR #2446, opened 2016-05-31
Fixed handling DoesNotExist error in importer

mkurek:
ImportedForeignKeyWidget now raises DoesNotExist error (through super ForeignKeyWidget) instead of returning None. This behavior is the same as it's super-class, ForeignKeyWidget.


szok:
🆗

ar4s:
:ok:

PR #2447, opened 2016-06-02
Share params between transition actions

mkurek:
Transition actions could share params using `shared_params` dict.

Other fixes and improvements:
- async transition properly restore history and shared params when action was rescheduled
- start of refactorization of `ralph.lib.transitions.models`

TODO:
- [x] tests (both sync and async)
- [x] docs
- [x] cleanup

### DIFF
--- a/src/ralph/lib/transitions/async.py
+++ b/src/ralph/lib/transitions/async.py
@@ -7,6 +7,12 @@ from collections import defaultdict
 from django.db import transaction
 
 from ralph.attachments.models import Attachment
+from ralph.lib.transitions.exceptions import (
+    AsyncTransitionError,
+    FailedActionError,
+    MoreThanOneStartedActionError,
+    RescheduleAsyncTransitionActionLater
+)
 from ralph.lib.transitions.models import (
     _check_action_with_instances,
     _check_instances_for_transition,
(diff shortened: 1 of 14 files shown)


ar4s:
:ok:

src/ralph/lib/transitions/async.py
```diff
@@ -2,11 +2,16 @@
 Asynchronous runner for transitions
 """
 import logging
-from collections import defaultdict
 
 from django.db import transaction
 
 from ralph.attachments.models import Attachment
+from ralph.lib.transitions.exceptions import (
```
szok, line 9: 👍

src/ralph/lib/transitions/conf.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
 
-
+DEFAULT_ASYNC_TRANSITION_SERVICE_NAME = 'ASYNC_TRANSITIONS'
```
szok, line 3: Maybe better choice for this setting will be django settings?

mkurek, line 3: https://github.com/allegro/ralph/blob/feature/deployment/src/ralph/settings/base.py#L289 - it is stored in ralph settings as well

szok:
:ok:

PR #2448, opened 2016-06-03
Nullable hostname

mkurek:
Change `hostname` of `Asset` (`DataCenterAsset` and `BackOfficeAsset`) to `NullableCharField`. This will result in saving empty `hostname` as `null` instead of empty string (`''`) to the database, which might help in the future using `hostname` column as a "clickable" column in admin or to mark `hostname` as unique (`null != null`, but `'' == ''`).

There is also minor change in `NullableCharField` as well - `to_python` now returns `None` instead of `''` in case of empty value - this affects only read-only views (like admin listing) - form fields would work as usual.


szok:
:ok:

ar4s:
:ok:

PR #2449, opened 2016-06-03
Hostname as a first column of DataCenterAsset listing

mkurek:
Moved hostname to the first place in DataCenterAsset listing. Hostname is now clickable column (it provides link to the details view).


vi4m:
LGTM

ar4s:
:ok:

PR #2450, opened 2016-06-03
Deployment

ar4s:
Deployment for physical and virtual servers.

Features:
- configurable preboots (ipxe, kickstart, initrd, kernel)
- DHCP entries support
- deployment actions for transition


mkurek:
:ok:

PR #2455, opened 2016-06-06
Catch multiple hypervisors error in openstack sync.

mkurek:
MultipleObjectsReturned is catched too when hypervisor is beign searched in openstack sync.


ar4s:
:ok:

szok:
:ok:

PR #2469, opened 2016-06-09
Added DNSServer to API

mkurek:
Expose DNSServer in REST API (lack of it brokes network API).

PR #2463, opened 2016-06-07
Ralph2 sync using hermes

mkurek:

PR #2470, opened 2016-06-09
Revert "Add sputnik to travis.yml"

mkurek:
Reverts allegro/ralph#2451. Sputnik (pylint) is working terribly in django (Ralph) project. See #2469 or #2463 for examples. Maybe it'll work better after tweaking pylint but we don't have time for it now.


ar4s:
:ok:

PR #2472, opened 2016-06-09
Fixed management IP/hostname form for DataCenterAsset

mkurek:
Working updating and deleting mgmt ip/hostname. Fixed 500 when mgmt ip is duplicated.


ghost:
:ok:

src/ralph/data_center/forms.py
```diff
@@ -10,17 +14,88 @@ class DataCenterAssetForm(RalphAdminForm):
 
     ip_fields = ['management_ip', 'management_hostname']
 
+    class Meta:
+        model = DataCenterAsset
+        exclude = ()
```
szok: Why exclude is empty tuple? by default it not empty?

szok:
:ok:

ar4s:
:ok:

PR #2475, opened 2016-06-10
revert filter by service uid

mkurek:


ar4s:
:ok:

src/ralph/data_center/tests/test_api.py
```diff
@@ -176,6 +176,16 @@ def test_filter_by_service_uid(self):
             response.data['count'], 1
         )
 
+    def test_filter_by_service_uid2(self):
+        url = reverse('datacenterasset-list') + '?service_env__service__uid={}'.format(  # noqa
```
xor-xor, line 180: Not sure if muting QA checks for the whole line is a good idea here...

xor-xor:
:ok:

PR #2480, opened 2016-06-13
Added networks tab to VirtualServer

mkurek:
Unified networks tab between DataCenterAsset and VirtualServer. In addition improve RalphDetailView admin class attributes (ex. inlines) - now they could be inherited from base class.


ghost:
:ok:

szok:
:ok:

PR #2482, opened 2016-06-13
Improved running transition through API

mkurek:
- use 'choices' param for ChoiceField (ex. callback function)
- return 400 in case of error, 201 when OK for sync transition and 202 for async
- additional tests for assigning new hostname through API

### DIFF
--- a/src/ralph/deployment/tests/test_transitions.py
+++ b/src/ralph/deployment/tests/test_transitions.py
@@ -1,0 +1,266 @@
+from django.contrib.auth import get_user_model
+from django.core.urlresolvers import reverse
+from django.test import Client
+from rest_framework.test import APIClient
+
+from ralph.assets.models import AssetLastHostname
+from ralph.data_center.models import DataCenterAsset
+from ralph.data_center.tests.factories import (
+    DataCenterAssetFactory,
+    RackFactory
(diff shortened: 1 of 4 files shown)


src/ralph/deployment/tests/test_transitions.py
```diff
@@ -0,0 +1,267 @@
...
+    DataCenterAssetFactory,
+    RackFactory
+)
+from ralph.lib.transitions.conf import TRANSITION_ORIGINAL_STATUS
+from ralph.lib.transitions.tests import TransitionTestCase
+from ralph.networks.tests.factories import (
+    NetworkEnvironmentFactory,
+    NetworkFactory
+)
+# from ralph.virtual.models import VirtualServer
```
szok: Why commented?

mkurek: for the same reason as below code - it was waiting for separated PR to be merged (#2481).

szok:
:ok:

src/ralph/deployment/tests/test_transitions.py
szok: ok

ghost:
:ok:

PR #2484, opened 2016-06-13
network components in API improvements & performance improvements for DataCenterAsset and Virtual Server

mkurek:
- filter ethernet by ipaddress
- attach IP info in ethernet
- filter IP by mac address
- attach ethernet (and simple IP) info for DataCenterAsset and VirtualServer
- various filters and performance (sql queries count) optimizations

### DIFF
--- a/src/ralph/data_center/tests/factories.py
+++ b/src/ralph/data_center/tests/factories.py
@@ -10,6 +10,7 @@ from ralph.assets.tests.factories import (
     BudgetInfoFactory,
     ConfigurationClassFactory,
     DataCenterAssetModelFactory,
+    EthernetWithIPAddressFactory,
     ServiceEnvironmentFactory
 )
 from ralph.data_center.models.physical import (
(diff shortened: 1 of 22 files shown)


szok:
:ok:

src/ralph/virtual/models.py
```diff
@@ -252,6 +252,7 @@ class VirtualServer(AdminAbsoluteUrlMixin, NetworkableBaseObject, BaseObject):
         verbose_name=_('SN'),
         unique=True,
     )
+    # TODO: remove this field
```
ghost, line 255: when?

src/ralph/data_center/tests/factories.py
```diff
@@ -115,6 +117,44 @@ class Meta:
         model = DataCenterAsset
 
 
+class DataCenterAssetFullFactory(DataCenterAssetFactory):
```
ghost, line 120: :+1:

src/ralph/data_center/tests/factories.py
```diff
@@ -115,6 +117,44 @@ class Meta:
...
+        quantity=3
+    )
+    support1 = factory.RelatedFactory(
+        'ralph.supports.tests.factories.BaseObjectsSupportFactory',
+        'baseobject'
+    )
+    support2 = factory.RelatedFactory(
+        'ralph.supports.tests.factories.BaseObjectsSupportFactory',
+        'baseobject'
+    )
```
ghost, line 155: Shouldn't be done this way?
http://factoryboy.readthedocs.io/en/latest/recipes.html#simple-many-to-many-relationship

src/ralph/virtual/models.py
mkurek, line 255: not everything at once :) i'll do separated PR for it.

src/ralph/data_center/tests/factories.py
mkurek, line 155: The usage there is `UserFactory.create(groups=(group1, group2, group3))` - so we'll have to provide explicitly supports and other related objects that we want to assign here. In my solution, support is created implicitly, so you just have to call `DataCenterAssetFullFactory()`

src/ralph/virtual/models.py
ghost, line 255: I meant.. "add some hint to others, that they could delete it, if condition will meet" :)
of if you handle that then it can be ignored

ghost:
:ok:

PR #2485, opened 2016-06-14
Fixed autocomplete layout.

ghost:
Before: `delete` icon was missing because of bad layout.
Now: `delete` icon is visible, layout fixed.


mkurek:
:ok:

PR #2486, opened 2016-06-15
Use default table for BaseObjectsSupport

mkurek:
Migration 0003 in supports created through model for BaseObjectSupport, keeping old DB table name, which breaks tests with SKIP_MIGRATION=1. This commit rename this table (to use default name for this model) and make this model managable.


ghost:
:ok:

szok:
:ok:

PR #2487, opened 2016-06-15
Memory component

mkurek:
TODO: 
- [x] tests
- [x] component model re-consideration (do we need this as it is right now - maybe regular charfield will be enough here)


szok:
:ok:

ghost:
:ok:

PR #2489, opened 2016-06-15
Change `%(class)s` to `%(class)s_set` as related model in component

mkurek:
Example change: ethernet -> ethernet_set

Singular name (ex. ethernet) could suggest, that there is only one related component. In fact there could be more or each type. Thic change should make it now more clear.


ghost:
:ok:

szok:
+100000
:ok:

PR #2494, opened 2016-06-22
Better way of disabling publishers during handlind subscribers in Ralph3 sync

mkurek:
Previously _handle_post_save attribute was assigned to the obj, which should prevent firing post_save publisher on it. Unfortunately it wasn't working perfectly - obj could be saved somewhere underneath and thus it wouldn't have this attr assigned, which caused signal (and publisher) to be fired, which may cause pub-sub cycle between Ralph2 and Ralph3.

This PR solves it by temporarly disabling post_save publisher signal during handling subscriber request.

Since signals (connectors) are not thread-safe, you may want to run runserver with --nothreading option (this would work with gunicorn out of the box, since it's using processes instead of threads).

### DIFF
--- a/src/ralph/export_to_ng/helpers.py
+++ b/src/ralph/export_to_ng/helpers.py
@@ -1,0 +1,30 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class WithSignalDisabled(object):
+    """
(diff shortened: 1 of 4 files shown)


szok:
:ok:

ghost:
:ok:

PR #2496, opened 2016-06-23
menu resync fix & tests

mkurek:
Fix menu resyncing (unpacking list to another list work only in python 3.5: http://stackoverflow.com/questions/3870778/why-does-this-cause-a-syntax-error)


szok:
:ok:

ghost:
:ok:

PR #2497, opened 2016-06-24
Fixed Ralph2 sync device subscriber mgmt ip assignment

mkurek:
- Delete mgmt when ip is empty, otherwise assign it
- Bumped pyhermes version to 0.1.3


szok:
:ok:

ghost:
:ok:

PR #2500, opened 2016-06-27
Venture (ConfigurationModule/Class) sync from Ralph2 - subscription handler

mkurek:
New subscription handler for Venture (ConfigurationModule) and VentureRole (ConfigurationClass) for sync with Ralph2.


ghost:
:ok: though test names could be more descriptive or/and more given/when/then

ar4s:
:ok:

PR #2501, opened 2016-06-27
Unique constraint on imported object

mkurek:
New unique constraint for content type, object_pk and old_object_pk for ImportedObject model.


ar4s:
:ok:

ar4s:
:ok:

ghost:
:ok:

PR #2502, opened 2016-06-28
Add sync custom fields (names, values and relations).

ar4s:
- [x] add mgmt command for sync


mkurek:
:ok:

src/ralph/export_to_ng/publishers.py
```diff
@@ -7,13 +7,23 @@
...
+    Venture,
+    VentureRole
+)
 from ralph.discovery.models import Device
 
 logger = logging.getLogger(__name__)
 
 
-def ralph3_sync(model):
+ROLE_PROPERTY_WHITELIST = getattr(
```
mkurek: why not just put this to the settings with default value? we should read it from env variable either way

src/ralph/export_to_ng/publishers.py
```diff
@@ -81,8 +91,29 @@ def sync_device_to_ralph3(sender, instance=None, created=False, **kwargs):
         'service': device.service.uid if device.service else None,
         'environment': device.device_environment_id,
         'venture_role': device.venture_role_id,
+        'custom_fields': {
+            k: v for k, v in device.get_property_set().items()
+            if k in settings.RALPH2_HERMES_ROLE_PROPERTY_WHITELIST
+        },
     }
-    return data
+    return {k: v for k, v in data.items()} if fields else data
```
mkurek: are you sure this is working? :) what's the difference between `{k: v for k, v in data.items()}` and `data`? :)

src/ralph/settings.py
```diff
@@ -809,3 +809,8 @@
     # non-conflicting. this is achieved by adding below constant
     'discovery_datacenter_constant': 1000,
 }
+
+import json
+RALPH2_HERMES_ROLE_PROPERTY_WHITELIST = json.loads(
```
mkurek: it's ralph2 - there is no need to read it from env

mkurek:
:ok:

PR #2503, opened 2016-06-28
Management command to sync to Ralph3

mkurek:
`ralph sync_to_ralph3 ...` will force syncing particular models to Ralph3.

Supported models:
- Venture
- VentureRole


ar4s:
:ok:

PR #2504, opened 2016-06-28
Add subscriber for custom fields

ar4s:


src/ralph/lib/custom_fields/models.py
```diff
@@ -159,3 +159,15 @@ class WithCustomFieldsMixin(models.Model):
 
     class Meta:
         abstract = True
+
+    @property
+    def custom_fields_as_dict(self):
+        return dict(self.custom_fields.values_list(
+            'custom_field__name', 'value'
+        ).all())
```
mkurek: all is not necessary here as well

src/ralph/lib/custom_fields/models.py
```diff
@@ -159,3 +159,15 @@ class WithCustomFieldsMixin(models.Model):
...
+
+    def update_custom_field(self, name, value):
+        cf = CustomField.objects.get(name=name)
+        cfv, _ = self.custom_fields.get_or_create(custom_field=cf)
+        cfv.value = value
+        cfv.save(update_fields=['value'])
```
mkurek, line 173: 👍

mkurek:
:ok:

PR #2505, opened 2016-06-28
Device's venture, role and custom fields cross validator

mkurek:


src/ralph/cross_validator/mappers.py
```diff
@@ -57,6 +57,24 @@ def diff(old, new):
     return diff
 
 
+def custom_fields_diff():
+    def diff(old, new):
```
ar4s: nested function? IMO is not necessary..

src/ralph/lib/custom_fields/models.py
```diff
@@ -159,3 +159,8 @@ class WithCustomFieldsMixin(models.Model):
 
     class Meta:
         abstract = True
+
+    def get_custom_fields(self):
```
ar4s: I created something similar see [my solution](https://github.com/allegro/ralph/pull/2504/files#diff-451bb9acae6c0c8fd98ffa81a06325d2R164)

ar4s:
:ok:

PR #2509, opened 2016-06-29
Ralph3 sync mgmt cmd improvements

mkurek:
- call post_save signal directly instead of calling .save()
- new `sleep` option for sleeping between succesive triggers
- sync ventures assigned to hosts (DC Asset) and custom fields assigned to hosts


ar4s:
:ok:

PR #2510, opened 2016-06-30
Virtual server publisher

ar4s:


src/ralph/export_to_ng/publishers.py
```diff
@@ -169,3 +170,19 @@ def sync_role_property_to_ralph3(sender, instance=None, created=False, **kwargs)
...
+    return {
+        'id': instance.id,
+        'type': instance.model.name if instance.model else None,
+        'hostname': instance.name,
+        'sn': instance.sn,
+        'service': instance.service.uid if instance.service else None,
+        'environment': instance.device_environment_id,
+        'venture_role': instance.venture_role_id,
+        'parent_id': instance.parent_id,
+    }
```
mkurek, line 193: what about custom fields?

src/ralph/export_to_ng/publishers.py
```diff
@@ -169,3 +170,19 @@ def sync_role_property_to_ralph3(sender, instance=None, created=False, **kwargs)
...
+        return
+    return {
+        'id': instance.id,
+        'type': instance.model.name if instance.model else None,
+        'hostname': instance.name,
+        'sn': instance.sn,
+        'service': instance.service.uid if instance.service else None,
+        'environment': instance.device_environment_id,
+        'venture_role': instance.venture_role_id,
+        'parent_id': instance.parent_id,
```
mkurek: ralph3 don't have info about Device ids; we should use asset id here

src/ralph/export_to_ng/publishers.py
```diff
@@ -169,3 +170,24 @@ def sync_role_property_to_ralph3(sender, instance=None, created=False, **kwargs)
         'choices': choices
     }
     return data
+
+
+@ralph3_sync(Device)
+def sync_virtual_server_to_ralph3(sender, instance=None, created=False, **kwargs):
+    if instance.model.type != DeviceType.virtual_server:
+        return
+    asset = instance.parent.get_asset(manager='admin_objects')
```
mkurek: what if instance will have empty parent?

mkurek:
:ok:

PR #2512, opened 2016-06-30
Filter BaseObject and DCHost in API by conf. path

mkurek:
New filters for BaseObject and DCHost in API: configuration_path and configuration_path__module__name


ar4s:
:ok:

PR #2513, opened 2016-06-30
Subscriber for  virtual server

ar4s:
- [x] add tests


src/ralph/ralph2_sync/subscribers.py
```diff
@@ -120,15 +155,9 @@ def sync_device_to_ralph3(data):
...
-            service__uid=service,
-            environment=ImportedObjects.get_object_from_old_pk(
-                Environment, environment
-            )
-        )
+    service_env = _get_service_env(data)
+    if service_env:
+        dca.service_env = service_env
     if 'venture_role' in data:
         if data['venture_role']:
```
mkurek, line 172: use `_get_configuration_path_from_venture_role` here

src/ralph/ralph2_sync/subscribers.py
```diff
@@ -120,15 +155,9 @@ def sync_device_to_ralph3(data):
...
-        service = data['service']
-        environment = data['environment']
-        dca.service_env = ServiceEnvironment.objects.get(
-            service__uid=service,
-            environment=ImportedObjects.get_object_from_old_pk(
-                Environment, environment
-            )
-        )
+    service_env = _get_service_env(data)
+    if service_env:
```
mkurek: what if it's None? shouldn't we assign empty service_env to dca as well?

mkurek:
do you plan to write some tests for this subscriber? :)

src/ralph/data_importer/models.py
```diff
@@ -24,7 +24,7 @@ def __str__(self):
         )
 
     class Meta:
-        unique_together = [('content_type', 'object_pk', 'old_object_pk')]
+        unique_together = [('content_type', 'object_pk')]
```
mkurek: could we exclude it to separated commit/PR?

src/ralph/ralph2_sync/subscribers.py
```diff
@@ -129,31 +166,16 @@ def sync_device_to_ralph3(data):
...
-    if 'service' in data and 'environment' in data:
-        service = data['service']
-        environment = data['environment']
-        dca.service_env = ServiceEnvironment.objects.get(
-            service__uid=service,
-            environment=ImportedObjects.get_object_from_old_pk(
-                Environment, environment
-            )
-        )
+    dca.service_env = _get_service_env(data)
```
mkurek: wrap it in if from 132

mkurek:
:ok:

PR #2514, opened 2016-06-30
Conf path venture publisher

ar4s:


src/ralph/ralph2_sync/subscribers.py
```diff
@@ -226,7 +226,6 @@ def sync_venture_role_to_ralph3(data):
     Supported fields:
     * symbol
     * parent
```
mkurek: parent -> module

mkurek:
:ok:

PR #2518, opened 2016-07-01
Remove ``old_object_pk`` from unique set in ImportedObjects

ar4s:


mkurek:
:ok: (but don't merge it now)

PR #2521, opened 2016-07-01
Cross validator improvements

mkurek:
- fixed DeviceInfo <-> Device connection (OneToOne field, added null=True)
- new checker is asset has linked device


ar4s:
:ok:

PR #2523, opened 2016-07-04
Fixed NG partial device sync

mkurek:
Fixed management command to partially sync Device to Ralph3 (removed limiting query result to 10 elements)


ar4s:
:ok:

PR #2526, opened 2016-07-04
Disable configuration module/class publishers during Ralph2 sync in subscriber

mkurek:
Add `sync_configuration_class_to_ralph2` and `sync_configuration_module_to_ralph2` to disabled publishers.


ar4s:
:ok:

PR #2529, opened 2016-07-04
Update DC Asset's subscriber

ar4s:
- add venture
- add venture_role


src/ralph/export_to_ng/subscribers.py
```diff
@@ -167,6 +167,8 @@ def sync_dc_asset_to_ralph2_handler(data):
     device = asset.get_ralph_device()
     device.name = data['hostname']
     device.management_ip = data.get('management_ip')
+    device.venture_id = data['venture']
```
mkurek: maybe wrap it in some if (check if venture is present in the data), so not-releasing ralph3 will not block us in releasing this.

mkurek:
:ok:

PR #2530, opened 2016-07-04
Add extra condition to prevent AttributeError

ar4s:


src/ralph/export_to_ng/publishers.py
```diff
@@ -174,7 +174,7 @@ def sync_role_property_to_ralph3(sender, instance=None, created=False, **kwargs)
 
 @ralph3_sync(Device)
 def sync_virtual_server_to_ralph3(sender, instance=None, created=False, **kwargs):
-    if instance.model.type != DeviceType.virtual_server:
+    if instance.model and instance.model.type != DeviceType.virtual_server:
```
mkurek: in this case, when device don't have any model assigned, this condition would not pass (thus not return), which is wrong. Maybe sth like: `if not instance.model or instance.model.type != DeviceType.virtual_server:`?

mkurek:
:ok:

PR #2536, opened 2016-07-05
VirtualServer cross validator

mkurek:
Compare virtual server in Ralph2 vs Ralph3. Compared fields/properties:
- hostname
- sn
- parent
- service
- configuration_path
- custom fields


ar4s:
:ok:

PR #2537, opened 2016-07-05
Venture role publisher - full name instead of name

mkurek:
Publish full name of venture role (includes parent names) instead of single name


ar4s:
:ok:

PR #2540, opened 2016-07-05
Ralph3 sync custom fields fix when value is None

mkurek:
When CF value is None, send is as empty string.

### DIFF
--- a/src/ralph/export_to_ng/tests/test_publishers.py
+++ b/src/ralph/export_to_ng/tests/test_publishers.py
@@ -6,6 +6,7 @@ from ralph.business.models import (
     RoleProperty,
     RolePropertyType,
     RolePropertyTypeValue,
+    RolePropertyValue,
     Venture,
     VentureRole,
 )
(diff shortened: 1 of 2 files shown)


ar4s:
:ok:

src/ralph/export_to_ng/tests/test_publishers.py
```diff
@@ -104,8 +105,15 @@ def test_publish_device_full(self):
     def test_devices_properties(self):
         property_symbol = 'test_symbol'
         property_value = 'test_value'
+        property_symbol2 = 'test_symbol2'
         self.device.venture_role.roleproperty_set.create(symbol=property_symbol)  # noqa
+        prop2 = self.device.venture_role.roleproperty_set.create(
+            symbol=property_symbol2
+        )
         self.device.set_property(property_symbol, property_value, None)
+        RolePropertyValue.get_or_create(
```
ar4s: `RolePropertyValue.objects.get_or_create`

PR #2543, opened 2016-07-06
Add info about latest diff between R2 and R3

ar4s:


.travis.yml
```diff
@@ -5,6 +5,7 @@ env:
         - DATABASE_USER=travis
         - DATABASE_PASSWORD=travis
         - DJANGO_CURRENT_VERSION=1.8.12
+        - RALPH2_RALPH3_CROSS_VALIDATION_ENABLED=1
```
mkurek, line 8: when is it used?

mkurek:
:ok:

.travis.yml
ar4s, line 8: I added because of this https://github.com/allegro/ralph/pull/2543/files#diff-fad12832045cc87fe41573929b84d835R67

PR #2544, opened 2016-07-06
Components: added support for fibre channel cards.

xor-xor:
Ready for review/merge.


ar4s:
:ok:

src/ralph/assets/models/components.py
```diff
@@ -177,3 +181,28 @@ class Meta:
...
+        verbose_name=_('WWN'), max_length=255, unique=True, null=True,
+        blank=True, default=None,
+    )
+
+    class Meta:
+        verbose_name = _('fibre channel card')
+        verbose_name_plural = _('fibre channel cards')
+
+    def __str__(self):
+        return '{}'.format(self.model)
```
mkurek: consider using wwn here (and maybe model_name instead of model?)

src/ralph/data_center/api/views.py
```diff
@@ -55,6 +55,7 @@ class DataCenterAssetViewSet(BaseObjectViewSetMixin, RalphAPIViewSet):
             'ethernet_set',
             queryset=Ethernet.objects.select_related('ipaddress')
         ),
+        'fibrechannelcard_set',  # XXX or 'fibre_channel_card_set'..?
```
mkurek: what about this comment? `fibrechannelcard_set` seems to be working since your tests are passing (otherwise you'll get some error on this prefetch_related)

src/ralph/assets/tests/factories.py
```diff
@@ -35,6 +35,20 @@ def next_mac(n):
...
+        0xaa,
+        0xbb,
+        0xcc,
+        0xdd,
+        0xee,
+        n >> 16 & 0xff,
+        n >> 8 & 0xff,
+        n & 0xff
+    ]
+    return ''.join(map(lambda x: '%02x' % x, wwn))
```
xor-xor, line 58: It's basically a copy of `next_mac` from the same file (lines 26-35), which was written by you or @ar4s :wink:

EDIT: I will include some examples how it works, for clarity.

src/ralph/data_center/api/views.py
xor-xor: Yes, right - I forgot to remove it.

src/ralph/assets/tests/factories.py
mkurek, line 58: yes, but it's still magical to me :see_no_evil:

src/ralph/assets/models/components.py
xor-xor: Right. I'm thinking about something like `return 'model: "{}", WWN: "{}"'.format(self.model_name, self.wwn)` -  those extra labels and double quotes are for the case when some of those fields (or both of them) would be empty, so we'd get something like `<FibreChannelCard: model: "", WWN: "">` instead of just `<FibreChannelCard:  >`. What do you think..?

mkurek: :ok: for me

mkurek:
:ok:

PR #2545, opened 2016-07-07
Use VentureRole.get_properties instead of Device.get_property_set for Ralph3 sync

mkurek:
`Device.get_property_set` all `RolePropertyValue`s attached to `Device`, which might comes from some previous role of this device, but it's still attached to this device in DB - but it's NOT visible in GUI or through `puppet-classifier` endpoint, so we should not sync it as well. `puppet-classifier` use `VentureRole.get_properties` method, which returns only properties in scope of current `VentureRole`, so it should be used in Ralph3 sync too.


ar4s:
:ok:

PR #2546, opened 2016-07-07
Use VentureRole.get_properties instead of Device.get_property_set for Ralph2 validator

mkurek:
See #2545 for details (TL;DR: Device.get_property_set returns invalid data in context of current VentureRole)


ar4s:
:ok:

PR #2547, opened 2016-07-07
Sync subscriber for device

ar4s:


src/ralph/export_to_ng/subscribers.py
```diff
@@ -61,6 +72,23 @@ def _get_publisher_signal_info(func):
     }
 
 
+def _handle_custom_fields(data, obj):
+    for field, value in data['custom_fields'].items():
+        if field not in settings.RALPH2_HERMES_ROLE_PROPERTY_WHITELIST:
+            continue
+        try:
+            prop = RoleProperty.objects.get(symbol=field)
```
mkurek: it won't be so simple. there might be many `RoleProperty`with this symbol (defined for many ventures, roles). consider using `Device.set_property`

PR #2548, opened 2016-07-07
Add _handle_custom_fields to device's subscriber

ar4s:


src/ralph/export_to_ng/subscribers.py
```diff
@@ -61,6 +72,25 @@ def _get_publisher_signal_info(func):
...
+        try:
+            prop = obj.set_property(symbol, value, None)
+        except RoleProperty.DoesNotExist:
+            logger.error(
+                'RoleProperty with symbol {} doesn\'t exist'.format(symbol)
+            )
+            continue
+        if not RolePropertyTypeValue.objects.filter(type=prop.type, value=value).exists():  # noqa
+            RolePropertyTypeValue.objects.create(type=prop.type, value=value)
+        RolePropertyValue.objects.filter(
```
mkurek: is this necessary? isn't `obj.set_property` enough?

src/ralph/export_to_ng/subscribers.py
```diff
@@ -61,6 +72,25 @@ def _get_publisher_signal_info(func):
...
+            continue
+        prop = None
+        try:
+            prop = obj.set_property(symbol, value, None)
+        except RoleProperty.DoesNotExist:
+            logger.error(
+                'RoleProperty with symbol {} doesn\'t exist'.format(symbol)
+            )
+            continue
+        if not RolePropertyTypeValue.objects.filter(type=prop.type, value=value).exists():  # noqa
```
mkurek: shouldn't this be executed before `obj.set_property`?

src/ralph/export_to_ng/subscribers.py
```diff
@@ -61,6 +70,18 @@ def _get_publisher_signal_info(func):
     }
 
 
+def _handle_custom_fields(data, obj):
+    for symbol, value in data['custom_fields'].items():
```
mkurek: maybe `data.get('custom_fields', {}).items()`? Otherwise we have to release ng first - make it backward compatible.

PR #2550, opened 2016-07-07
Add publisher for IP adresses

ar4s:


src/ralph/export_to_ng/publishers.py
```diff
@@ -194,3 +195,34 @@ def sync_virtual_server_to_ralph3(sender, instance=None, created=False, **kwargs
...
+        'dhcp_expose': False
+    }
+
+
+@ralph3_sync(IPAddress)
+def sync_ipaddress_to_ralph3(sender, instance=None, created=False, **kwargs):
+    return {
+        'id': instance.id,
+        'address': instance.address,
+        'device_id': instance.device_id,
```
mkurek: it wont be so simple :) IMO we should return here more information, like type (asset, virtual etc) and then it's id. In case of physical device, return its asset id

src/ralph/export_to_ng/publishers.py
```diff
@@ -194,3 +195,34 @@ def sync_virtual_server_to_ralph3(sender, instance=None, created=False, **kwargs
...
+        'hostname': None,
+        'is_management': False,
+        'is_gateway': True,
+        'dhcp_expose': False
+    }
+
+
+@ralph3_sync(IPAddress)
+def sync_ipaddress_to_ralph3(sender, instance=None, created=False, **kwargs):
+    return {
```
mkurek, line 347: i don't know for sure, but maybe it'll be better to sync it as a part of `sync_device_to_ralph3`/`sync_virtual_server_to_ralph3`? 
Arguments:
- we don't want to store empty IP addresses (without assigned device) in Ralph3
- right now, scan saves IP address after every scan, so (almost) all of our ip addresses will be synced every time our networks are scanned
- we can't just check if ip has device assigned - there could be the case, where device is detached in current request (save) from ip (but we could workaround it, probably)

ar4s, line 347: +1

src/ralph/export_to_ng/publishers.py
```diff
@@ -194,3 +229,5 @@ def sync_virtual_server_to_ralph3(sender, instance=None, created=False, **kwargs
         'parent_id': asset.id if asset else None,
         'custom_fields': _get_custom_fields(instance),
     }
+    data.update(_get_ips_list(instance))
```
mkurek: shouldn't ips be treated just like any other field? if it is on fields list, it will be included in result, otherwise no (in your case ips are always included in result).

src/ralph/export_to_ng/publishers.py
```diff
@@ -51,8 +53,8 @@ def wrapped_func(sender, instance=None, **kwargs):
                 # `_handle_post_save` set to False
                 getattr(instance, '_handle_post_save', True)
             ):
+                result = func(sender, instance, **kwargs)
```
mkurek: undo it

src/ralph/export_to_ng/publishers.py
```diff
@@ -91,24 +93,57 @@ def _get_custom_fields(device):
...
+            ))
+        return mac
+
+    for ip in device.ipaddress_set.all():
+        ips_list.append({
+            'address': ip.address,
+            'hostname': ip.hostname,
+            'is_management': ip.is_management,
+            'mac': get_mac(ip),
+            'dhcp_expose': in_dhcp(ip)
```
mkurek: imo we should check here just if mac is not empty: 

```
mac = get_mac(ip)
{
..
'mac': mac,
'dhcp_expose': bool(mac),
}
```

mkurek:
:ok:

PR #2553, opened 2016-07-08
Change url in communication with DNSaaS and add message after save

szok:
- Change DNSaaS urls
- Add message after user save or delete record
- Add build_url method to generate urls for DNSaas

### DIFF
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -23,6 +23,31 @@ class DNSaaS:
             _headers.update(headers)
         self.session.headers.update(_headers)
 
+    def build_url(self, resource_name, id=None, version='v2', get_params=None):
+        """
+        Return Url for DNSAAS endpoint
+
+        Args:
+            resource_name: Resource name (example: domains)
+            id: Record ID,
+            version: Api Version
+            get_params: List of tuple to URL GET params: (example:
+                [('name', 'value')]
+            )
+        Returns:
+            string url
+        """
+        result_url = urljoin(
+            settings.DNSAAS_URL,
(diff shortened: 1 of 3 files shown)


src/ralph/dns/dnsaas.py
```diff
@@ -47,7 +47,7 @@ def get_dns_records(self, ipaddresses):
         ipaddresses = [('ip', i) for i in ipaddresses]
         url = urljoin(
             settings.DNSAAS_URL,
-            'api/records/?{}'.format(
+            'api/v2/records/?{}'.format(
```
ar4s: can you extract version to variable?

ar4s: or create some function to build url - some like this:
`
def build_url(resource_name, id=None, version='v2', **get_kwargs)
`

szok: 👍

src/ralph/dns/views.py
```diff
@@ -71,14 +73,26 @@ def post(self, request, *args, **kwargs):
...
+                    messages.success(
+                        request, _('DNS record has been deleted.')
+                    )
             elif posted_form.cleaned_data.get('pk'):
                 errors = self.dnsaas.update_dns_record(
                     posted_form.cleaned_data
                 )
+                if not errors:
+                    messages.success(
+                        request, _('DNS record has been saved.')
```
mkurek: if you use deleted/created in other cases, then there `updated` suits better than `saved`.

ar4s:
:ok:

mkurek:
:ok:

PR #2554, opened 2016-07-08
DCAsset's configuration_path Ralph2 sync fix

mkurek:
Set configuration_path to None when venture_role in data is None


ar4s:
:ok:

PR #2555, opened 2016-07-08
Show base object service owners in API

mkurek:
New fields `business_owners` and `technical_owners` added to `DataCenterAsset`, `VirtualServer`, `CloudHost`, `Cluster` and `ServiceEnvironment` serializers. It's list of short user info (username, first and last name).


ar4s:
:ok:

ar4s:
:ok:

PR #2557, opened 2016-07-08
Add `use as configuration variable` property to custom fields

mkurek:
- this field allow to specify variables which should be exposed to (for example) configuration management tools
- new API field: `configuration_variables`, which contains subset of `custom_fields` field - only the ones with `use_as_configuration_variable` checked


vi4m:
+1

ar4s:
:ok:

PR #2561, opened 2016-07-11
Components: add support for processors.

xor-xor:
Ready for review.

@mkurek - do we want to add this component to `ralph.virtual.api`..? I guess we want, because [memory is already there](https://github.com/allegro/ralph/pull/2487/files).

And also, see my remark at `ProcessorViewSet` (`filter_fields`).

### DIFF
--- a/src/ralph/assets/api/serializers.py
+++ b/src/ralph/assets/api/serializers.py
@@ -343,6 +343,8 @@ class ProcessorSerializer(ProcessorSimpleSerializer):
 # used by DataCenterAsset and VirtualServer serializers
 class NetworkComponentSerializerMixin(OwnersFromServiceEnvSerializerMixin):
     ethernet = EthernetSimpleSerializer(many=True, source='ethernet_set')
+    memory = MemorySimpleSerializer(many=True, source='memory_set')
+    processor = ProcessorSimpleSerializer(many=True, source='processor_set')
     ipaddresses = fields.SerializerMethodField()
 
     def get_ipaddresses(self, instance):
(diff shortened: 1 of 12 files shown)


src/ralph/assets/api/views.py
```diff
@@ -152,6 +152,13 @@ class FibreChannelCardViewSet(RalphAPIViewSet):
     prefetch_related = ['base_object', 'base_object__tags']
 
 
+class ProcessorViewSet(RalphAPIViewSet):
+    queryset = models.Processor.objects.all()
+    serializer_class = serializers.ProcessorSerializer
+    filter_fields = ['base_object']  # XXX what else we may need here..?
```
mkurek: cores?

mkurek:
@xor-xor yes - add it to [ComponentSerializerMixin](https://github.com/allegro/ralph/pull/2561/files#diff-889fa9efd0f7b9649338bd17e38febf5R329) (it'll be attached to VirtualServer in one of my PRs: https://github.com/allegro/ralph/pull/2555/files#diff-7ef1276786ff2fdc9106fa5ea385678eR165)

src/ralph/assets/api/serializers.py
```diff
@@ -322,9 +327,24 @@ class Meta:
...
+    class Meta:
+        depth = 1
+        model = Processor
+        exclude = ('model',)
+
+
 # used by DataCenterAsset and VirtualServer serializers
 class NetworkComponentSerializerMixin(OwnersFromServiceEnvSerializerMixin):
     ethernet = EthernetSimpleSerializer(many=True, source='ethernet_set')
+    memory = MemorySimpleSerializer(many=True, source='memory_set')
```
mkurek: memory is used in `ComponentSerializerMixin` (which inherits from `NetworkComponentSerializerMixin`). Please move processor there as well.

mkurek:
:ok:

PR #2563, opened 2016-07-12
Fix #2562

### LINKED ISSUE #2562: default: cp: cannot stat ‘/vagrant/conf/nginx.conf’: No such file or directory
Fresh install on vagrant up dies with this error:
default: cp: cannot stat ‘/vagrant/conf/nginx.conf’: No such file or directory
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.

ar4s:
Added missing file (`nginx.conf`) for vagrant's installation

### DIFF
--- a/vagrant/conf/nginx.conf
+++ b/vagrant/conf/nginx.conf
@@ -1,0 +1,14 @@
+server {
+    listen 80 default_server;
+    listen [::]:80 default_server ipv6only=on;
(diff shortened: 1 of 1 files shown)


mkurek:
:ok:

Issue #2478, opened 2016-06-10
500 inernal after upgrade ralph-core


daro1337:
Hey guys I have got 500 internal after upgrade ralph-core from 3.0.0-snapshot-20160607-5736 to  3.0.0-snapshot-20160610
this did help https://github.com/allegro/ralph/issues?page=2&q=is%3Aissue+is%3Aopen

Error:

ralph migrate
/opt/ralph/ralph-core/lib/python3.4/site-packages/django/contrib/contenttypes/models.py:161: RemovedInDjango19Warning: Model class django.contrib.contenttypes.models.ContentType doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class ContentType(models.Model):

/opt/ralph/ralph-core/lib/python3.4/site-packages/django/contrib/admin/models.py:28: RemovedInDjango19Warning: Model class django.contrib.admin.models.LogEntry doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class LogEntry(models.Model):

/opt/ralph/ralph-core/lib/python3.4/importlib/_bootstrap.py:321: RemovedInDjango19Warning: django.contrib.contenttypes.generic is deprecated and will be removed in Django 1.9. Its contents have been moved to the fields, forms, and admin submodules of django.contrib.contenttypes.
  return f(_args, *_kwds)

/opt/ralph/ralph-core/lib/python3.4/importlib/_bootstrap.py:321: RemovedInDjango19Warning: The utilities in django.db.models.loading are deprecated in favor of the new application loading system.
  return f(_args, *_kwds)

/opt/ralph/ralph-core/lib/python3.4/site-packages/django/contrib/auth/models.py:41: RemovedInDjango19Warning: Model class django.contrib.auth.models.Permission doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class Permission(models.Model):

/opt/ralph/ralph-core/lib/python3.4/site-packages/django/contrib/auth/models.py:98: RemovedInDjango19Warning: Model class django.contrib.auth.models.Group doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class Group(models.Model):

[...]

PR #2565, opened 2016-07-12
Fix ``LookupError`` when RALPH2_RALPH3_CROSS_VALIDATION_ENABLED is off

ar4s:
Fixed problem with cross-validator (https://github.com/allegro/ralph/issues/2478#issuecomment-231994849)


mkurek:
:ok:

PR #2572, opened 2016-07-13
Add support for racks and DNS servers in network subscriber

ar4s:


src/ralph/ralph2_sync/subscribers.py
```diff
@@ -325,6 +326,16 @@ def sync_venture_role_to_ralph3(data):
     logger.info('Synced configuration class {}'.format(conf_class))
 
 
+def _handle_m2m(ids, model, target, field):
+    if ids:
+        objs = []
+        for obj_id in ids:
+            obj, _ = _get_obj(model, obj_id)
+            if obj:
```
mkurek, line 334: else -> error that object was not found

mkurek:
:ok:

PR #2573, opened 2016-07-13
Check values before running validator

mkurek:
272daa9f introduces a bug when errors are detected (diff should not be executed then)


ar4s:
:ok:

PR #2574, opened 2016-07-14
Add gateway field to network

mkurek:
Multiple networks (especially when they are not real networks, but artificial ones created to separate ip ranges in Ralph) could share the same gateway address, which wasn't possible before (one IP could be assigned to only one network).


ar4s:
:ok:

PR #2584, opened 2016-07-15
Fix generic foreign key prefetch related in polymorphic api

mkurek:
Bug: When there is more than one type on `BaseObject` API endpoint (`/api/base-objects/`), custom fields are returned only for objects of the first type

`CustomFieldValue` is model based on `ContentType`s, so it has `GenericForeignKey` to all other models. According to Django documentation for prefetch_related (https://docs.djangoproject.com/en/1.8/ref/models/querysets/#prefetch-related):
"It also supports prefetching of GenericRelation and GenericForeignKey, however, it must be restricted to a homogeneous set of results. For example, prefetching objects referenced by a GenericForeignKey is only supported if the query is restricted to one ContentType."

In case of `BaseObject`, there could be more than one type of object, so prefetch related was working only for the first of them (implicitly adding filter by content type to `CustomFieldValue` prefetch query), resulting in returning empty list of custom fields for objects of different type.

Since `ralph.api.utils.PolymorphicViewSetMixin` is injecting fields to be prefetched for every polymorphic descendant (in this case of BaseObject, ex. DataCenterAsset, BackOfficeAsset etc), the solution was to prefetch custom fields (and potentially any other field) in final viewset for the type (ex. `DataCenterAssetViewSet`), which is later injected to polymorphic (sub)query for objects of this type (instead of calling it once for all).

### DIFF
--- a/src/ralph/assets/api/views.py
+++ b/src/ralph/assets/api/views.py
@@ -76,21 +76,27 @@ class BaseObjectFilterSet(NetworkableObjectFilters):
         model = models.BaseObject
 
 
+base_object_descendant_prefetch_related = [
+    Prefetch('licences', queryset=BaseObjectLicence.objects.select_related(
+        *BaseObjectLicenceViewSet.select_related
+    )),
+    Prefetch(
+        'custom_fields',
+        queryset=CustomFieldValue.objects.select_related('custom_field')
+    ),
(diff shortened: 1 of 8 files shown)


ar4s:
:ok:

szok:
:ok:

PR #2586, opened 2016-07-15
Add hermes publisher to DC Asset, Virtualserver and Cluster

szok:


ar4s:
:ok:

mkurek:
new version in https://github.com/allegro/ralph/pull/2621

PR #2587, opened 2016-07-15
Network saving optimizations

mkurek:
- Don't save every other network when single network is saved (save only those on which currently saved network might have impact, especially when address has changed).
- Allow to enable not-saving potential subnetworks during Ralph2 sync


src/ralph/networks/models/networks.py
```diff
@@ -263,11 +263,22 @@ def size(self):
             return 2
         return self.max_ip - self.min_ip - 1
 
+    @property
+    def _address_has_changed(self):
```
vi4m: we should stick to some naming standard in case of boolean functions  /variables

I try to use primarly `is_.*` convention, in this case `has.*` -  `_has_address_changed` is  readable too, thus we can even remove this meaningless docstring

szok:
:ok:

PR #2588, opened 2016-07-18
Skip hostname checking for reserved IPs during syncing

mkurek:
- allow to skip hostname checking during saving reserved IP in network sync
- don't overwrite IP status to reserved when it's already exist


ar4s:
:ok:

PR #2590, opened 2016-07-18
Fix CHECK_IP_HOSTNAME_ON_SAVE default

mkurek:
53b9b140a43e100e0707708b6912ef9df27be901 allows to read `CHECK_IP_HOSTNAME_ON_SAVE` from env variable, but has changed default value from True to False.


ar4s:
:ok:

PR #2594, opened 2016-07-18
Fix top and bottom margin of network default values

mkurek:
Set bottom margin (counting from the beginning of the network) to 10 and top margin to 0.
Naming of these variables will be fixed in another PR.


ar4s:
:ok:

PR #2597, opened 2016-07-19
Exclude mgmt IP from deployment cleaning

mkurek:
Don't remove device's management IP during cleaning in deployment.


ar4s:
:ok:

PR #2601, opened 2016-07-19
Unification of mac address

mkurek:
- new MACAddressField storing mac in unified form (EUI-48) upper case,  separated by semicolon
- fix for host sync when IP has no ethernet


ar4s:
:ok:

PR #2606, opened 2016-07-21
Ralph2 sync: handle mgmt ip for cluster

mkurek:
- New class (mixin): WithManagementIPMixin - for handling mgmt ip/hostname in DataCenterAsset and Cluster


ar4s:
:ok:

PR #2608, opened 2016-07-21
Add ethernet and ips to cluster API

mkurek:
- New Cluster API fields: ethernet and ipaddresses
- Add `is_management` field to cluster ethernet form


ar4s:
:ok:

PR #2610, opened 2016-07-21
Fix masters for Cluster

mkurek:
- return proper type (BaseObject instead of BaseObjectCluster - broken in f703dc0bc053523ad11376844a7c69a2446a8ba7)
- refined prefetch_related for Cluster viewset to get final type of BaseObject


ar4s:
:ok:

PR #2615, opened 2016-07-22
Management IP assignment fix

mkurek:
Fix for management IP assignment when:
- IPAddress is reserved (but not connected with any BaseObject), or
- mgmt ip is changed to address that already exist in Ralph

### DIFF
--- a/src/ralph/data_center/models/mixins.py
+++ b/src/ralph/data_center/models/mixins.py
@@ -1,7 +1,24 @@
-from ralph.networks.models import IPAddress
+from django.core.exceptions import ValidationError
+
+from ralph.networks.models import IPAddress, IPAddressStatus
 
 
 class WithManagementIPMixin(object):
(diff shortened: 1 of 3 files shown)


src/ralph/data_center/models/mixins.py
```diff
@@ -10,11 +27,35 @@ def _get_management_ip(self):
...
+                except IPAddress.DoesNotExist:
+                    pass
+                else:
+                    # check if it's not assigned to any object
+                    if (
+                        ip.ethernet and
+                        ip.ethernet.base_object_id != self.pk
+                    ):
+                        raise ValidationError(
+                            'IP is already assigned to another object'
```
ar4s: can you add some information about assigned object (link or hostname)?

ar4s:
:ok:

ghost:
:ok:

PR #2616, opened 2016-07-22
Ralph3 sync: handle ips in subscriber

mkurek:
Handle IP addresses (and DHCPEntries) in subscibers for VirtualServer, StackedSwitch and Device


ar4s:
:ok:

PR #2618, opened 2016-07-25
Fix copy network behaviour

ar4s:
After change function change netmask to /1 which broke Ralph3 sync (temporarily creates network 0.0.0.0/1, which causes to assign every other network as subnetwork to it).


ghost:
:ok:

mkurek:
:ok:

PR #2622, opened 2016-07-26
Filter service by active status in API

mkurek:
Since DRF's boolean filter is broken (django-filters NullBooleanField is not API-friendly - it's working with forms widgets), Ralph have to define it's own boolean API filter.

Use 1, true, True or yes in `active` query param to get active services.
Use 0, false, False or no in `active` query param to get inactive services.


ar4s:
:ok:

PR #2624, opened 2016-07-26
Fix syncing IP addresses when mac exist

mkurek:
Fix for case when:
- Ethernet with particular mac exists and is attached to base object X
- IP assigned to this ethernet (mac) has changed

Previously Ralph throws an error (Duplicate entry 'XXX' for key 'ethernet_id'), now it handles this case correctly.

### DIFF
--- a/src/ralph/ralph2_sync/subscribers.py
+++ b/src/ralph/ralph2_sync/subscribers.py
@@ -159,22 +159,37 @@ def _handle_ips(obj, ips):
                 is_management=False
             )
         )
-        if created or not ip.ethernet:
-            mac = ip_dict['mac']
-            if mac is None:
-                ip.ethernet = Ethernet.objects.create(
-                    base_object=obj, mac=None
-                )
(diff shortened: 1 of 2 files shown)


src/ralph/ralph2_sync/subscribers.py
```diff
@@ -159,22 +159,37 @@ def _handle_ips(obj, ips):
...
+            # if it's already assigned to other ip address, clean it
+            try:
+                if eth.ipaddress != ip:
+                    if eth.ipaddress.status != IPAddressStatus.reserved:
+                        # delete ip if it's not reserved
+                        eth.ipaddress.delete()
+                    else:
+                        # otherwise clean it's ethernet assignment
+                        other_ip = eth.ipaddress
+                        other_ip.ethernet = None
```
ar4s, line 178: add `other_ip.dhcp_expose = False`

ar4s:
:ok:

PR #2625, opened 2016-07-26
Fix syncing m2m relations to Ralph2

mkurek:
Fix for syncing related objects (m2m) when saving in GUI. Now syncing is triggered after saving m2m relations.


ar4s:
:ok:

PR #2628, opened 2016-07-26
Network reserved ips as simple int field

mkurek:
POC of using plain int field to mark some ips as reserved (in network & dhcp) instead of creating instance of the ip with reserved status.

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -3,12 +3,10 @@ import ipaddress
 import logging
 import socket
 import struct
-from itertools import chain
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models, transaction
-from django.db.models import Q
 from django.db.models.signals import post_migrate
 from django.db.utils import ProgrammingError
 from django.dispatch import receiver
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -263,6 +263,20 @@ class Network(
         verbose_name=_('DNS servers'),
         blank=True,
     )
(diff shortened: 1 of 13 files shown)


src/ralph/ralph2_sync/subscribers.py
```diff
@@ -499,13 +500,13 @@ def sync_network_to_ralph3(data):
...
     net.vlan = data['vlan']
     net.dhcp_broadcast = data['dhcp_broadcast']
-    if data['reserved_ips']:
-        for address in data['reserved_ips']:
-            ip = IPAddress.objects.get_or_create(
-                address=address,
-                defaults=dict(status=IPAddressStatus.reserved)
-            )[0]
-            ip.save()  # trigger save to reassign to proper network
+    # if data['reserved_ips']:
```
ghost: could we delete this commented block?

src/ralph/networks/models/networks.py
```diff
@@ -314,6 +312,8 @@ def save(self, *args, **kwargs):
             >>> network.min_ip, network.max_ip, network.gateway
             (3232235776, 3232236031, None)
         """
+        # TODO: gateway status? reserved? or maybe regular ip field instead of
+        # PK?
         if self.gateway_id and not self.gateway.is_gateway:
```
ghost, line 362: `self.gateway.is_gateway`?

ar4s, line 362: but what if `self.gateway` will be `None`?

ghost:
:ok: to me maybe excluding the migration code which is untested :)

ar4s:
:ok:

szok:
:ok:

PR #2635, opened 2016-07-28
Integrate stocktaking view with reversion

Dragerus:
Stocktaking performed by user himself now adds useful comment
to history.

### DIFF
--- a/src/ralph/accounts/views.py
+++ b/src/ralph/accounts/views.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from datetime import date
 
+import reversion
 from django.conf import settings
 from django.contrib import messages
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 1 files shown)


src/ralph/accounts/views.py
```diff
@@ -58,30 +59,36 @@ def get_context_data(self, **kwargs):
 class InventoryTagView(View):
     http_method_names = ['post']
 
+    @staticmethod
+    def _add_tags(request, asset, tags):
+        asset.tags.add(*tags)
+        with reversion.create_revision():
+            asset.save()
+            reversion.set_user(request.user)
+            reversion.set_comment('Added tags {}'.format(', '.join(tags)))
```
ar4s: maybe ``Added inventory tags {}``? Then we have easy way to distinguish between normal adding and adding by self-inventory mechanism.

ar4s:
:ok:

mkurek:
:ok:

PR #2637, opened 2016-07-28
Fix not working with standard stocktaking

Dragerus:
Assets tagged by stocktaking staff were not marked as tagged in
my_equipment view.


src/ralph/accounts/admin.py
```diff
@@ -140,17 +140,20 @@ def confirm_ownership(self, item):
         region_stocktaking_enabled = item.user.regions.filter(
             stocktaking_enabled=True
         ).exists()
+        has_inv_tag = [n.startswith(settings.INVENTORY_TAG)
```
mkurek: use any (https://docs.python.org/3/library/functions.html#any)

src/ralph/back_office/models.py
```diff
@@ -50,6 +50,7 @@
 class Warehouse(NamedMixin, TimeStampMixin, models.Model):
     _allow_in_dashboard = True
     stocktaking_enabled = models.BooleanField(default=False)
+    stocktaking_tag_suffix = models.CharField(max_length=8, default='WHOUSE')
```
mkurek: why not empty default? i'd check if `stocktaking_tag_suffix` is set and only if it is i'd set this tag.

mkurek:
Could you make simple entry in docs about stocktaking in Ralph?

ar4s:
:ok:

src/ralph/back_office/migrations/0005_warehouse_stocktaking_tag_suffix.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('back_office', '0004_warehouse_stocktaking_enabled'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='warehouse',
+            name='stocktaking_tag_suffix',
+            field=models.CharField(default='WHOUSE', max_length=8),
```
mkurek: regenerate migrations (remove this and create new one) - this one still has default value `WHOUSE`

mkurek:
:ok:

ghost:
:ok:

PR #2639, opened 2016-07-29
Handle errors not defined by DNSForm

ghost:
DNSaaS can return errors in fields not defined in DNSForm.
Because of that, there is a need of handling such case.
This PR handles such case.
The PR also adds handler for insufficient permimssion requests (202 code).
This means Ralph sends requests with user defined in `settings.DNSAAS_OWNER` to DNSAAS
If DNSAAS_OWNER isn't `superadmin` in `DNSaaS` then 202 is returned.


mkurek:
:ok:

ar4s:
:ok:

PR #2644, opened 2016-07-29
Limit async transition to one at a time for object

mkurek:
For particular object, only one async transition could be running at once.


ghost:
:ok:

szok:
:ok:

PR #2648, opened 2016-08-01
Fix broken build.

vi4m:
Installing debian package was not possible, since
debian/raph-core.postinst script was not valid.
I added empty(:) instruction to 'if' clause to be parsed
correctly.


mkurek:
:ok:

ghost:
:ok:

PR #2650, opened 2016-08-01
Fix default settings to be sufficient to run app

vi4m:
RALPH_DEBUG=True will be env setting in deb default installation.
User can change it without writing config file. Default is False.

### DIFF
--- a/src/ralph/settings/prod.py
+++ b/src/ralph/settings/prod.py
@@ -3,7 +3,7 @@ import os
 
 from ralph.settings import *  # noqa
 
-DEBUG = os.environ.get('RALPH_DEBUG', False) 
+DEBUG = os_env_true('RALPH_DEBUG', False)
 
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'  # noqa
 
--- a/src/ralph/settings/prod.py
+++ b/src/ralph/settings/prod.py
@@ -3,8 +3,7 @@ import os
 
 from ralph.settings import *  # noqa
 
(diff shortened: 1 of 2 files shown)


mkurek:
i can't agree with this change. can't you use dev settings in deb? Or create new file for deb settings?

vi4m:
It would require to copy all the file contents. I think getenv() variable will be better?

mkurek:
you could use (in `ralph.settings.deb` or sth similar (deb vs dev typo)):
```
from ralph.settings.prod import *
DEBUG = True
```
if you want to use env variable name it other than `DEBUG` (ex. `RALPH_FORCE_DEBUG`)

vi4m:
Why not RALPH_DEBUG ?

mkurek:
`RALPH_DEBUG` is ok too - it just has to be variable which is not-so-easy (™) to overwrite (user has to set this value on purpose, not accidentally)

vi4m:
Ok, this will come with RALPH_DEBUG env setting with the safe default value.

src/ralph/settings/prod.py
```diff
@@ -3,8 +3,7 @@
 
 from ralph.settings import *  # noqa
 
-# FIXME: when going for full production, change it to False
-DEBUG = False
+DEBUG = os.environ.get('RALPH_DEBUG', False)
```
mkurek: use `os_env_true` (it's shortcut to properly handle boolean env variables)

szok:
:ok:

mkurek:
:ok:

PR #2654, opened 2016-08-02
Fix internal server error during creating user

mkurek:
Use `created=False` param in Ralph2SyncAdminMixin post_save trigger (it's not used in sync function anyway), to not trigger token creation twice.


szok:
:ok:

ar4s:
:ok:

PR #2656, opened 2016-08-02
Fix OperationalError on network when lft > rght

ar4s:
Network's view raised `OperationalError` (BIGINT UNSIGNED value is out of
range) when lft > rght.


ghost:
:ok:

mkurek:
:ok:

PR #2659, opened 2016-08-02
Fix for read bool from env in prod settings

mkurek:
- change `os_env_true` to `bool_from_env` in prod settings
- add travis test that checks if `ralph` cmd is working with test and prod settings


ar4s:
:ok:

ghost:
:ok:

PR #2663, opened 2016-08-03
Add post_send on ``device`` to ``AddVM``

ar4s:


szok:
:ok:

src/ralph/ui/views/deploy.py
```diff
@@ -503,6 +504,9 @@ def post(self, request, *args, **kwargs):
         resp['Location'] = LATEST_API.canonical_resource_for(
             'dev'
         ).get_resource_uri(device)
+        post_save.send(
```
mkurek, line 509: add `device._handle_post_save = False` in line 488

ar4s, line 509: I added `handle_post_save` to `create` method

mkurek:
:ok:

PR #2670, opened 2016-08-04
Add use hostname counter flag to NetworkEnvironment

szok:
- Add new flag use hostname counter to NetworkEnvironment
- Add next_hostname_without_model_counter method which return hostname based on already added hostname

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -81,6 +81,10 @@ class NetworkEnvironment(TimeStampMixin, NamedMixin, models.Model):
         blank=True,
         null=True,
     )
+    use_hostname_counter = models.BooleanField(
+        default=True,
+        help_text='If set to false hostname based on already added hostnames.'
+    )
 
     def __str__(self):
         return self.name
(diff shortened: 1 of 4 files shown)


src/ralph/networks/models/networks.py
```diff
@@ -81,6 +81,7 @@ class NetworkEnvironment(TimeStampMixin, NamedMixin, models.Model):
         blank=True,
         null=True,
     )
+    use_hostname_counter = models.BooleanField(default=True)
```
mkurek: add some help text here

src/ralph/networks/models/networks.py
```diff
@@ -93,21 +94,65 @@ def next_free_hostname(self):
...
+        if hostnames:
+            hostnames = sorted(hostnames, reverse=True)
+            counter = int(hostnames[0]) + 1
+        hostname = AssetLastHostname(
+            prefix=self.hostname_template_prefix,
+            counter=counter,
+            postfix=self.hostname_template_postfix
+        )
+        if create_last_hostname:
+            hostname.save()
```
mkurek: we don't need to save (create?) `AssetLastHostname` when `use_hostname_counter` is `False`

mkurek:
:ok:

ghost:
:ok:

PR #2672, opened 2016-08-04
[API] Embed full Manufacturer object in AssetModel.

xor-xor:


szok:
:ok:

mkurek:
:ok:

src/ralph/assets/api/views.py
```diff
@@ -81,6 +81,8 @@ class CategoryViewSet(RalphAPIViewSet):
 class AssetModelViewSet(RalphAPIViewSet):
     queryset = models.AssetModel.objects.all()
     serializer_class = serializers.AssetModelSerializer
+    save_serializer_class = serializers.AssetModelSaveSerializer
+    prefetch_related = ['manufacturer']
```
ar4s: `select_related` instead of `prefetch_related`

ar4s:
:ok:

PR #2674, opened 2016-08-04
Add permissions for proxy models

ar4s:


src/ralph/lib/permissions/models.py
```diff
@@ -288,3 +289,77 @@ def _get_objects_for_user(cls, user, queryset=None):
...
+
+    perms = [
+        Permission(codename=codename, name=name, content_type=ct)
+        for ct, (codename, name) in searched_perms
+        if (ct.pk, codename) not in all_perms
+    ]
+    # Validate the permissions before bulk_creation to avoid cryptic
+    # database error when the verbose_name is longer than 50 characters
+    permission_name_max_length = Permission._meta.get_field('name').max_length
+    verbose_name_max_length = permission_name_max_length - 11  # len('Can change ') prefix
```
xliiv: `11`?

ghost:
:ok:

mkurek:
:ok:

src/ralph/data_center/apps.py
```diff
@@ -0,0 +1,9 @@
+from ralph.apps import RalphAppConfig
+
+
+class DataCenterConfig(RalphAppConfig):
+    name = 'ralph.data_center'
+    verbose_name = 'Data Center Assets'
+
+    def get_proxies(self):
+        return ['DCHost']
```
szok: Perhaps a better solution would be to get all of model proxy and exclude class that have a parameter (exclude_permission?)? or classes that have a parameter (permission?)?

mkurek: `import this; print(this.s.split('\n')[3].translate(str.maketrans(this.d)))`

szok: Ok, so that every time I added the proxy model gonna have to remember, "hmm whether definitely added model to apps.py file" wtf?

get_proxies? what is this? 

Documentation?

Permissions have to work automatically for all ralph models/extra views, and so we give a special exception for which each time will have to keep in mind.

mkurek: ok. so @ar4s what do you think about making it working automatically for all proxy models?

We now have 7 proxy models:

```
<class 'ralph.operations.models.Change'>
<class 'ralph.operations.models.Incident'>
<class 'ralph.operations.models.Problem'>
<class 'ralph.operations.models.Failure'>
<class 'ralph.deployment.models.Deployment'>
<class 'ralph.data_center.models.hosts.DCHost'>
<class 'ralph.dhcp.models.DHCPEntry'>
```

ar4s: I choose this solution because i have a problem when i added permission for all proxy models - some pair (app_label, codename) in permission table was duplicated and then `sitetree` raised `MultipleObjectsReturned`.

szok: Propose tomorrow to discuss :)

szok:
:ok:

PR #2675, opened 2016-08-04
Fix DNS TXT records for virtual server

mkurek:
Call parent.last_descendant instead of parent to get proper instance type.


ghost:
:ok:

ar4s:
:ok:

PR #2676, opened 2016-08-04
Add index to object_id in to custom fields

ar4s:
Custom field values inline in details view of asset generate query with where
statement which contains object_id. Now is more efficient -- ~1ms vs ~20ms in environment.


mkurek:
Isn't only one commit in this PR appropriate?

ar4s:
@mkurek my mistake

mkurek:
:ok:

szok:
:ok:

PR #2678, opened 2016-08-04
Fix cleaning DNS entries during deployment when object has no ip addresses.

mkurek:
Currently, all records will be fetched from DNSaaS (and all could be potentially deleted -_-). This commit fix it (when there is no ip, no dns record is returned) and additionally limit max DNS entries to be removed in single deployment to 30 (by default).


ghost:
:ok:

szok:
:ok:

PR #2684, opened 2016-08-05
Add NetworkEnvironment to deployment wait for dhcp servers action

szok:


src/ralph/dhcp/models.py
```diff
@@ -35,6 +39,7 @@ class DHCPServer(models.Model):
     ip = models.GenericIPAddressField(
         verbose_name=_('IP address'), unique=True
     )
+    network_environment = models.ForeignKey(NetworkEnvironment, blank=True)
```
mkurek: what about null?

mkurek:
:ok:

ar4s:
:ok:

PR #2686, opened 2016-08-05
DHCP agent cache fix

mkurek:
- fix for cache set (use key (url) instead of value (last modified value))
- fix for different urls for get (with query params) and set (without query params)
- fix for python 3 (cast config to str when saving to file)


ar4s:
:ok:

szok:
:ok:

PR #2687, opened 2016-08-05
Fix dc hosts update signal to not be called twice when Ralph2 sync is on

mkurek:
To be removed when Ralph2 sync will be remove


szok:
:ok:

ar4s:
:ok:

PR #2688, opened 2016-08-05
Reduce render calls by fragment caching

ar4s:
Add cache to header menu and filter on list view thus reduce template calls
from 78 to 12 on Data Center Asset list view.


mkurek:
Is it using redis for caching? If yes, shouldn't we clear the cache with every release?

src/ralph/admin/templates/admin/base.html
```diff
@@ -25,31 +25,32 @@
...
-                        {% endblock %}
-                      </ul>
-                    </li>
-                  {% endblock %}
-                {% endif %}
-              {% endblock %}
-              {% block nav-global %}{% endblock %}
-            </ul>
-          </section>
+          {% cache 3600 header request.user request.path %}
```
mkurek: what if I change permissions for the user? do i have to wait an hour to see its effect?

ar4s: 10 minutes will be alright?

mkurek: we should think about same cache invalidation first :)

src/ralph/accounts/models.py
```diff
@@ -179,6 +182,13 @@ def save(self, *args, **kwargs):
     def autocomplete_str(self):
         return '{} <i>{}</i>'.format(str(self), self.department)
 
+    @property
+    def cache_param(self):
```
mkurek: consider renaming it to `permissions_hash`

szok:
:ok:

src/ralph/admin/templates/admin/base.html
```diff
@@ -25,34 +25,49 @@
...
-                        {% block userlinks %}
-                          {% include 'admin/partials/userlinks.html' %}
-                        {% endblock %}
-                      </ul>
-                    </li>
-                  {% endblock %}
-                {% endif %}
-              {% endblock %}
-              {% block nav-global %}{% endblock %}
-            </ul>
```
ghost, line 51: this `<ul>` is missing

ar4s, line 51: done

src/ralph/accounts/models.py
```diff
@@ -179,6 +182,16 @@ def save(self, *args, **kwargs):
     def autocomplete_str(self):
         return '{} <i>{}</i>'.format(str(self), self.department)
 
+    @property
+    def permissions_hash(self):
+        """
+        Property used in template as a param to cache invalidation.
+        Hash for caching is calculated from user ID and its permissions.
+        """
+        perms_set = sorted(self.get_all_permissions())
```
mkurek: so first of all, use ''.join(perms_set) instead of urlqote - it's over 100 times faster:

```
x = user.get_all_permissions()

def test1():
    return ''.join(x)

def test2():
    return urlquote(x)

>>> timeit.timeit('test1()', setup='from __main__ import test1', number=10000)
... 0.10609826701693237
>>> timeit.timeit('test2()', setup='from __main__ import test2', number=10000)
... 15.055276021943428
```

Consider using frozenset instead of sorting:

```
perms_set = frozenset(self.get_all_permissions())
key = '{}:{}'.format(self.id, hash(perms_set))
```

It will be much faster than sorting and dumping whole perms.

```
def test_perms1():
    perms_set = frozenset(user.get_all_permissions())
    key = '{}:{}'.format(user.id, hash(perms_set))
    return hashlib.md5(force_bytes(key)).hexdigest()

def test_perms2():
    perms_set = sorted(user.get_all_permissions())
    key = '{}:{}'.format(user.id, ''.join(perms_set))
    return hashlib.md5(force_bytes(key)).hexdigest()

>>> timeit.timeit('test_perms1()', setup='from __main__ import test_perms1', number=10000)
... 1.1444226970197633
>>> timeit.timeit('test_perms2()', setup='from __main__ import test_perms2', number=10000)
... 3.365936833899468
```

mkurek:
:ok:

PR #2689, opened 2016-08-05
DCHost API - add configuration_variables field

mkurek:
Add configuration_variables field to DCHosts api view.
Remove unnecessary prefetch_related and memory field.


szok:
:ok:

ar4s:
:ok:

PR #2690, opened 2016-08-05
Reduce API urls count

mkurek:
Reduce API urls count by almost halt (287 vs 158 now). We don't need URLs extensions with format (ex. /api/base-objects.json/).
See http://www.django-rest-framework.org/api-guide/format-suffixes/#format_suffix_patterns for details.


ar4s:
:ok:

szok:
:ok:

PR #2696, opened 2016-08-08
[API] Fix for 'isnull' filter + improvement for DateField/DateTimeField.

xor-xor:
This PR adds the following improvements/fixes to API filtering:
- extends filtering on `DateField` by adding `lte`, `gte`, `lt` and `gt` filters.
- extends filtering on `DateTimeField` by adding `year`, `month`, `day`, `week_day`, `range`, `isnull`, `lte`, `gte`, `lt`, and `gt` filters.
- fixes `isnull` filter by properly interpreting truthy values (`1`, `'1'`, `'true'`, `'True'`, `'yes'`) and falsey values (`0`, `'0'`, `'false'`, `'False'`, `'no'`).

### DIFF
--- a/src/ralph/api/filters.py
+++ b/src/ralph/api/filters.py
@@ -188,12 +188,13 @@ class LookupFilterBackend(BaseFilterBackend):
             ))
             if lookup in field_lookups:
                 if lookup == 'isnull':
-                    value_old = value
-                    value = BOOL_VALUES.get(value, value_old)
-                    if value == value_old:
+                    if value not in BOOL_VALUES:
                         logger.debug(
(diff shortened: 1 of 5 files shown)


szok:
:ok:

src/ralph/api/filters.py
```diff
@@ -183,6 +187,13 @@ def _validate_single_query_lookup(
                 model, model_field_name, field_lookups
             ))
             if lookup in field_lookups:
+                if lookup == 'isnull':
+                    value_old = value
```
mkurek: maybe you could use just `value in TRUE_VALUES`?

mkurek:
:ok:

PR #2702, opened 2016-08-09
Fix network environment calculation for DCAsset & VirtualServer

mkurek:
Don't rely only on rack in which server is installed - use ips assigned to server when calculating network env as well (assume, that server has non-management ips only in one environment).


ar4s:
:ok:

ghost:
:ok:

szok:
:ok:

PR #2703, opened 2016-08-09
Fix commiting async job to worker

mkurek:
When async transition (job) is scheduled to worker, transaction has to be already commited to allow worker to fetch job from database. This cannot be done in atomic block (especially when whole request is atomic), so from now transition view is non-atomic (gui + api) for async transitions and is commited manually after creating job object in database.

### DIFF
--- a/src/ralph/lib/transitions/models.py
+++ b/src/ralph/lib/transitions/models.py
@@ -659,7 +659,6 @@ def update_models_attrs(sender, **kwargs):
     """
     if sender.name == 'ralph.' + Transition._meta.app_label:
         for model, field_names in _transitions_fields.items():
-            ContentType.objects.get_for_model(model)
             content_type = ContentType.objects.get_for_model(model)
             for field_name in field_names:
                 transition_model, _ = TransitionModel.objects.get_or_create(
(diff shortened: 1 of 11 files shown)


src/ralph/lib/transitions/models.py
```diff
@@ -647,22 +653,22 @@ class TransitionJobAction(TimeStampMixin):
...
-        for field_name in field_names:
-            transition_model, _ = TransitionModel.objects.get_or_create(
-                content_type=content_type,
-                field_name=field_name
-            )
-            transition_models[field_name] = transition_model
-        setattr(model, 'transition_models', transition_models)
+    if sender.name == 'ralph.' + Transition._meta.app_label:
+        for model, field_names in _transitions_fields.items():
+            ContentType.objects.get_for_model(model)
```
szok: Delete this?

src/ralph/lib/transitions/tests/test_async.py
```diff
@@ -2,22 +2,24 @@
...
+from ralph.lib.transitions.tests import TransitionTestCaseMixin
 from ralph.tests.models import AsyncOrder, Foo, OrderStatus
 
 
-class AsyncTransitionsTest(TransitionTestCase):
+class AsyncTransitionsTest(TransitionTestCaseMixin, TransactionTestCase):
 
     def setUp(self):
         super().setUp()
+        # update_models_attrs()
```
szok: delete?

szok:
🆗

src/ralph/lib/transitions/models.py
mkurek: this what?

ghost:
:ok:

ar4s:
:ok:

PR #2706, opened 2016-08-10
Add private/public network filter

ghost:
New filter in `Network` tab allowing to filter networks by classes (like private, public, all).

### DIFF
--- a/src/ralph/networks/filters.py
+++ b/src/ralph/networks/filters.py
@@ -1,5 +1,6 @@
 import ipaddress
 
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.admin.filters import ChoicesListFilter, TextListFilter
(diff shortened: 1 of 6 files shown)


src/ralph/networks/filters.py
```diff
@@ -45,3 +50,26 @@ def queryset(self, request, queryset):
...
+    def queryset(self, request, queryset):
+        if not self.value():
+            return queryset
+
+        filter_ = Q()
+        for private_cidr in PRIVATE_NETWORK_CIDRS:
+            network = ipaddress.ip_network(private_cidr)
+            min_ip = int(network.network_address)
+            max_ip = int(network.broadcast_address)
+            filter_ |= Q(min_ip__gte=min_ip, max_ip__lte=max_ip)
```
mkurek: you could construct this query (filters) only once, during application loading (or maybe during first call to this filter). There is no need to parse this with each request.

mkurek:
:ok:

szok:
:ok:

PR #2708, opened 2016-08-10
Filter BaseObject by name extended for conf. class name and service name

mkurek:
- changed path field on ConfigurationClass from TextField to CharField
- improved some conf. class select and prefetch related


ghost:
:ok:

szok:
:ok:

PR #2710, opened 2016-08-10
Fix wrong cache key for user

ar4s:
TypeError exceptions muted in Django templates by default (see
https://github.com/django/django/blob/1.8.13/django/template/base.py#L854)

### DIFF
--- a/src/ralph/accounts/models.py
+++ b/src/ralph/accounts/models.py
@@ -188,7 +188,7 @@ class RalphUser(
         Hash for caching is calculated from user ID and its permissions.
         """
         perms_set = frozenset(self.get_all_permissions())
-        key = ':'.join((self.id, hash(perms_set)))
+        key = ':'.join((str(self.id), str(hash(perms_set))))
         return hashlib.md5(force_bytes(key)).hexdigest()
 
 
--- a/src/ralph/admin/templates/admin/base.html
+++ b/src/ralph/admin/templates/admin/base.html
@@ -25,37 +25,40 @@
               <a href="#"><span>Menu</span></a>
             </li>
           </ul>
-          {% cache 86400 header_menu opts.app_label opts.model_name request.user.permissions_hash %}
-          <!-- cache -->
-          {% spaceless %}
-            <section class="top-bar-section">
+          <section class="top-bar-section">
+            <!-- user cache key: {{ request.user.permissions_hash }} -->
+            {% cache 86400 header_menu opts.app_label opts.model_name request.user.permissions_hash %}
+            {% spaceless %}
+            <!-- cache -->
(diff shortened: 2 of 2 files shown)


src/ralph/accounts/models.py
```diff
@@ -188,7 +188,7 @@ def permissions_hash(self):
         Hash for caching is calculated from user ID and its permissions.
         """
         perms_set = frozenset(self.get_all_permissions())
-        key = ':'.join((self.id, hash(perms_set)))
+        key = ':'.join((str(self.id), str(hash(perms_set))))
```
mkurek, line 191: so how was it working before? :)

src/ralph/admin/templates/admin/base.html
```diff
@@ -25,37 +25,40 @@
               <a href="#"><span>Menu</span></a>
             </li>
           </ul>
-          {% cache 86400 header_menu opts.app_label opts.model_name request.user.permissions_hash %}
-          <!-- cache -->
-          {% spaceless %}
-            <section class="top-bar-section">
+          <section class="top-bar-section">
+            <!-- user cache key: {{ request.user.permissions_hash }} -->
```
mkurek, line 29: so maybe `permissions_hash` should be `cached_property`? :)

src/ralph/admin/templates/admin/base.html
```diff
@@ -25,37 +25,40 @@
...
-                  {% endif %}
-                {% endblock %}
-                {% block nav-global %}{% endblock %}
-              </ul>
-            </section>
-          {% endspaceless %}
-          <!-- end cache -->
-          {% endcache %}
+              {% endif %}
+              {% block usertools %}
```
mkurek, line 43: why we're not caching it anymore?

src/ralph/accounts/models.py
ar4s, line 191: Ha! it was one from django's gotchas - if function/property raise `KeyError` then django mute this exception by default and returns nothing to template's context therefore our cache key for doesn't work properly - all of users saw same menu

src/ralph/admin/templates/admin/base.html
ar4s, line 43: IMO we don't need because the benefits of caching are negligible but if something goes wrong with cache key (duplicates, empty) then user see own name in the right corner and then user not panic..
BTW `contextual_search_form` was cached which is wrong

src/ralph/accounts/models.py
mkurek: Hmm I've just tested it and it seems to work on two different users. Maybe change of permissions will not work here (because empty string is injected as hash of permissions every time)?

mkurek:
:ok:

szok:
:ok:

ghost:
:ok:

PR #2711, opened 2016-08-10
Add ``string_if_invalid`` to templates settings

ar4s:
To unmute excpetion if some variable/property/function not found or raise
`TypeError`.


src/ralph/settings/dev.py
```diff
@@ -40,6 +40,7 @@ def only_true(request):
                 'django.template.loaders.app_directories.Loader',
                 'ralph.lib.template.loaders.AppTemplateLoader',
             ],
+            'string_if_invalid': TEMPLATE_STRING_IF_INVALID
```
mkurek, line 43: btw what's the difference between base TEMPLATES and dev TEMPLATES?

ghost:
What's the idea behind this change?
This should be disabled in production, according to this
https://docs.djangoproject.com/en/1.9/ref/templates/api/#how-invalid-variables-are-handled

src/ralph/settings/test.py
```diff
@@ -68,6 +68,8 @@
     }
 }
 
+TEMPLATES[0]['OPTIONS']['string_if_invalid'] = TEMPLATE_STRING_IF_INVALID
```
vi4m, line 71: where is the comment? what's that?

PR #2718, opened 2016-08-12
IP validation during deployment

ghost:
Raise validation error when user puts IP not yet added to `networks`.

### DIFF
--- a/src/ralph/deployment/deployment.py
+++ b/src/ralph/deployment/deployment.py
@@ -18,6 +18,7 @@ Deployment order:
 * Wait for ping
 
 """
+import ipaddress
 import logging
 import time
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


mkurek:
:ok:

szok:
:ok:

src/ralph/deployment/deployment.py
```diff
@@ -397,23 +398,41 @@ def _assign_hostname(instance, new_hostname, net_env=None):
...
+            )
+
+
+def check_ip_from_defined_network(address):
+    """
+    Validate if `address` belongs to any network already added
+    """
+    ip = ipaddress.ip_address(address)
+    try:
+        Network.objects.get(min_ip__lte=int(ip), max_ip__gte=int(ip))
```
mkurek: there could be many networks to which this ip will fit - we should use filter here (but DoesNotExist will not be raised then)

mkurek:
:ok:

PR #2719, opened 2016-08-12
Ralph3 sync: fix vm ack and delete vm

mkurek:
- Fix sending ACK to Ralph3 for VirtualServer (send model=VirtualServer instead of Device)
- new handler for deleting virtual server


ghost:
:ok:

szok:
:ok:

PR #2720, opened 2016-08-12
Ralph2 sync: delete virtual server

mkurek:


szok:
:ok:

src/ralph/ralph2_sync/publishers.py
```diff
@@ -330,3 +330,14 @@ def sync_network_to_ralph2(sender, instance=None, created=False, **kwargs):
...
+
+@ralph2_sync(VirtualServer, signal=post_delete)
+def delete_virtual_server_in_ralph2(sender, instance=None, created=False, **kwargs):  # noqa
+    ralph2_id = _get_obj_id_ralph_2(instance)
+    if ralph2_id:
+        data = {
+            'id': instance.id,
+            'ralph2_id': ralph2_id,
+        }
+        return data
```
ghost, line 343: This return non-dicts. Is this ok?

mkurek, line 343: yes - only castable-to-True result will be published. It doesn't matter if it's None or {} - it won't be published

ghost:
:ok:

PR #2724, opened 2016-08-16
Filter in API by id__startswith

mkurek:
New lookup filter for AutoField which allows to filter by startswith and exact lookups.


szok:
:ok:

ar4s:
:ok:

PR #2726, opened 2016-08-16
Fix available networks when rack is not assigned

mkurek:
Fix for list of available networks when rack is not assigned (either directly to physical host, or indirectly, through vm's hypervisor).


ar4s:
:ok:

szok:
:ok:

ghost:
:ok:

PR #2727, opened 2016-08-16
Prefetch related models for transition when checking for available actions for status

mkurek:
Reduce sql queries on detail view from N*3+1 (where N is number of transitions for object, ex. 16 for 5 transitions) to only 2.


ar4s:
:ok:

szok:
:ok:

PR #2729, opened 2016-08-17
Fix preboot variables tooltip formatting

mkurek:
Use {{{{ instead of {{ in configuration field help text ({{ was parsed by str.format).


szok:
:ok:

ghost:
:ok:

PR #2734, opened 2016-08-17
Fix network env validation during deployment

mkurek:
Check if instance (DCAsset, vm) has any available network environment, not if it currently has one.


ar4s:
:ok:

ghost:
:ok:

szok:
:ok:

PR #2735, opened 2016-08-17
Make dropdowns in top bar disappear after delay

4i60r:


ar4s:
:ok:

src/ralph/admin/templates/admin/includes/javascripts.html
```diff
@@ -98,5 +98,26 @@
...
+            var dropdowns = parent.find("li.has-dropdown .dropdown");
+            dropdowns.each(function() {
+                var $drop = $(this);
+                var width = 'auto';
+                if(!$drop.is($thisdrop)) {
+                    width = '';
+                    $drop.hide();
+                    setTimeout(function() {
+                        $drop.show();
+                    }, 300);
```
mkurek: isn't 300ms too much?

mkurek: btw, could we do sth like this: if the cursor is further from dropdown then visibility time decreases? :)

src/ralph/static/src/scss/_overrides.scss
```diff
@@ -85,3 +85,23 @@ select {
...
     overflow-x: auto;
 }
+
+.top-bar {
+  .top-bar-section {
+    li.has-dropdown {
+
+      &.hover, &.not-click {
+        > .dropdown {
+          transition: clip 0s 1s linear;
```
mkurek: please change it to some smaller value, ex. 0.5s.

mkurek:
make it mergable.

szok:
🆗

mkurek:
:ok:

PR #2738, opened 2016-08-18
Add filters to ipaddress, virtual, clusters

szok:
- Add location column to All hosts
- Add configuration_path filters to Hardware
- Add ip address filter to hardware
- Add ip hostname filter to Virtual, Cluster, Allhosts
- Add hostname filter to ipaddress


mkurek:
:ok:

ar4s:
:ok:

PR #2739, opened 2016-08-18
Add search box to change views

ar4s:


src/ralph/admin/mixins.py
```diff
@@ -57,6 +58,36 @@ def get_inline_media():
     )
 
 
+def initialize_search_form(model, context, fields_from_model=True):
+    """
+    Add to template's context extra variables (search_fields,
```
mkurek: -> Add extra variables (search_fields, search_url) which are used by search form to template's context.

src/ralph/admin/mixins.py
```diff
@@ -57,6 +58,36 @@ def get_inline_media():
...
+    Args:
+        context (dict): context from view
+        fields_from_model (bool): if True (by default) then fetching
+            field name from database model, otherwise fetching from
+            admin model (search_field)
+    """
+    model_admin = ralph_site._registry[model]
+    verbose_search_fields = []
+    admin_search_fields = model_admin.search_fields
+    if fields_from_model:
```
mkurek: when is this option used right now? why do we need two different ways to construct placeholder?

mkurek:
:ok:

szok:
:ok:

PR #2742, opened 2016-08-19
Send ip addresses instead of hostname

ghost:
Changes on `Data Center Asset`, `Virtual Server`, `Cluster` publish data.
Those data is used to create auto txt records in DNSaaS.
Now the data includes `ip addresses` instead of `hostname`.

### DIFF
--- a/src/ralph/assets/utils.py
+++ b/src/ralph/assets/utils.py
@@ -21,7 +21,7 @@ class DNSaaSPublisherMixin:
             if not purpose:
                 continue
             data.append({
-                'name': self.hostname or '',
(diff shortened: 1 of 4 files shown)


src/ralph/assets/utils.py
```diff
@@ -21,7 +21,7 @@ def get_auto_txt_data(self):
             if not purpose:
                 continue
             data.append({
-                'name': self.hostname or '',
+                'ips': [ip.address for ip in self.ipaddresses],
```
mkurek: exclude management ip here

szok:
exclude mgmt IP and :ok: for me :)

ar4s:
:ok:

PR #2745, opened 2016-08-19
Add possibility to turn off redirection to object detail if only one search result

mkurek:
New settings variable: REDIRECT_TO_DETAIL_VIEW_IF_ONE_SEARCH_RESULT


szok:
:ok:

ar4s:
:ok:

PR #2746, opened 2016-08-19
Fix dc-hosts queries count

mkurek:
Location field was injected without proper select_related.

From now:
- location is shown for VirtualServer and CloudHost as well
- location is no longer "clickable" (it pointed to dcasset view with some filter) - it might be some TODO after this PR to create rack/server_room/dc filter for DCHost
- there is no redundant (n+1) sql on this view


szok:
:ok:

ar4s:
:ok:

PR #2752, opened 2016-08-23
Add hostname from IP address to search and filters

ar4s:


src/ralph/admin/filters.py
```diff
@@ -535,6 +535,32 @@ def queryset(self, request, queryset):
...
+    title = _('Hostname')
+    parameter_name = 'hostname'
+    template = 'admin/filters/text_filter.html'
+
+    def queryset(self, request, queryset):
+        if not self.value():
+            return queryset
+        queries = [
+            Q(hostname__icontains=self.value()) |
+            Q(ethernet_set__ipaddress__hostname__icontains=self.value())
```
mkurek: could you add this condition (for hostname of ip) to DCHostHostnameFilter https://github.com/allegro/ralph/blob/ng/src/ralph/data_center/admin.py#L79 ?

mkurek:
:ok:

mkurek:
:ok:

szok:
:ok:

PR #2754, opened 2016-08-23
Filter by env name in API 

mkurek:
Filter by env:
- base-objects
- dc-hosts
- data-center-assets
- virtual-servers
- cloud-hosts


szok:
:ok:

ar4s:
:ok:

PR #2760, opened 2016-08-23
Bumped django-mptt version to 0.8.6

mkurek:
Newest version of mptt provides fix for read only listing of mptt models.


ar4s:
:ok:

szok:
:ok:

PR #2761, opened 2016-08-23
Improved cloudhost-list api SQL queries count

mkurek:
Reduced SQL queries count (N+1) in cloudhost-list API view (fixed wrong key for checking prefetched objects).


ar4s:
:ok:

szok:
:ok:

PR #2776, opened 2016-08-30
Add signal which send data with old and new value of address and host…

ar4s:
Related to https://github.com/allegro/django-powerdns-dnssec/pull/146

### DIFF
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -245,4 +245,6 @@ class DNSaaS:
         logger.info('Send update data: {}'.format(ip_record_data))
         url = self.build_url('ip_record')
         response = self._post(url, ip_record_data)
+        if response.status_code >= 400:
+            logger.info('DNSaaS returned {}'.format(response.status_code))
         return response.json()
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -215,3 +215,7 @@ class DNSaaS:
             }
         elif request.status_code != 204:
             return request.json()
+
+    def send_ipaddress_data(self, data):
+        # TODO: send data to dedicated endpoint
+        logger.warning('Not implemented yet')
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -200,7 +200,7 @@ class DNSaaS:
 
     def delete_dns_record(self, record_id):
         """
-        Delete record in DNSAAS
+        Delete record in DNSAAS.
 
         Args:
             record_ids: ID's to delete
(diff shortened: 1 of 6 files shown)


src/ralph/lib/mixins/models.py
```diff
@@ -102,3 +102,15 @@ class TaggableMixin(models.Model):
 
     class Meta:
         abstract = True
+
+
+class PreviousStateMixin(models.Model):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        fields = self._meta.get_all_field_names()
```
ghost: https://docs.djangoproject.com/en/1.9/ref/models/meta/#migrating-from-the-old-api

ghost: [f.name for f in MyModel._meta.get_fields()]

src/ralph/networks/receivers.py
```diff
@@ -0,0 +1,23 @@
...
+    data_to_send = {
+        'new': {
+            key: instance.__dict__[key] for key in keys
+        },
+        'old': {
+            key: instance._previous_state[key] for key in keys
+        }
+    }
+    data_to_send['action'] = 'add' if created else 'update'
+    DNSaaS().send_ipaddress_data(data_to_send)
```
mkurek: do we want to send "event" even when hostname is empty?

src/ralph/dns/dnsaas.py
```diff
@@ -216,3 +216,8 @@ def delete_dns_record(self, record_id):
             }
         elif response.status_code != 204:
             return response.json()
+
+    def send_ipaddress_data(self, update_data):
```
ghost: Perhaps some info about `update_data` could be added.

ghost:
:ok:

mkurek:
could you write some tests (mocking sending data to dnsaas (but it'd perfect to have integration tests for that in the future))? ex. for adding new ip, updating ip, deleting ip, deleting server (ex. DataCenterAsset or VirtualServer)

src/ralph/dns/dnsaas.py
```diff
@@ -216,3 +216,35 @@ def delete_dns_record(self, record_id):
...
+                ``old`` and ``new`` contains previous and next state of
+                IP address record (in Ralph).
+        Returns:
+            JSON response from API
+        """
+        logger.info('Send update data: {}'.format(ip_record_data))
+        url = self.build_url('ip_record')
+        response = self._post(url, ip_record_data)
+        if response.status_code >= 400:
+            logger.info('DNSaaS returned {}'.format(response.status_code))
```
szok: maybe 'warning' or 'error' should be better?

szok (approved):

PR #2777, opened 2016-08-31
Change default content-type for DNSaaS REST client

ar4s:
Now by default is `application/json`.
Moreover, according to RFC2616 HTTP/1.1 message SHOULD include `Content-Type`.

### DIFF
--- a/src/ralph/dns/dnsaas.py
+++ b/src/ralph/dns/dnsaas.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import json
 import logging
 from urllib.parse import urlencode, urljoin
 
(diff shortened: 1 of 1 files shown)


src/ralph/dns/dnsaas.py
```diff
@@ -194,7 +196,7 @@ def create_dns_record(self, record):
         return self._post(url, data)
 
     def _post(self, url, data):
-        response = self.session.post(url, data=data)
+        response = self.session.post(url, data=json.dumps(data))
```
mkurek: use `json` param: http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests

mkurek:
:ok:

szok:
:ok:

PR #2779, opened 2016-08-31
Add components for virtual server

mkurek:
- components tab for virtual server
- reorganization of components
- new Ethernet components on view
- remove speed field from networks ethernet inline

Besides components there is new virtualserver api filter: hypervisor_service


ar4s:
:ok:

szok:
:ok:

PR #2781, opened 2016-09-01
[WIP] Add button on rack

ar4s:


mkurek:
could we show it only in "edit" mode?

PR #2783, opened 2016-09-01
Add subscriptions for VIP-related events from Hermes.

xor-xor:
In this PR, only `createVipEvent` and `deleteVipEvent` are fully handled - the routine for `updateVipEvent` is just a stub (such event doesn't carry anything useful right now) and `refreshVipEvent` has no use for Ralph at all (it will be used by `ralph_pricing` ("Scrooge") instead).


src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,30 @@ def __str__(self):
         return 'Database: {}'.format(self.service_env)
 
 
+class VIPProtocol(Choices):
+    _ = Choices.Choice
+
+    unknown = _('unknown')
+    TCP = _('TCP')
+    HTTP = _('HTTP')
```
mkurek: IMO its UDP :)

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,30 @@ def __str__(self):
...
+
+
 class VIP(AdminAbsoluteUrlMixin, BaseObject):
     class Meta:
```
mkurek, line 47: put meta and methods below fields.

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,30 @@ def __str__(self):
...
         verbose_name = _('VIP')
         verbose_name_plural = _('VIPs')
+        unique_together = ('ip', 'port', 'protocol')
 
     def __str__(self):
-        return 'VIP: {}'.format(self.service_env)
+        return self.name
+
+    name = models.CharField(_('name'), max_length=255, blank=True, null=True)
+    ip = models.ForeignKey(IPAddress, blank=True, null=True, default=None)
```
mkurek: how could this be blank or null? doesn't every vip has defined ip (and port and protocol)?

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,30 @@ def __str__(self):
...
     class Meta:
         verbose_name = _('VIP')
         verbose_name_plural = _('VIPs')
+        unique_together = ('ip', 'port', 'protocol')
 
     def __str__(self):
-        return 'VIP: {}'.format(self.service_env)
+        return self.name
+
+    name = models.CharField(_('name'), max_length=255, blank=True, null=True)
```
mkurek: this shouldn't be nullable either

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,30 @@ def __str__(self):
         return 'Database: {}'.format(self.service_env)
 
 
+class VIPProtocol(Choices):
+    _ = Choices.Choice
+
+    unknown = _('unknown')
```
mkurek: i'd remove this option

src/ralph/data_center/tests/factories.py
```diff
@@ -224,7 +230,13 @@ class Meta:
 
 
 class VIPFactory(DjangoModelFactory):
+    name = factory.Sequence(lambda n: 'ralph-test{}.local'.format(n))
+    # IPAddressFactory is given as string to avoid circular imports here.
+    ip = factory.SubFactory('ralph.networks.tests.factories.IPAddressFactory')
+    port = FuzzyInteger(1024, 49151)
+    protocol = factory.Iterator([VIPProtocol.TCP.id, VIPProtocol.HTTP.id])
     service_env = factory.SubFactory(ServiceEnvironmentFactory)
+    # XXX @mkurek - do we need 'parent' field here as well..?
```
mkurek: create `VIPFullFactory`, inheriting from VIPFactory, in which you'll make parent field as a SubFactory to `Cluster`

src/ralph/data_center/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from ralph.apps import RalphAppConfig
+
+
+class DataCenter(RalphAppConfig):
```
mkurek: rename it to `DataCenterConfig`

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+from django.core.exceptions import ValidationError
+
+from ralph.assets.models.assets import ServiceEnvironment
+from ralph.data_center.models import Cluster, ClusterType, VIP, VIPProtocol
+from ralph.networks.models.networks import IPAddress
+
+logger = logging.getLogger(__name__)
+
+
+def validate_event_data(data):
```
mkurek: rename it to `validate_vip_event_data`

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+    if not environment:
+        err = 'missing environment'
+        errors.append(err)
+    return errors
+
+
+def get_vip(ip, port, protocol):
+    try:
+        # There should be *at most* one such VIP.
+        vip = VIP.objects.get(ip=ip, port=port, protocol=protocol)
```
mkurek: you could return directly here

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+    if errors:
+        msg = (
+            'Error(s) detected in event data: {}. Ignoring received create '
+            'event.'
+        )
+        logger.error(msg.format('; '.join(errors)))
+        return
+
+    # Check if VIP already exists.
+    ip, _ = IPAddress.objects.get_or_create(address=data['ip'])
```
mkurek: this ip should be assigned to cluster (through ethernets)

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+    protocol = getattr(
```
mkurek: use `VIPProtocol.from_name(data['protocol'].upper())`

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+    logger.debug('VIP {} deleted successfully.'.format(vip.name))
+
+    # Delete IP address associated with it, but only when no other VIP uses it.
+    if not VIP.objects.filter(ip=ip).exists():
+        ip.delete()
+        msg = (
+            'IP address {} has been deleted since it is no longer being used '
+            'by any VIP.'
+        )
+        logger.info(msg.format(ip.address))
```
mkurek: info -> warning

src/ralph/data_center/subscribers.py
```diff
@@ -0,0 +1,182 @@
...
+        VIPProtocol, data['protocol'].upper(), VIPProtocol.unknown
+    )
+    vip = get_vip(ip, data['port'], protocol)
+    if vip is None:
+        msg = ("VIP designated by IP address {}, port {} and protocol {} "
+               "doesn't exist. Ignoring received delete event.")
+        logger.warning(msg.format(ip.address, data['port'], protocol.name))
+        return
+    vip.delete()
+    logger.debug('VIP {} deleted successfully.'.format(vip.name))
```
mkurek: debug -> info

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,28 @@ def __str__(self):
...
+    )
+
     class Meta:
         verbose_name = _('VIP')
         verbose_name_plural = _('VIPs')
+        unique_together = ('ip', 'port', 'protocol')
 
     def __str__(self):
-        return 'VIP: {}'.format(self.service_env)
+        return self.name
```
xor-xor: @mkurek - I'm just wondering, how VIPs should be represented, i.e. is `self.name` enough for that, or maybe we should go with "full" representation (IP + port + protocol), since names alone are not unique here..?

mkurek: :+1: for full representation

src/ralph/data_center/tests/test_subscribers.py
```diff
@@ -0,0 +1,217 @@
...
+        self.assertEqual(VIP.objects.count(), 1)
+
+    def test_create_with_invalid_event_data(self):
+        self.data['service'] = None
+        self.assertEqual(VIP.objects.count(), 0)
+        handle_create_vip_event(self.data)
+        self.assertEqual(VIP.objects.count(), 0)
+
+    def tearDown(self):
+        logging.disable(logging.NOTSET)
```
mkurek: do we need this logging-level-setup in these test cases? i'd prefer to manipulate them directly in settings (https://github.com/allegro/ralph/blob/ng/src/ralph/settings/test.py#L49)

xor-xor: So what about disabling all loggers during tests, on all levels? Are you OK with that..?

mkurek: yes

szok:
:ok:

src/ralph/data_center/models/virtual.py
```diff
@@ -26,13 +27,32 @@ def __str__(self):
...
+    TCP = _('TCP')
+    UDP = _('UDP')
+
+
 class VIP(AdminAbsoluteUrlMixin, BaseObject):
+    name = models.CharField(_('name'), max_length=255)
+    ip = models.ForeignKey(IPAddress)
+    port = models.PositiveIntegerField(verbose_name=_('port'), default=0)
+    protocol = models.PositiveIntegerField(
+        verbose_name=_('protocol'), choices=VIPProtocol(),
```
mkurek: add default (TCP)

mkurek:
please add this default and it's :ok: for me

PR #2788, opened 2016-09-07
Add new fields to Domains: website_type, website_url

ghost:


szok:
:ok:

src/ralph/domains/models/domains.py
```diff
@@ -70,10 +79,41 @@ class Domain(BaseObject, AdminAbsoluteUrlMixin):
         null=True,
         help_text=_("Company which receives invoice for the domain")
     )
+    website_type = models.PositiveIntegerField(
+        default=WebsiteType.none.id,
```
mkurek: shouldn't direct be default choice?

ar4s:
:ok:

PR #2794, opened 2016-09-09
Show vertical scroll on bulk edit view

ar4s:


szok:
:ok:

src/ralph/static/src/scss/components/bulk_edit.scss
```diff
@@ -1,5 +1,8 @@
-.bulk-edit {
+.bulk-edit #content-main {
     overflow-x: auto;
+    .with-tabs {
+        margin-bottom: 60px;
```
mkurek: isn't it height of footer? could we store it in some variable (constant)?

ghost:
:ok:

PR #2805, opened 2016-09-13
Change Domain import export

ghost:
Change fk related fields from pk to name (which are unique).
Fields:
- DomainContractResource.domain
- DomainResource.BusinessSegment


szok:
:ok:

ar4s:
:ok:

PR #2824, opened 2016-09-29
Check next free IP in DNSaaS (if integration is enabled)

ar4s:
- [x] tests

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -512,20 +512,11 @@ class Network(
         free_ip_as_int = None
         for free_ip_as_int in range(min_ip, max_ip + 1):
             if free_ip_as_int not in used_ips:
-                break
-        if free_ip_as_int is None:
-            return None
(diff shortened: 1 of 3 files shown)


src/ralph/networks/models/networks.py
```diff
@@ -503,11 +513,14 @@ def get_first_free_ip(self):
         for free_ip_as_int in range(min_ip, max_ip + 1):
             if free_ip_as_int not in used_ips:
                 break
-        # TODO: do it better
-        last = free_ip_as_int in used_ips
+        next_free_ip = ipaddress.ip_address(free_ip_as_int)
+        while is_in_dnsaas(next_free_ip) and free_ip_as_int <= max_ip:
+            logger.error('IP {} is already in DNS'.format(next_free_ip))
+            free_ip_as_int += 1
```
mkurek: we should check now if `free_ip_as_int` is not in `used_ips` (just like in line 504/514)

mkurek (changes requested):

szok (approved):

PR #2826, opened 2016-09-30
Change logger error in dnsaas integration

szok:


mkurek (approved):

ghost (approved):

src/ralph/dns/dnsaas.py
```diff
@@ -154,7 +154,7 @@ def _response2result(self, response):
             )
             return {
                 'non_field_errors': [
-                    _("Your request couldn't be handled, try later.")
+                    "Your request couldn't be handled, try later."
                 ]
             }
         elif response.status_code != 201:
```
xliiv: 404 should be handled because 404 case cant be jsonized

PR #2832, opened 2016-10-05
Add new settings ``RACK_LISTING_NUMBERING_TOP_TO_BOTTOM``

ar4s:
Set `RACK_LISTING_NUMBERING_TOP_TO_BOTTOM = True` for reverse number ordering in rack
visualization.

```
RACK_LISTING_NUMBERING_TOP_TO_BOTTOM = True   # 1..max_u_height from top to bottom
RACK_LISTING_NUMBERING_TOP_TO_BOTTOM = False  # max_u_height..1 from top to bottom
```

image


ar4s:
image

src/ralph/dc_view/templates/dc_view/settings_for_angular.js
```diff
@@ -0,0 +1,9 @@
+{% load cache staticfiles %}
+
+{% cache 3600 settings_js %}
```
mkurek: Do we have to cache it?

ar4s: nope :)

ghost (approved):

src/ralph/settings/base.py
```diff
@@ -339,6 +339,12 @@ def str_to_bool(s: str) -> bool:
 }
 
 # =============================================================================
+# DC view
+# =============================================================================
+
+RACK_LISTING_REVERSE = True
```
mkurek: default should be False. And consider changing name of this variable to better show numeration (from top to bottom or from bottom to top).

mkurek (changes requested):

src/ralph/settings/base.py
ar4s: done

mkurek (approved):

PR #2845, opened 2016-10-20
Add new transition actions: Assign new IP and Replace IP

szok:


src/ralph/deployment/deployment.py
```diff
@@ -386,6 +386,7 @@ def _assign_hostname(instance, new_hostname, net_env=None):
         kwargs['history_kwargs'][instance.pk]['hostname'] = '{}{}'.format(
             new_hostname, ' (from {})'.format(net_env) if net_env else ''
         )
+        kwargs['shared_params']['hostnames'][instances[0].pk] = new_hostname
```
mkurek: `instances[0].pk` -> `instance.pk`

src/ralph/deployment/deployment.py
```diff
@@ -438,6 +439,92 @@ def check_ip_from_defined_network(address):
...
+            network = Network.objects.get(
+                pk=ip_or_network['value']
+            )
+            ip = network.issue_next_free_ip()
+        logger.info('Assigning {} to {}'.format(ip, instance))
+        ethernet = Ethernet.objects.create(base_object=instance)
+        logger.info('Bounding {} to {} ethernet'.format(ip, ethernet))
+        ip.ethernet = ethernet
+
+        hostname = kwargs['shared_params']['hostnames'].get(instance.pk)
```
mkurek: what if there is no `hostnames` in `kwargs['shared_params']`?

src/ralph/deployment/deployment.py
```diff
@@ -438,6 +439,92 @@ def check_ip_from_defined_network(address):
...
+                [network.id, network]
+            )
+    return networks
+
+
+@deployment_action(
+    verbose_name=_('Replace IP'),
+    form_fields={
+        'ipaddress': {
+            'field': ChoiceFieldWithOtherOption(
```
mkurek: it should be basic choice field (we don't allow other value here - we should choose one of current IPs of this base object)

src/ralph/deployment/deployment.py
```diff
@@ -438,6 +439,92 @@ def check_ip_from_defined_network(address):
...
+                label=_('IP Address'),
+                auto_other_choice=False,
+            ),
+            'choices': base_object_ip_choices,
+            'exclude_from_history': True,
+        },
+        'network': {
+            'field': ChoiceFieldWithOtherOption(
```
mkurek: the same here - regular choice field instead of choice with other option

mkurek (changes requested):

src/ralph/deployment/deployment.py
```diff
@@ -438,15 +459,93 @@ def check_ip_from_defined_network(address):
...
+            ),
+            'choices': base_object_network_choices,
+            'exclude_from_history': True,
+        },
+    }
+)
+def replace_ip(cls, instances, ipaddress, network, **kwargs):
+    ip = IPAddress.objects.get(pk=ipaddress)
+    network = Network.objects.get(pk=network)
+    ip.address = str(network.get_first_free_ip())
```
mkurek: some logging will be useful here (ex. 'Replacing <OLD-IP> to <NEW-IP> for <HOSTNAME>')

mkurek (changes requested):

src/ralph/deployment/deployment.py
```diff
@@ -438,15 +459,97 @@ def check_ip_from_defined_network(address):
...
+                ip.hostname = hostname
+        except KeyError:
+            pass
+
+        ip.save()
+
+
+def base_object_ip_choices(actions, objects):
+    ipaddresses = []
+    for instance in objects:
```
mkurek, line 495: one more thing - we should forbid running this action (Replace IP) with more than one object.

mkurek (changes requested):

src/ralph/deployment/deployment.py
szok: Done

mkurek (approved):

ghost (approved):

PR #2850, opened 2016-10-27
Fix updating DNS record for cloud hosts

mkurek:
Do not send info to DNSaaS about ip/hostname update if it is CloudHost.

### DIFF
--- a/src/ralph/networks/receivers.py
+++ b/src/ralph/networks/receivers.py
@@ -1,12 +1,29 @@
+from django.contrib.contenttypes.models import ContentType
+
 from ralph.dns.dnsaas import DNSaaS
 from ralph.virtual.models import CloudHost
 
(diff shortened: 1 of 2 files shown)


src/ralph/networks/receivers.py
```diff
@@ -1,12 +1,29 @@
...
+
 from ralph.dns.dnsaas import DNSaaS
 from ralph.virtual.models import CloudHost
 
+INVALID_MODELS = [CloudHost]
+
+
+def _should_send_dnsaas_request(ip_instance):
+    """
+    Return True if update/delete info should be sent to DNSaaS.
```
vi4m, line 11: describe more about our specific assumptions

mkurek: It is described - see line below

PR #2862, opened 2016-11-15
Middleware for calculating Ralph performance metrics

mkurek:
When enabled, this middleware collects:
* request processing time
* database queries count
* database queries time
* requests count

Right now it's (easy separable) part of ralph, but as soon as it will work out for us, I'm planning to make contribution to metrology.

### DIFF
--- a/src/ralph/lib/metrics/middlewares.py
+++ b/src/ralph/lib/metrics/middlewares.py
@@ -62,9 +62,10 @@ class RequestMetricsMiddleware(object):
         processing_time = int(
             (time.time() - request._request_start_time) * 1000
         )
-        logger.info('Processing time [ms]: {}'.format(processing_time), extra={
-            'processing_time': processing_time, **common_log_params
-        })
+        logger.info(
+            'Processing time [ms]: {}'.format(processing_time),
+            extra=dict(processing_time=processing_time, **common_log_params)
+        )
         processing_timer = Metrology.timer(processing_time_metric_name)
         processing_timer.update(processing_time)  # in miliseconds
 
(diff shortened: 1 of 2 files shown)


szok (approved):

src/ralph/lib/metrics/middlewares.py
```diff
@@ -0,0 +1,108 @@
...
+
+    How to use it:
+    * add this middleware at the beginning of MIDDLEWARE_CLASSES in your
+      settings
+    * define and start Metrology reporter in you Django settings (more details
+      here: https://metrology.readthedocs.io/en/latest/)
+    """
+
+    def process_request(self, request):
+        request._request_start_time = time.time()
```
romcheg: Better to use monotonic time to avoid a rare problem that may happen when request comes before a time change and the response is ready after it.

ar4s (approved):

src/ralph/lib/metrics/middlewares.py
```diff
@@ -0,0 +1,111 @@
...
+            'url_name': url_name,
+            'status_code': response.status_code,
+        }
+
+        # processing time
+        processing_time_metric_name = METRIC_NAME_TMPL.format(
+            prefix=PROCESSING_TIME_METRIC_PREFIX, **common_log_params
+        )
+        processing_time = int(
+            (time.time() - request._request_start_time) * 1000
```
ghost: Shouldn't it be calculated on top of this method (I mean as soon as response gets processing)?

src/ralph/lib/metrics/middlewares.py
```diff
@@ -0,0 +1,111 @@
...
+            'Processing time [ms]: {}'.format(processing_time),
+            extra=dict(processing_time=processing_time, **common_log_params)
+        )
+        processing_timer = Metrology.timer(processing_time_metric_name)
+        processing_timer.update(processing_time)  # in miliseconds
+
+        # database queries count and time
+        queries_time = 0.0  # in miliseconds
+        queries_count = 0
+        for q in connection.queries:
```
ghost, line 75: enumerate?

ghost (approved):

src/ralph/lib/metrics/middlewares.py
mkurek: We want to be as close to user as possible in these metrics

src/ralph/lib/metrics/middlewares.py
mkurek: why?

src/ralph/lib/metrics/middlewares.py
ghost: ok

src/ralph/lib/metrics/middlewares.py
mkurek: done

PR #2867, opened 2016-11-21
Support for multiple providers in openstack_sync mgmt command

mkurek:
* set `provider` key in site definition in settings to distinguish OpenStack providers
* allow to specify different auth_url and version for keystone
* support for keystone v3

### DIFF
--- a/src/ralph/virtual/management/commands/openstack_sync.py
+++ b/src/ralph/virtual/management/commands/openstack_sync.py
@@ -291,8 +291,7 @@ class Command(BaseCommand):
                     name=provider_name,
                 )
                 self._save_object(
-                    self.cloud_provider,
-                    'Add {} CloudProvider'.format(self.openstack_provider_name)
+                    cloud_provider, 'Add {} CloudProvider'.format(provider_name)
                 )
             return cloud_provider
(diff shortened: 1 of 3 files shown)


src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -36,6 +39,9 @@
     nova_client_exists = False
 
 
+DEFAULT_OPENSTACK_PROVIDER_NAME = 'openstack'
```
ghost: from settings?

ghost (approved):

src/ralph/virtual/management/commands/openstack_sync.py
mkurek: Yes and no :) As you can see [here](https://github.com/allegro/ralph/pull/2867/files#diff-4fa30bd6416c1bceb147b0bf9dbf6999L237), it was harcoded value previously.

mkurek: Ok, i moved it to settings

xor-xor (approved):

szok (approved):

PR #2873, opened 2016-11-25
Fix importing data center with management ip

ghost:


src/ralph/data_center/models/mixins.py
```diff
@@ -74,6 +74,10 @@ def management_ip(self):
 
     @management_ip.setter
     def management_ip(self, value):
+        if not value:
```
mkurek, line 77: maybe `del self.management_ip`?

mkurek (changes requested):

src/ralph/data_center/models/mixins.py
ghost: done

mkurek (approved):

PR #2878, opened 2016-11-29
Fix random tests

mkurek:
* barcode: don't use ean8, use incremented sequence
* additional message for admin list views queries count

### DIFF
--- a/src/ralph/back_office/tests/factories.py
+++ b/src/ralph/back_office/tests/factories.py
@@ -51,7 +51,7 @@ class BackOfficeAssetFactory(DjangoModelFactory):
     force_depreciation = False
     office_infrastructure = factory.SubFactory(OfficeInfrastructureFactory)
     sn = factory.Faker('ssn')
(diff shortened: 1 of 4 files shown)


src/ralph/back_office/tests/factories.py
```diff
@@ -51,7 +51,7 @@ class BackOfficeAssetFactory(DjangoModelFactory):
     force_depreciation = False
     office_infrastructure = factory.SubFactory(OfficeInfrastructureFactory)
     sn = factory.Faker('ssn')
-    barcode = factory.Faker('ean8')
+    barcode = factory.Sequence(lambda n: str(n + 10**8))
```
ghost: Shouldn't this be more like this?
`barcode = factory.Sequence(lambda n: str(10**8 + n))`

ghost: ok, got it, it's int instead of string ;P

ghost (approved):

szok (approved):

PR #2896, opened 2016-12-14
Add link to service in CloudHost change view

szok:


src/ralph/virtual/admin.py
```diff
@@ -310,6 +310,19 @@ def get_cloudproject(self, obj):
     get_cloudproject.allow_tags = True
     get_cloudproject._permission_field = 'parent'
 
+    def get_service(self, obj):
+        return '<a href="{}">{}</a>'.format(
```
mkurek: what if obj has no service_env?

mkurek (changes requested):

src/ralph/virtual/admin.py
szok: Done

src/ralph/virtual/admin.py
```diff
@@ -310,6 +310,21 @@ def get_cloudproject(self, obj):
     get_cloudproject.allow_tags = True
     get_cloudproject._permission_field = 'parent'
 
+    def get_service(self, obj):
+        if self.service_env_id:
+            return '<a href="{}">{}</a>'.format(
+                reverse(
+                    "admin:assets_service_change",
+                    args=(obj.service_env.service.id,)
```
mkurek: obj.service_env.service.id -> obj.service_env.service_id

src/ralph/virtual/admin.py
```diff
@@ -310,6 +310,21 @@ def get_cloudproject(self, obj):
...
+                ),
+                obj.service_env
+            )
+        return ''
+    get_service.short_description = _('Service env')
+    get_service.admin_order_field = 'parent'
```
mkurek: `parent`? why not `service_env`

src/ralph/virtual/admin.py
```diff
@@ -310,6 +310,21 @@ def get_cloudproject(self, obj):
...
+    get_service.allow_tags = True
+    get_service._permission_field = 'parent'
```
mkurek: ditto

mkurek (changes requested):

src/ralph/virtual/admin.py
szok: done

src/ralph/virtual/admin.py
szok: done

src/ralph/virtual/admin.py
szok: done, but it is not necessary, because I have get str(obj.service_env.service) in line 320

mkurek (approved):

PR #2899, opened 2016-12-15
Add update position if parent change

szok:


ghost (approved):

src/ralph/data_center/models/physical.py
```diff
@@ -461,9 +456,18 @@ def __repr__(self):
...
+        if self.pk:
+            update_qs = DataCenterAsset.objects.filter(parent=self)
+            # When changing rack we search and save all descendants
+            if self._previous_state['rack_id'] != self.rack_id:
+                update_qs.update(rack=self.rack)
+            # When changing position if is blade,
+            # we search and save all descendants
+            if (
+                self.is_blade and
+                self._previous_state['rack_id'] != self.position
```
mkurek: rack_id will be different than position, for sure :expressionless:

mkurek: and i don't get this if - if current server is blade, then it's child - so `update_qs` should be empty (blade should have no children)

mkurek (changes requested):

src/ralph/data_center/models/physical.py
szok: Done

mkurek (approved):

PR #2904, opened 2016-12-28
Reduce XHR requests

ar4s:


mkurek (approved):

src/ralph/admin/static/elements/badge-input.html
```diff
@@ -116,6 +116,7 @@
         this.clearInput();
       },
       setBadges: function(badges) {
+        console.log(badges);
```
mkurek: rm it

src/ralph/admin/filters.py
```diff
@@ -360,6 +361,16 @@ def get_related_url(self):
...
 
+    def get_prefetch_data(self):
+        value = self.value()
+        results = {}
+        if value:
+            queryset = self.field_model.objects.filter(
+                pk__in=value.split(',')
+            )
+            results = get_results(queryset, True)
+        return json.dumps(results)
```
mkurek, line 374: what if i select `<empty value>` in filter? will this work too?

mkurek (changes requested):

mkurek (approved):

PR #2943, opened 2017-01-20
Clenaup debian dependencies

ghost:
Hi, I was trying to build the debian package from the latest source and hit some issues. The commits are fixing those.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,13 +2,12 @@ Source: ralph
 Section: python
 Priority: extra
 Maintainer: pylabs@allegrogroup.com
-Build-Depends: debhelper (>= 9), python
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -2,13 +2,12 @@ Source: ralph
 Section: python
 Priority: extra
 Maintainer: pylabs@allegrogroup.com
-Build-Depends: debhelper (>= 9), python
```
mkurek, line 5: IMO git, python3-virtualenv, python3-dev and dh-virtualenv should not be required to install deb package - they are required only to build package

vi4m: That's why it is on Build-Depends probably? When the user install it, it will not be included, right?

mkurek: true - my mistake

mkurek (approved):

ar4s (approved):

ghost (approved):

PR #2945, opened 2017-01-20
Add hermes subscribers for Service

szok:


src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+from django.contrib.auth import get_user_model
+
+from ralph.assets.models.assets import Environment, Service, ServiceEnvironment
+
+logger = logging.getLogger(__name__)
+
+
+def _update_service_owners(service, business_owners, technical_owners):
+    """
+    Update service owners
```
xor-xor: What's the point for such docstrings, if they're just a copy of funciton's name..? :wink:

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+        for i in technical_owners
+    ]
+
+    service.business_owners = business_owners
+    service.technical_owners = technical_owners
+
+
+def _update_service_environments(service, environments):
+    """
+    Update service environments.
```
xor-xor: Improve docstring.

src/ralph/assets/tests/test_subscribers.py
```diff
@@ -0,0 +1,112 @@
...
+                    'subscriber_name': subscriber_name
+                }
+            ),
+            json.dumps(event_data),
+            content_type="application/json",
+            follow=False
+        )
+        return response
+
+    def test_add_service(self):
```
xor-xor: What about using "given-when-then" naming scheme in test cases..?

src/ralph/assets/apps.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from ralph.apps import RalphAppConfig
+
+
+class AssetsConfig(RalphAppConfig):
+
+    name = 'ralph.assets'
+
+    def get_load_modules_when_ready(self):
```
xor-xor, line 9: I can hardly see a link between name of this function and its body :wink:

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+    for env_id in to_delete:
+        ServiceEnvironment.objects.get(
+            service=service, environment_id=env_id
+        ).delete()
+
+
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['CREATE'])
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['UPDATE'])
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['REFRESH'])
+def update_service_handler(data):
```
xor-xor: I'd use `data` name only as a last resort (i.e. when you **really** can't come up with anything more descriptive).

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Update service owners
+
+    Args:
+        service: Service
+        business_owners: dict of owners
+        technical_owners: dict of owners
+    """
+    business_owners = [
+        get_user_model().objects.get_or_create(username=i['username'])[0]
```
mkurek: i'd prefer to not use get_or_create here - just get

src/ralph/assets/apps.py
szok: https://github.com/allegro/ralph/blob/ng/src/ralph/apps.py#L18

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+    new = set([u.id for u in new_envs])
+    to_delete = current - new
+    to_add = new - current
+    for env_id in to_add:
+        ServiceEnvironment.objects.create(
+            service=service,
+            environment_id=env_id
+        )
+    for env_id in to_delete:
+        ServiceEnvironment.objects.get(
```
mkurek: there should be check if there is no foreign key to this service env - if yes, then we shouldn't delete it but report an error

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+
+
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['DELETE'])
+def delete_service_handler(data):
+    """
+    Set service active to False if service deleted.
+    """
+    try:
+        Service.objects.filter(uid=data['uid']).update(active=False)
```
mkurek: check first if there is anything related to this service (for example any BaseObject) - if yes, then report an error and don't deactivate it

src/ralph/settings/test.py
```diff
@@ -88,3 +88,20 @@ def __getitem__(self, item):
 
 EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
 ENABLE_EMAIL_NOTIFICATION = True
+
+ENABLE_HERMES_INTEGRATION = True
+HERMES['ENABLED'] = ENABLE_HERMES_INTEGRATION
+HERMES_SERVICE_TOPICS = {
```
mkurek: move it to base settings

mkurek (changes requested):

src/ralph/settings/test.py
szok: it is in base settings, but returns "none", that only in production set topic

src/ralph/assets/apps.py
xor-xor: OK

ghost (approved):

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Returns:
+        True if service environment updated
+        False if can not delete ServiceEnvironment
+    """
+    new_envs = []
+    for env_name in environments:
+        new_envs.append(Environment.objects.get_or_create(name=env_name)[0])
+
+    current = set([u.id for u in service.environments.all()])
```
xor-xor: Minor nitpick - why `u` not `e` if you iterate over environments here..?

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,158 @@
...
+            service=service,
+            environment_id=env_id
+        )
+    for env_id in to_delete:
+        service_env = ServiceEnvironment.objects.get(
+            service=service, environment_id=env_id
+        )
+        if BaseObject.objects.filter(service_env=service_env).exists():
+            logger.error(
+                'Can not delete service environment because when assigned to objects.',  # noqa
```
xor-xor: `# noqa: E501`

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,158 @@
...
+
+    return True
+
+
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['CREATE'])
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['UPDATE'])
+@pyhermes.subscriber(topic=settings.HERMES_SERVICE_TOPICS['REFRESH'])
+def update_service_handler(service_data):
+    """
+    Add, remove and update Service from hermes event.
```
xor-xor: Hmm... Looking at the decorators here, I see "create", "update" and "refresh" - I guess that "remove" is included in "refresh" then..? If so, then maybe add some explaination here..?

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,158 @@
...
+                extra=service_data
+            )
+            return
+
+        Service.objects.filter(uid=service_data['uid']).update(active=False)
+    except Exception as e:
+        logger.exception(e, extra=service_data)
+    else:
+        logger.info(
+            'Delete service `{}` with UID `{}`.'.format(
```
xor-xor: Or rather - "service with UID ... deleted"..?

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,160 @@
...
+            service=service,
+            environment_id=env_id
+        )
+    for env_id in to_delete:
+        service_env = ServiceEnvironment.objects.get(
+            service=service, environment_id=env_id
+        )
+        if BaseObject.objects.filter(service_env=service_env).exists():
+            logger.error(
+                'Can not delete service environment because when assigned to objects.',  # noqa: E501
```
mkurek: `Can not delete service environment - it has assigned some base objects`

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Set service active to False if service deleted.
+    """
+    try:
+        service_envs = ServiceEnvironment.objects.filter(
+            service__uid=service_data['uid']
+        )
+        if BaseObject.objects.filter(service_env__in=service_envs).exists():
+            logger.error(
+                'Can not delete service because when assigned to objects.',
```
mkurek: ditto

mkurek (changes requested):

src/ralph/assets/subscribers.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Update business and technical owners for selected service.
+
+    Args:
+        service: Service
+        business_owners: dict of owners
+        technical_owners: dict of owners
+    """
+    business_owners = [
+        get_user_model().objects.get(username=i['username'])
```
mkurek: what if user (username?) is not found?

mkurek (changes requested):

src/ralph/assets/subscribers.py
szok: Seriously?, previously I have get_or_create you wanted to change  to get() and I told you (face to face) about it that user will not exist.

Which version i have select?

mkurek: The question we have to ask here if we want to have a service without proper owners or we don't want such situation.

Consider sth like: `business_owners = get_user_model().objects.filter(username__in=[i['username'] for i in business_owners])`

szok: Done

src/ralph/assets/subscribers.py
szok: Done

src/ralph/assets/subscribers.py
szok: Done

mkurek (approved):

PR #2948, opened 2017-01-25
Custom fields summary in GUI

mkurek:
screen shot 2017-01-25 at 12 45 35

If anyone has better idea to value in "inherited from" if value is assigned directly to current object (currently `<none>`), please step forward.

### DIFF
--- a/src/ralph/lib/custom_fields/admin.py
+++ b/src/ralph/lib/custom_fields/admin.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 from django.conf.urls import url
+from django.contrib.admin.utils import unquote
+from django.contrib.contenttypes.models import ContentType
 
 from ralph.admin import RalphAdmin, register
 from ralph.admin.mixins import RalphGenericTabularInline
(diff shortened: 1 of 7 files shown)


src/ralph/lib/custom_fields/admin.py
```diff
@@ -50,6 +52,70 @@ class CustomFieldValueInline(RalphGenericTabularInline):
...
+    def _get_custom_fields_values(obj):
+        """
+        Return custom fields values (summary, with inherited values) for
+        current object, sorted by custom field name.
+
+        Fields for single item:
+        * name - name of the custom field
+        * value - value of the custom field
+        * object - name of the object, from which CFV is inherited, if it's
+          inherited, otherwise <none>
```
ghost, line 74: `self`, instead of `<none>`?

ghost (approved):

szok (approved):

vi4m:
A nie lepiej rozdzielić te tabelki na: explicit/simple custom fields + dziedziczone custom fields?

mkurek:
Advantage of this (mine) solution is summary only in one place - in your case I'll have to skip custom fields assigned directly to this object in summary of inherited custom fields - and you'll have to "jump" between direct/simple custom fields assigned to this object, and inherited ones to get full view of custom fields state.

vi4m:
2b1ed1e6-e300-11e6-9e8c-8a51b10b9d2f-2

vi4m:
Well, either you put some data, or you inherit some data. We can than rename this table to be more readable, and remove the duplicated field(monitoring_prototype).  

So it's more applicable to the user intend (edit / add information)

mkurek:
Yes, but what should happen if I set monitoring prototype of one of object "parents" (ex. configuration_path)? It's, in fact, inherited, but overwritten by this object. It might be confusing for the user to not see this CF in this table (since it's inherited).

vi4m:
So is it inherited or overwritten? What's the point for showing inherited from "<none>"  ?

vi4m:
Oh, you mean the presence of this field, based on inheritance?

vi4m:
So the user experience will be:

1. User see inheritance table, when custom fields is not inherited
2. User decide to overwrite the custom field, inheritance table is one entry less, and the value is presented in values table
3. When the user decide to remove it, inheritance entry is back 

What potential problem do you see here?

mkurek:
i'll have to look at both places to get final state of custom fields (or look some particular custom field)

And what problem do you see in my solution? Because I cant see any specific objection above :)

vi4m:
The problem is that it's antipattern. You have 2 tables (one editable, one not) presenting the same data entries. It's even simillary named "Custom fields values" "Custom fields values summary". It's hard to understand the logic behind it, without prior knowledge of inheritance in Ralph.

The proposed design is here to elicit understanding of 2 options (either inherit, either overwrite it).

mkurek:
> It's hard to understand the logic behind it, without prior knowledge of inheritance in Ralph.

In your solution I have to know about inheritance either :)

Let's discuss it f2f tomorrow.

PR #2967, opened 2017-02-23
Add new SupportStatus

PdeNiet:
Added active/inactive status values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "repository": "allegro/ralph",
   "version": "3.0.0",
   "dependencies": {
-    "bower": "~1.4.1",
+    "bower": "~1.8.0",
     "gulp": "~3.9.0",
     "gulp-autoprefixer": "~2.3.1",
     "gulp-bower": "0.0.10",
--- a/src/ralph/admin/static/elements/badge-input.html
+++ b/src/ralph/admin/static/elements/badge-input.html
@@ -86,7 +86,7 @@
         </template>
       </div>
       <div id='input-wrapper'>
-        <input id='text-getter' is="iron-input">
+        <input id='text-getter' is="iron-input" autocomplete="off">
       </div>
     </div>
     <div style="clear: both;"></div>
--- a/src/ralph/admin/templates/admin/base.html
+++ b/src/ralph/admin/templates/admin/base.html
@@ -13,6 +13,7 @@
     <script type="text/javascript" src="{% static "admin/js/core.js" %}"></script>
     <script src="{% static "vendor/js/webcomponents-lite.js" %}"></script>
     <link rel="import" href="{% static "elements/elements-min.html" %}">
+    {% include 'admin/includes/javascripts.html' %}
   </head>
   <body class="{% if is_popup %}popup {% endif %}{% block bodyclass %}{% endblock %}"{% block body_extra_attrs %}{% endblock %}>
     {% if not is_popup %}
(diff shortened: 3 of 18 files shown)


vi4m:
It seems there is some old migrations problem - @mkurek can you take a look ? zrzut ekranu 2017-02-23 o 18 11 05

vi4m:
And second thing - what's the difference between `new` and `active` ?

PdeNiet:
There is no difference but removing 'new' resulted in ralph3 crashing (atleast with me). Based on the output ralph3 gave me (see below) it appears to expect this entry to be in there. I'm not knowledgeable enough to resolve this dependency and perhaps it is even by design. 

In order to avoid confusion, i chose to add 'inactive' and it's antonym 'active'. instead of using 'inactive' and 'new'.

---
```
  File "/opt/ralph/ralph-core/lib/python3.4/site-packages/ralph/supports/models.py", line 87, in Support
    default=SupportStatus.new.id,
AttributeError: type object 'SupportStatus' has no attribute 'new'
```

If you feel 'new' and 'inactive' would be cleaner, please let me know and i will adjust it.

mkurek:
I'm wondering if we need support status at all - right now there is only one, possible value - new - which is confusing. I'm thinking if we could deduct support status from `Date to` field - if current date is before `date_to`, then status is 'active', if current date is after `date_to`, support is 'inactive'. @PatrickdeNiet what do you think about it?

PdeNiet:
I also think in it's current state it can only cause confusion. 

I'll elaborate a bit on why i think such a field can have an added value:
Currently, i take all support contract from the API, and then see how many days until expiry are left.
If this value (for example) = 30 days or >=0 days, a mail is sent to inform the administrators action has to be taken to extend the support contract.  In some cases though, organizations decide that extending the support is not wanted (usually financial). Currently i have no way to differentiate items that should be in support and items that should be left to expire. I thought of three ways i could solve this:

1. Delete the entry from the support contracts. Pretty simple solution but i would be deleting historic data possibly containing documents etc. So not a wanted solution.
2. Check whether or not a support contract is linked to an asset. But this would break the relation between a contract and asset which i can imagine is also unwanted in some cases.
3. Use the status of a contract to determine whether it is an active contract which should be extended or if it's dormant and is OK to have expired (or going to expire).

I think the status value may be the most appropriate approach in marking support contracts as active or dormant (inactive). So to me, an expired support contract (today > date_to) can be either active or inactive, where active indicates action is required to extend it asap and inactive indicates a conscious decision has been made to not extend this support contract.

mkurek:
Interesting :) Let me discuss it with our guys handling our supports and think about it over the weekend and I'll come back to you at the beginning of the next week with solution :)

PdeNiet:
cool, thanks for your thoughts as well so far :)

mkurek:
Ok. Let's do it as you wish :)

@PatrickdeNiet could you:
* replace `new` status by `active`
* add new status `inactive`
* make `active` the default one ([here](https://github.com/allegro/ralph/blob/ng/src/ralph/supports/models.py#L87))
* run `ralph makemigrations` to generate database migration?

PdeNiet:
Will do it as soon as i find some time :) (need to read up on git and set up a non-apt version of ralph for this i think)

PdeNiet:
Unfortunately i have not been able to implement this due to errors in my setup which i am unable to resolve.

vi4m:
Not a problem, we will take care of this

romcheg:
Closing this PR due to no activity for a long period of time.

PR #2973, opened 2017-03-06
Fix child ipaddresses on network view

mkurek:
There was wrong relation passed to prefetch_related and it was silently ignored by Django


vi4m:
LGTM

vi4m (approved):

PR #2976, opened 2017-03-09
Use DRF ValidationError in mac serializer

romcheg:
In order to make validation error for mac address return 400 instead of
500, ValidationError from DRF must be userd. The one from Django causes
a 500.

### DIFF
--- a/src/ralph/lib/mixins/fields.py
+++ b/src/ralph/lib/mixins/fields.py
@@ -5,13 +5,13 @@ from django import forms
 from django.conf import settings
 from django.contrib.admin.widgets import AdminTextInputWidget
 from django.contrib.contenttypes.models import ContentType
-from rest_framework.exceptions import ValidationError
 from django.db import models
 from django.db.models.loading import get_model
 from django.forms.utils import flatatt
 from django.utils import six
 from django.utils.html import format_html, smart_urlquote
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


src/ralph/lib/mixins/fields.py
```diff
@@ -5,13 +5,13 @@
...
 from django.contrib.admin.widgets import AdminTextInputWidget
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models.loading import get_model
 from django.forms.utils import flatatt
 from django.utils import six
 from django.utils.html import format_html, smart_urlquote
 from django.utils.translation import ugettext_lazy as _
+from rest_framework.exceptions import ValidationError
```
mkurek, line 14: if you swap exceptions like this, I'm afraid it'd not work in GUI (DRF's ValidationError is not descendant of Django's ValidationError)

romcheg: It works like this:

screen shot 2017-03-15 at 16 09 02

romcheg: It's not visible on the screenshot -- I have written "BOOO!!" as a value to the MAC Address filter.

mkurek: Ok - approved then :)

mkurek (approved):

PR #2979, opened 2017-03-15
Allow filtering networks by member IP address

romcheg:


src/ralph/networks/filters.py
```diff
@@ -77,3 +80,45 @@ def queryset(self, request, queryset):
...
+        filter_str = filter_str.translate(
+            str.maketrans(
+                SEARCH_OR_SEPARATORS,
+                ' ' * len(SEARCH_OR_SEPARATORS)
+            )
+        )
+
+        filter_query = Q()
+
+        for str_addr in filter_str.split(' '):
```
mkurek: why do you replace all separators to spaces to then later split by space? couldn't you split by separators, using [re.split](https://docs.python.org/3/library/re.html#re.split) ?

src/ralph/networks/filters.py
```diff
@@ -77,3 +80,45 @@ def queryset(self, request, queryset):
...
+            try:
+                address = int(ipaddress.ip_address(str_addr))
+            except ValueError:
+                return queryset
```
mkurek: you could add warning message in such case. See [here](https://github.com/allegro/ralph/blob/ng/src/ralph/admin/filters.py#L351) for example.

mkurek (changes requested):

src/ralph/networks/filters.py
romcheg: Done

mkurek (approved):

romcheg:
@mkurek some of the tests are still broken. I will resolve that as the first thing in the morning. Please don't merge yet :)

mkurek:
@romcheg could you make some test for ipv6 too?

romcheg:
@mkurek sure!

ar4s (approved):

ghost (approved):

PR #2982, opened 2017-03-21
Add `security scan` column to `All hosts` listing

ghost:
image

image

### DIFF
--- a/src/ralph/data_center/admin.py
+++ b/src/ralph/data_center/admin.py
@@ -8,6 +8,7 @@ from django.contrib.admin.views.main import ChangeList
 from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
 from django.db.models import Prefetch, Q
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
 
 from ralph.admin import RalphAdmin, RalphTabularInline, register
@@ -60,6 +61,7 @@ from ralph.networks.forms import SimpleNetworkWithManagementIPForm
 from ralph.networks.models.networks import Network
 from ralph.networks.views import NetworkWithTerminatorsView
 from ralph.operations.views import OperationViewReadOnlyForExisiting
+from ralph.security.models import SecurityScan
 from ralph.security.views import SecurityInfo
 from ralph.supports.models import BaseObjectsSupport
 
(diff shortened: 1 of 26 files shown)


src/ralph/static/src/scss/_styles.scss
```diff
@@ -158,3 +158,13 @@ i.empty {
     color: #888;
     font-style: normal;
 }
+
+.field-scan_status {
+    text-align: center;
+}
+    .fa-check {
```
mkurek: why is this indented?

src/ralph/data_center/admin.py
```diff
@@ -595,6 +598,36 @@ def show_location(self, obj):
     show_location.short_description = _('Location')
     show_location.allow_tags = True
 
+    def scan_status(self, obj):
```
mkurek: * extract this to mixin
* mix it in:
  * DataCenterAsset
  * CloudHost
  * VirtualServer
  * DCHost
* this way will make extra query (at least one) for each row in the list - we have to make it better (with constant number of queries :))

mkurek (changes requested):

xliiv:
TODO:
 hardware, virtual, cloud

src/ralph/data_center/admin.py
ghost: Where should I put the mixin?

src/ralph/security/views.py
```diff
@@ -1,6 +1,45 @@
...
+from django.utils.translation import ugettext_lazy as _
+
 from ralph.admin.views.extra import RalphDetailView
-from ralph.security.models import SecurityScan
+
+
+class ScanStatusInChangeListMixin(object):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.list_select_related += ['securityscan', ]
```
mkurek, line 13: it would not guarantee constant number of queries - each row (each host) will have to check number of vulnerabilities for each scan - each in separated query. We need to make it better :)

src/ralph/data_center/admin.py
ghost: done

mkurek (changes requested):

vi4m:
Replace icons with text or add text next to icon.

xliiv:
Add text to icon

mkurek (approved):

ar4s (approved):

PR #2991, opened 2017-03-24
XSS flaw bugfix

asteinhauser:
Inserting Django-autosanitized values to unquoted HTML attribute values is unsecure, because the attacker can append other HTML attributes including JavaScript events. This flaw is exploitable by the following technique:
http://blog.portswigger.net/2015/11/xss-in-hidden-input-fields.html


mkurek (approved):
Thanks!

ghost (approved):

PR #2996, opened 2017-03-27
Extend the list of available operation statuses

romcheg:
This patch adds the following statuses:

 - reopened
 - todo
 - blocked


romcheg:
This won't work because it depends on another pull request

PR #2998, opened 2017-03-27
Event processing for change management

romcheg:
This patch adds a generic subscriber for receiving and recording
change management events and a Jira processor for it. Base object
loading logic should be implemented for every secific case so it
is not included.

It also makes operations identifiable by their thicket_ids and extends
the list of possible operation statuses.

### DIFF
--- a/src/ralph/admin/apps.py
+++ b/src/ralph/admin/apps.py
@@ -11,4 +11,3 @@ class RalphAdminConfig(RalphAppConfig):
     def ready(self):
         register_custom_filters()
         super().ready()
-        from ralph.operations.changemanagement.subscribtions import receive_chm_event  # noqa
--- a/src/ralph/admin/apps.py
+++ b/src/ralph/admin/apps.py
@@ -11,3 +11,4 @@ class RalphAdminConfig(RalphAppConfig):
     def ready(self):
         register_custom_filters()
         super().ready()
+        from ralph.operations.changemanagement.subscribtions import receive_chm_event  # noqa
--- a/src/ralph/operations/changemanagement/jira.py
+++ b/src/ralph/operations/changemanagement/jira.py
@@ -1,9 +1,13 @@
+import logging
 from datetime import timezone
 
 from dateutil.parser import parse as parse_datetime
 from django.conf import settings
 
 from ralph.operations.models import OperationStatus
+
+
+logger = logging.getLogger(__name__)
 
 
 def get_title(event_data):
@@ -30,7 +34,15 @@ def get_operation_status(event_data):
         status_conf['BLOCKED']: OperationStatus.blocked
     }
 
-    return status_map[event_data['issue']['fields']['status']['name']]
+    try:
+        status_str = event_data['issue']['fields']['status']['name']
+        return status_map[status_str]
+    except KeyError:
+        logger.error(
+            'Received an operation with unexpected '
+            'status: {}. Please check the settings.'.format(status_str)
+        )
+        raise
 
 
 def get_operation_name(event_data):
--- a/src/ralph/operations/changemanagement/jira.py
+++ b/src/ralph/operations/changemanagement/jira.py
@@ -1,0 +1,69 @@
+from datetime import timezone
+
+from dateutil.parser import parse as parse_datetime
+from django.conf import settings
+
+
+def get_title(event_data):
+    return event_data['issue']['fields']['summary']
+
+
+def get_description(event_data):
+    return event_data['issue']['fields']['description']
+
+
+def get_ticket_id(event_data):
+    return event_data['issue']['key']
+
+
+def get_operation_status(event_data):
+    from ralph.operations.models import OperationStatus
+
+    status_conf = settings.CHANGE_MGMT_OPERATION_STATUSES
+    status_map = {
+        status_conf['OPENED']: OperationStatus.opened,
+        status_conf['IN_PROGRESS']: OperationStatus.in_progress,
+        status_conf['RESOLVED']: OperationStatus.resolved,
+        status_conf['CLOSED']: OperationStatus.closed,
+        status_conf['REOPENED']: OperationStatus.reopened,
+        status_conf['TODO']: OperationStatus.todo,
+        status_conf['BLOCKED']: OperationStatus.blocked
+    }
+
+    return status_map[event_data['issue']['fields']['status']['name']]
+
+
+def get_operation_name(event_data):
+    return event_data['issue']['fields']['issuetype']['name']
+
+
+def get_assignee_username(event_data):
+    try:
+        return event_data['issue']['fields']['assignee']['key']
+    except TypeError:
+        # NOTE(romcheg): This means there is no assignee.
+        return None
+
+
+def get_creation_date(event_data):
+    return _safe_load_datetime(event_data, 'created')
+
+
+def get_last_update_date(event_data):
+    return _safe_load_datetime(event_data, 'updated')
+
+
+def get_resolution_date(event_data):
+    return _safe_load_datetime(event_data, 'resolutiondate')
+
+
+def _safe_load_datetime(event_data, field):
+    """Safely serialize an ISO 8601 datetime string into a datetime object."""
+
+    try:
+        datetime_str = event_data['issue']['fields'][field]
+        return parse_datetime(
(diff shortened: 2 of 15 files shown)


src/ralph/admin/apps.py
```diff
@@ -11,3 +11,4 @@ class RalphAdminConfig(RalphAppConfig):
     def ready(self):
         register_custom_filters()
         super().ready()
+        from ralph.operations.changemanagement.subscribtions import receive_chm_event  # noqa
```
mkurek: move this to `ralph.operation.apps`

src/ralph/operations/changemanagement/jira.py
```diff
@@ -0,0 +1,69 @@
...
+        status_conf['OPENED']: OperationStatus.opened,
+        status_conf['IN_PROGRESS']: OperationStatus.in_progress,
+        status_conf['RESOLVED']: OperationStatus.resolved,
+        status_conf['CLOSED']: OperationStatus.closed,
+        status_conf['REOPENED']: OperationStatus.reopened,
+        status_conf['TODO']: OperationStatus.todo,
+        status_conf['BLOCKED']: OperationStatus.blocked
+    }
+
+    return status_map[event_data['issue']['fields']['status']['name']]
```
mkurek: what if we'll receive not-mapped status (afaik jira allows to create custom statuses)?

src/ralph/operations/changemanagement/jira.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def get_operation_name(event_data):
+    return event_data['issue']['fields']['issuetype']['name']
+
+
+def get_assignee_username(event_data):
+    try:
+        return event_data['issue']['fields']['assignee']['key']
+    except TypeError:
```
mkurek, line 55: weird, it looks like `KeyError` :)

src/ralph/settings/base.py
```diff
@@ -431,6 +431,45 @@ def str_to_bool(s: str) -> bool:
...
+    ),
+    'TODO': os.getenv(
+        'CHANGE_MGMT_OPERATION_STATUS_TODO', 'Todo'
+    ),
+    'BLOCKED': os.getenv(
+        'CHANGE_MGMT_OPERATION_STATUS_BLOCKED', 'Blocked'
+    )
+}
+
+CHANGE_MGMT_BO_LOADER = os.getenv('CHANGE_MGMT_BO_LOADER', None)
```
mkurek: i'd change `BO` to `BASE_OBJECT` - `BO` could be misleading here (someone could evaluate it to `BackOffice` for example)

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,102 @@
...
+                event_data
+            ),
+            created_date=change_processor.get_creation_date(event_data),
+            update_date=change_processor.get_last_update_date(event_data),
+            resolution_date=change_processor.get_resolution_date(event_data),
+            bo_ids=bo_ids
+        )
+
+    except Exception as e:
+        # TODO(romcheg): Change this for something sensible
```
mkurek: 👍

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,102 @@
...
+
+    if username is None:
+        return
+
+    model = get_user_model()
+
+    user, created = model.objects.get_or_create(username=username)
+
+    if created:
+        user.is_active = False
```
mkurek: couldn't you pass it to `get_or_create`? Sth like `user, created = model.objects.get_or_create(username=username, defaults={'is_active': False})`

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,102 @@
...
+@transaction.atomic
+def record_operation(title, status, description, operation_name, ticket_id,
+                     assignee_username=None, created_date=None,
+                     update_date=None, resolution_date=None, bo_ids=None):
+
+    operation_type = _safe_load_operation_type(operation_name)
+
+    # NOTE(romcheg): Changes of an unknown type should not be recorded.
+    if operation_type is None:
+        return
```
mkurek: add some logging here (warning?)

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,102 @@
...
+    if bo_ids:
+        operation.base_objects = _load_base_objects(bo_ids)
+        operation.save()
+
+
+@pyhermes.subscriber(topic=settings.HERMES_CHANGE_MGMT_TOPICS['CHANGES'])
+def receive_chm_event(event_data):
+    """Process messages from the change management system."""
+    try:
+        change_processor = import_module(settings.CHANGE_MGMT_PROCESSOR)
```
mkurek: you could import it in top-level imports, not only for this function

mkurek (changes requested):

src/ralph/operations/changemanagement/subscribtions.py
romcheg: Oh.. right...

src/ralph/operations/changemanagement/jira.py
romcheg: This place has also caught me. However, if there is no assignee, event_data['issue']['fields']['assignee'] will be None. Applying [] operator to None produces a TypeError.

mkurek: 👍

src/ralph/admin/apps.py
romcheg: Done

src/ralph/operations/changemanagement/jira.py
romcheg: I would suggest to act like in the case of an unknown operation type which is to log a warning and ignore the record. What do you think?

mkurek: 👍 - but log error :)

src/ralph/settings/base.py
romcheg: Done

src/ralph/operations/changemanagement/subscribtions.py
romcheg: Done

src/ralph/operations/changemanagement/subscribtions.py
romcheg: Done

src/ralph/operations/changemanagement/subscribtions.py
romcheg: Done

src/ralph/operations/changemanagement/jira.py
romcheg: Done

src/ralph/operations/changemanagement/subscribtions.py
romcheg: Done

mkurek (approved):

ar4s (approved):

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,119 @@
...
+
+from ralph.assets.models import BaseObject
+from ralph.operations.models import Operation, OperationType
+
+
+change_processor = import_module(settings.CHANGE_MGMT_PROCESSOR)
+base_object_loader = None
+
+if settings.CHANGE_MGMT_BASE_OBJECT_LOADER:
+    global base_object_loader
```
szok, line 17: Why global ? ;(

src/ralph/operations/changemanagement/subscribtions.py
```diff
@@ -0,0 +1,119 @@
...
+    base_object_loader = import_module(settings.CHANGE_MGMT_BASE_OBJECT_LOADER)
+
+
+logger = logging.getLogger(__name__)
```
szok, line 21: please move logger to top of file

romcheg:
@szok, thank you for those catches. Since they are not affecting the functionality of the code and we need this patch ASAP for testing purposes, I would propose to merge it as is. I will resolve your concerns in a separate patch.

PR #3000, opened 2017-03-30
Fix unique constraint on Operation ticket_id field

mkurek:


romcheg:
@mkurek -- I see that unique property if being removed and changed to False in the migration. How is that supposed to add the unique constraint to the database?

mkurek:
@romcheg Django will recognise, that previously `unique` was set to `False` and now it has new value (state) - `True`, so it'll generate proper SQL command to apply this constraint.

ghost (approved):

szok (approved):

ar4s (approved):

PR #3002, opened 2017-03-31
Add is_patched to security_scan

ghost:


src/ralph/security/migrations/0007_auto_20170331_0934.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('security', '0006_securityscan_is_patched'),
+    ]
+
+    operations = [
+        migrations.RunPython(update_is_patched, reverse_func),
```
romcheg: According to the documentation the **reverse_code** parameter is optional. You can remove the entire reverse_func from this file, if there's not going to be any implementation.

src/ralph/security/migrations/0007_auto_20170331_0934.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
romcheg, line 1: Does it make sense to merge this migration with 0006_securityscan_is_patched?

src/ralph/security/migrations/0007_auto_20170331_0934.py
romcheg: I would also add a migration that sets the nullable property to False after this one.

src/ralph/security/migrations/0007_auto_20170331_0934.py
```diff
@@ -0,0 +1,40 @@
...
+        vulnerability__patch_deadline__gt=datetime.now()
+    ).values_list(
+        'securityscan_id', flat=True
+    ).distinct()
+
+    patched_scans = SecurityScan.objects.filter(id__in=is_patched_ids).update(
+        is_patched=True
+    )
+
+    print()
```
romcheg: Was this empty print left intentionally? Not an issue, just checking :)

src/ralph/security/models.py
```diff
@@ -82,6 +86,11 @@ class SecurityScan(
         on_delete=models.CASCADE,
     )
     vulnerabilities = models.ManyToManyField(Vulnerability, blank=True)
+    # this is a quirk field, it is updated manually (for now it's in API)
+    # this is because it's hard to handling it automatically
+    # (its value is computated depending on M2M field and M2M signals are
+    # complicated)
+    is_patched = models.BooleanField(default=False)
```
romcheg, line 93: Perhaps it makes sense to make it non-nullable, doesn't it?

romcheg (changes requested):
A few small things. Otherwise LGTM.

src/ralph/security/migrations/0007_auto_20170331_0934.py
ghost: yup

src/ralph/security/migrations/0007_auto_20170331_0934.py
ghost: 0007 is custom migrations. i'd like to have them seperated unless there is a reason.
Is there a reason? :)

src/ralph/security/migrations/0007_auto_20170331_0934.py
ghost: If I remove it,  revers migration will be still possible?

src/ralph/security/models.py
ghost: I think it is.
https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.BooleanField

> If you need to accept null values then use NullBooleanField instead.

src/ralph/security/migrations/0007_auto_20170331_0934.py
romcheg: No, but there is a special noop code for this available as [RunPython.noop](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#django.db.migrations.operations.RunPython.noop)

src/ralph/security/migrations/0007_auto_20170331_0934.py
romcheg: No reason IIRC except personal preferences actually. Let's leave it as is it then.

src/ralph/security/migrations/0007_auto_20170331_0934.py
ghost: I don't see example of using it so, I prefere my version which it :sparkles: just-works :sparkles: 
anyway, when I'm using it like this:

```
      operations = [                                                            
          migrations.RunPython(update_is_patched, migrations.RunSQL.noop),         
      ] 
```

i get:

`ValueError: RunPython must be supplied with callable arguments `

src/ralph/security/models.py
romcheg: Ok, makes sense

src/ralph/security/migrations/0007_auto_20170331_0934.py
romcheg: That's weird... let's leave it as is then.

romcheg (approved):

src/ralph/security/migrations/0007_auto_20170331_0934.py
```diff
@@ -0,0 +1,40 @@
...
+
+def update_is_patched(apps, schema_editor):
+    SecurityScan = apps.get_model("security", "SecurityScan")
+    is_patched_ids = SecurityScan.vulnerabilities.through.objects.filter(
+        vulnerability__patch_deadline__gt=datetime.now()
+    ).values_list(
+        'securityscan_id', flat=True
+    ).distinct()
+
+    patched_scans = SecurityScan.objects.filter(id__in=is_patched_ids).update(
```
mkurek: in this case, when my sec scan has two vulnerabilities - one with deadline before today, and one with deadline after today, it'll be marked as is_patched, isn't it?

src/ralph/security/migrations/0007_auto_20170331_0934.py
mkurek: @xliiv `migrations.RunPython.noop` (not `migrations.RunSQL.noop`)

mkurek (changes requested):

src/ralph/security/migrations/0007_auto_20170331_0934.py
ghost: now fixed

mkurek (approved):

PR #3009, opened 2017-04-07
Fix `Assets Supports` export

andrzej-jankowski:
Now export works even when exists support without price.

Also:
* Use MySQL as default DB in tests (SQLite is not supported anymore)

### DIFF
--- a/src/ralph/settings/test.py
+++ b/src/ralph/settings/test.py
@@ -7,26 +7,13 @@ sys.path.append(os.path.join(BASE_DIR, '..', '..', 'contrib', 'dhcp_agent'))
 
 DEBUG = False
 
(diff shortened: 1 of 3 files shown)


src/ralph/settings/test.py
```diff
@@ -7,26 +7,13 @@
 
 DEBUG = False
 
-TEST_DB_ENGINE = os.environ.get('TEST_DB_ENGINE', 'sqlite')
-
-if TEST_DB_ENGINE == 'mysql':
-    # use default mysql settings
-    if not os.environ.get('DATABASE_PASSWORD'):
```
mkurek: could you keep this options for mysql? :)

mkurek (approved):

PR #3014, opened 2017-04-11
Refactor operation status

romcheg:
List of possible operation statuses along with their meaning is
a flexible matter, especially in cases, when an external change
management software is used. Given that, having to rewrite and
release a new version of Ralph in order to adapt to the changes
is not a way to go.

This patch removes a hard-coded list of statuses from Ralph and
replaces it with a reference to a manageable entity called
OperationStatus. Those entities are manageable botj over the API
and the UI.

The change management listener's behavior was also set to create
new operation types in responce to incomming messages instead of
ignoring operations that have an unknown status.

### DIFF
--- a/src/ralph/operations/api.py
+++ b/src/ralph/operations/api.py
@@ -3,7 +3,7 @@ from django.db.models import Prefetch
 from ralph.accounts.api_simple import SimpleRalphUserSerializer
 from ralph.api import RalphAPISerializer, RalphAPIViewSet, router
 from ralph.assets.models import BaseObject
-from ralph.operations.models import Operation, OperationType
+from ralph.operations.models import Operation, OperationStatus, OperationType
 
 
 class OperationTypeSerializer(RalphAPISerializer):
@@ -13,9 +13,17 @@ class OperationTypeSerializer(RalphAPISerializer):
         depth = 1
 
 
+class OperationStatusSerializer(RalphAPISerializer):
+
+    class Meta:
+        model = OperationStatus
+        depth = 1
+
+
 class OperationSerializer(RalphAPISerializer):
     type = OperationTypeSerializer()
     assignee = SimpleRalphUserSerializer()
+    status = OperationStatusSerializer()
 
     class Meta:
         model = Operation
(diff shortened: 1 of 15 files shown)


src/ralph/operations/api.py
```diff
@@ -13,9 +13,17 @@ class Meta:
...
+
+    class Meta:
+        model = OperationStatus
+        depth = 1
+
+
 class OperationSerializer(RalphAPISerializer):
     type = OperationTypeSerializer()
     assignee = SimpleRalphUserSerializer()
+    status = OperationStatusSerializer()
```
andrzej-jankowski: Maybe SlugRelatedField (http://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) will be better here? You can define this field like that:
```
status = SlugRelatedField(
    many=False,
    read_only=False,
    slug_field='name',
    queryset= OperationStatus.objects.all()
)
```

Should work :)

romcheg: That's an interesting option actually. However, I would like to discuss this topic a little in order to find the best solution. My concern is caused by the fact that now statuses are not hardcoded but stored in DB and allow dynamic management. Perhaps slugifying will bring more issues to API consumers than comfort. Say, someone wants to store an operation in an external data store and then use its status in order to find other operations. If the status is slugified, it won't be possible, if the name of the status was changed. So, I would vote for exposing statuses as structures. @mkurek, @vi4m, your opinions?

andrzej-jankowski (approved):

src/ralph/operations/filters.py
```diff
@@ -0,0 +1,22 @@
...
+        self._choices_list = [
+            (status.id, status.name) for status in statuses
+        ]
+
+    def queryset(self, request, queryset):
+        if not self.value():
+            return queryset
+
+        return queryset.filter(
+            status__id=int(self.value())
```
mkurek: `status_id` (`status__id` will probably do join here)

mkurek (changes requested):

src/ralph/operations/filters.py
romcheg: Good catch. Thanks!

mkurek (approved):

PR #3021, opened 2017-04-14
Fix VIP event data validation

andrzej-jankowski:


src/ralph/data_center/subscribers.py
```diff
@@ -35,7 +35,7 @@ def validate_vip_event_data(data):
     except (TypeError, ValidationError):
         err = 'invalid IP address "{}"'.format(ip)
         errors.append(err)
-    if not port or port < 1024 or port > 49151:
+    if not port or port < 0 or port > 65535:
```
mkurek: 0 is invalid too as far as I remember, so please change to `<= 0`

mkurek (approved):

PR #3026, opened 2017-04-19
Fix counting vulnerabilities in listing

ghost:
Replace vuls. counting in listing with simple string "vulnerable".
This increase view rendering. Items are now links to details of scan.

### DIFF
--- a/src/ralph/data_center/admin.py
+++ b/src/ralph/data_center/admin.py
@@ -283,7 +283,7 @@ class DataCenterAssetOperation(OperationViewReadOnlyForExisiting):
 
 
 class DataCenterAssetSecurityInfo(SecurityInfo):
-    url_name = 'datacenter_asset_security_info'
+    url_name = 'security_info'
 
 
 @register(DataCenterAsset)
--- a/src/ralph/data_center/admin.py
+++ b/src/ralph/data_center/admin.py
@@ -283,7 +283,7 @@ class DataCenterAssetOperation(OperationViewReadOnlyForExisiting):
 
 
 class DataCenterAssetSecurityInfo(SecurityInfo):
-    url_name = 'security_info'
+    url_name = 'datacenter_asset_security_info'
 
 
 @register(DataCenterAsset)
--- a/src/ralph/security/views.py
+++ b/src/ralph/security/views.py
@@ -18,17 +18,16 @@ logger = logging.getLogger(__name__)
 
 
 @lru_cache()
-def _url_name_for_change_view(obj, view_name):
+def _url_name_for_change_view(obj_class, view_name):
     """
     Returns `url_name` attribute of view named by `view_name` or None
 
-    Takes registred ModelAdmin for `obj` and searches its `change_views` for
-    view with name `view_name`.  Then returns `url_name` for found view.
+    Takes registred ModelAdmin for `obj_class` and searches its `change_views`
+    for view with name `view_name`.  Then returns `url_name` for found view.
     """
-    obj_type = type(obj)
-    if obj_type not in ralph_site._registry:
+    if obj_class not in ralph_site._registry:
         return None
-    model_admin = ralph_site._registry[obj_type]
+    model_admin = ralph_site._registry[obj_class]
     found_view = None
     for change_view in getattr(model_admin, 'change_views', []):
         if getattr(change_view, 'name', '') == view_name:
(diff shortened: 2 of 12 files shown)


ar4s (approved):

src/ralph/data_center/admin.py
```diff
@@ -283,7 +283,7 @@ class DataCenterAssetOperation(OperationViewReadOnlyForExisiting):
 
 
 class DataCenterAssetSecurityInfo(SecurityInfo):
-    url_name = 'datacenter_asset_security_info'
+    url_name = 'security_info'
```
mkurek: change it for cluster too: 
```
django.core.urlresolvers.NoReverseMatch: Reverse for 'data_center_cluster_security_info' with arguments '(78698,)' and keyword arguments '{}' not found. 0 pattern(s) tried: []
```

mkurek: And, if you change all url_names to `security_info`, breadcrumbs in this tab do not work.

andrzej-jankowski (approved):

src/ralph/security/views.py
```diff
@@ -17,7 +20,16 @@ def __init__(self, *args, **kwargs):
     def get_queryset(self, request):
         qs = super().get_queryset(request)
         qs = qs.annotate(
-            vulnerabilities_count=Count('securityscan__vulnerabilities')
+            exceeded_vulnerabilities_count=Sum(
```
mkurek: this query is sooooo slow - on my instance it runs over 400ms for list of objects and over 3 seconds (!!!) when searching (comparing to 50ms for list and 350ms for search - so queries are now 10x longer). We have to do it better. Could we use `is_patched` here?

mkurek (changes requested):

src/ralph/security/views.py
```diff
@@ -31,16 +43,19 @@ def scan_status(self, obj):
...
                     html = self._to_span(
                         "alert", "Got vulnerabilities: {}".format(
-                            obj.vulnerabilities_count
+                            obj.exceeded_vulnerabilities_count
                         )
                     )
                 else:
                     html = self._to_span("success", "Host clean")
             else:
                 html = self._to_span("warning", "Scan failed")
```
mkurek: where do you find these classes (success, warning) for span? It's not working right now (afair it should be red when host has vulnerabilities)?

mkurek (changes requested):

src/ralph/data_center/admin.py
ghost: ok, then i can revert this rename thing.. and create for example some map dict or something, it's productive however could be more errorproof..
what do you think?

mkurek: yes. if sth will not be in this map then just don't reverse url (and display it as simple text) and log error.

src/ralph/security/views.py
ghost: about `is_patched`, I dont see it as solution, could you expand your concept?

xliiv: https://github.com/allegro/ralph/blob/cf9ec22565e79334cce588a14103e119edebe8b1/src/ralph/static/src/scss/_styles.scss#L162
from PR:
https://github.com/allegro/ralph/pull/2982/files#diff-3c0e4dc1a693950df4b64a82cfa9b6a2

src/ralph/security/views.py
ghost: https://github.com/allegro/ralph/blob/cf9ec22565e79334cce588a14103e119edebe8b1/src/ralph/static/src/scss/_styles.scss#L162
from PR:
https://github.com/allegro/ralph/pull/2982/files#diff-3c0e4dc1a693950df4b64a82cfa9b6a2

mkurek: ok - so probably i've got some old static files. Ignore this comment.

src/ralph/security/views.py
```diff
@@ -1,26 +1,43 @@
...
+from ralph.admin.helpers import get_admin_url
+from ralph.admin.sites import ralph_site
 from ralph.admin.views.extra import RalphDetailView
 from ralph.security.models import Vulnerability
 
 
+logger = logging.getLogger(__name__)
+
+
+def _url_name_for_change_view(obj, view_name):
```
mkurek: could we cache it (based on `type(obj)` and `view_name`)? use for example `django.utils.lru_cache`

mkurek: and add some docstring here

mkurek (changes requested):

src/ralph/security/views.py
```diff
@@ -16,7 +17,14 @@
 logger = logging.getLogger(__name__)
 
 
+@lru_cache()
```
mkurek: this cache will not work right now - the param to this function has to be type, not object

mkurek (changes requested):

mkurek (approved):

PR #3030, opened 2017-04-20
Fetching available networks for cloud hosts

romcheg:
This patch allows to fetch available networks for cloud hosts
basing on the rack that contains the corresponding DC asset. It
also makes possible to fetch available networks for virtual servers
that are hosted by cloud hosts.

### DIFF
--- a/src/ralph/virtual/models.py
+++ b/src/ralph/virtual/models.py
@@ -206,8 +206,8 @@ class CloudHost(PreviousStateMixin,
 
     @cached_property
     def rack(self):
-        if isinstance(self.hypervisor, DataCenterAsset):
-                return self.hypervisor.rack
+        if self.hypervisor:
+            return self.hypervisor.rack
         return None
 
     @property
@@ -373,10 +373,7 @@ class VirtualServer(
     def rack(self):
         if self.parent_id:
             polymorphic_parent = self.polymorphic_parent.last_descendant
-            if (
-                isinstance(polymorphic_parent, DataCenterAsset) or
-                isinstance(polymorphic_parent, CloudHost)
-            ):
+            if (isinstance(polymorphic_parent, (DataCenterAsset, CloudHost))):
                 return polymorphic_parent.rack
         return None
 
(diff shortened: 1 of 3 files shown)


src/ralph/virtual/tests/test_models.py
```diff
@@ -25,6 +30,35 @@
 )
 
 
+# NOTE(romcheg): If at some point someone finds a better name for this
+#                they are welcome to propose it or change it oneself.
+class CommonCodeMixin(object):
```
andrzej-jankowski: Maybe: VirtualAssetTestMixin?

src/ralph/virtual/models.py
```diff
@@ -360,7 +373,10 @@ def rack_id(self):
     def rack(self):
         if self.parent_id:
             polymorphic_parent = self.polymorphic_parent.last_descendant
-            if isinstance(polymorphic_parent, DataCenterAsset):
```
mkurek, line 363: `if isinstance(polymorphic_parent, (DataCenterAsset, CloudHost))`

src/ralph/virtual/tests/test_models.py
mkurek: maybe `NetworkableBaseObjectTestMixin`?

src/ralph/virtual/models.py
```diff
@@ -197,6 +200,16 @@ class Meta:
     def __str__(self):
         return self.hostname
 
+    @cached_property
+    def rack_id(self):
+        return self.rack.id if self.rack else None
+
+    @cached_property
+    def rack(self):
+        if isinstance(self.hypervisor, DataCenterAsset):
```
mkurek: hypervisor has to be either DataCenterAsset or None, so simple check `if self.hypervisor` will be enough here

mkurek (changes requested):

src/ralph/virtual/models.py
romcheg: Done

andrzej-jankowski (approved):

mkurek (approved):

PR #3033, opened 2017-04-24
Add management command `update_is_patched` (security module)

ghost:


src/ralph/security/management/commands/update_is_patched.py
```diff
@@ -0,0 +1,38 @@
...
+from ralph.security.models import SecurityScan
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Update `is_patched` field on each SecurityScan'
+
+    def _update_is_patched(self):
```
mkurek: wrap it into transaction

src/ralph/security/management/commands/update_is_patched.py
```diff
@@ -0,0 +1,38 @@
...
+        # mark all as patched
+        patched = SecurityScan.objects.update(is_patched=True)
```
mkurek: IMO it'd be better to not do this, and instead:
* fetch all not_patched_ids (like below)
* filter sec. scan by these ids and only the ones which are patched (`is_patched=True`)
* update them to be not patched

(theoretically transactions should protect us before messing up here, but it'd be more db-friendly - less queries and, whats more important, less modifications (updates))

mkurek (changes requested):

mkurek (approved):

andrzej-jankowski (approved):

PR #3037, opened 2017-04-24
Match cloud hosts and DC assert on OpenStack sync

romcheg:


andrzej-jankowski (approved):

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -588,6 +664,8 @@ def _update_ralph(self):
         for project_id in self.openstack_projects:
             self._add_project(self.openstack_projects[project_id], project_id)
 
+        self._match_physical_and_cloud_hosts()
```
mkurek: could you run this only if some argument is passed to this command (ex. `ralph openstack_sync --match-ironic-physical-hosts`)?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -649,6 +727,9 @@ def handle(self, *args, **options):
         if not keystone_client_exists:
             logger.error("keystoneclient module is not installed")
             raise ImportError("No module named keystoneclient")
+        if not ironic_client_exists:
```
mkurek: this should be required only in case of matching with ironic, so put this cmd argument here to `if args.match_ironic_physical_host and not ironic_client_exists` etc

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +368,75 @@ def _get_hypervisor(host_name, server_id):
             ))
             return None
 
+    def _match_physical_and_cloud_hosts(self):
+        """Connect CloudHosts and DC assets according to data from Ironic."""
+
+        provider = getattr(
+            settings, 'OPENSTACK_IRONIC_PROVIDER', 'openstack-ironic'
+        )
+        os_conf = next((conf for conf in settings.OPENSTACK_INSTANCES
```
mkurek: here you're processing only first found ironic config - what if there are multiple of them?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +368,75 @@ def _get_hypervisor(host_name, server_id):
             ))
             return None
 
+    def _match_physical_and_cloud_hosts(self):
+        """Connect CloudHosts and DC assets according to data from Ironic."""
+
+        provider = getattr(
```
mkurek: if you'd use this flag (`--match-ironic-physical-hosts` or similar), this wouldn't be necessary - just use current provider passed in args

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +368,75 @@ def _get_hypervisor(host_name, server_id):
...
+        )
+        os_conf = next((conf for conf in settings.OPENSTACK_INSTANCES
+                        if conf['provider'] == provider), None)
+
+        if os_conf is None:
+            logger.error('Ironic is not configured.')
+            return
+
+        ironic_client = get_ironic_client(
+            api_version=os_conf['version'],
```
mkurek: `os_conf['version']` contains nova version - is it **always** the same as ironic api version?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +368,75 @@ def _get_hypervisor(host_name, server_id):
...
+                )
+            except CloudHost.DoesNotExist:
+                logger.warning(
+                    not_found_message_tpl.format(
+                        'Cloud host',
+                        node.instance_uuid
+                    )
+                )
+            except DataCenterAsset.MultipleObjectsReturned:
+                logger.error(
```
mkurek, line 427: fyi: https://github.com/allegro/ralph/blob/ng/src/ralph/assets/models/assets.py#L349 - unless `node.extra['serial_number']` is None, this case will never occur (sn is unique)

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +368,75 @@ def _get_hypervisor(host_name, server_id):
...
+        not_found_message_tpl = (
+            '{} with the host id or serial number {} was not found. Check if '
+            'Ralph is synchronized with OpenStack or add it manually.'
+        )
+
+        for node in nodes:
+            try:
+                host = CloudHost.objects.get(host_id=node.instance_uuid)
+                asset = DataCenterAsset.objects.get(
+                    sn=node.extra['serial_number']
```
mkurek: what if there is no `serial_number` in `node.extra`? What if it is present, but it's `None`?

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -39,6 +39,13 @@
     nova_client_exists = False
 
 
+try:
+    from ironicclient.client import get_client as get_ironic_client
```
mkurek, line 43: please add this to openstack requirements

mkurek (changes requested):
I'd be glad to see some (even minor) test for this functionality :)

src/ralph/virtual/management/commands/openstack_sync.py
romcheg: If that's an ironic configuration, I suppose, it will provide Ironic API version.

src/ralph/virtual/management/commands/openstack_sync.py
romcheg: Done

src/ralph/virtual/management/commands/openstack_sync.py
romcheg: Done

src/ralph/virtual/management/commands/openstack_sync.py
romcheg: Looks like this code will handle this case

src/ralph/virtual/management/commands/openstack_sync.py
mkurek: so why don't you remove it? :)

src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -361,6 +373,78 @@ def _get_hypervisor(host_name, server_id):
...
 
+    def _match_physical_and_cloud_hosts(self):
+        """Connect CloudHosts and DC assets according to data from Ironic."""
+
+        provider = getattr(
+            settings, 'OPENSTACK_IRONIC_PROVIDER', 'openstack-ironic'
+        )
+
+        for os_conf in settings.OPENSTACK_INSTANCES:
+            if os_conf['provider'] != provider:
```
mkurek: use `self.openstack_provider_name` here

mkurek (changes requested):

src/ralph/virtual/management/commands/openstack_sync.py
romcheg: I need a different provider here

romcheg: Done

mkurek (approved):

PR #3060, opened 2017-05-09
Filtering by configuration path and module for more objects

romcheg:
This patch adds ability to filter objects DataCenterAsset, DCHost,
VirtualServer or CloudHost types by their configuration_path.module
and configuration_path.path.

### DIFF
--- a/src/ralph/virtual/admin.py
+++ b/src/ralph/virtual/admin.py
@@ -98,7 +98,7 @@ class VirtualServerAdmin(
     list_filter = [
         BaseObjectHostnameFilter, 'sn', 'service_env', IPFilter,
         'parent', TagsListFilter, MacAddressFilter,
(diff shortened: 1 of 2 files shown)


ghost (approved):

src/ralph/virtual/admin.py
```diff
@@ -92,12 +92,13 @@ class VirtualServerAdmin(
     CustomFieldValueAdminMixin,
     TransitionAdminMixin,
     RalphAdmin
-):
+    ):
```
mkurek: hmm.. no :)

mkurek (changes requested):

src/ralph/virtual/admin.py
romcheg: That change was for sure made by evil hackers!

mkurek (approved):

andrzej-jankowski (approved):

PR #3072, opened 2017-05-18
Add new transition to back-office: assign_hostname

ghost:


src/ralph/back_office/tests/test_models.py
```diff
@@ -345,6 +345,31 @@ def test_assign_owner(self):
             request=self.request
         )
 
+    def test_assign_hostname(self):
```
mkurek: please make the same test but with hostname already assigned to BackOfficeAsset (it should not change then)

mkurek (changes requested):

mkurek (approved):

andrzej-jankowski (approved):

PR #3080, opened 2017-05-24
Fixed override of AUTH_LDAP_USER_FLAGS_BY_GROUP

ni-cwinkelmann:
- The attribute `AUTH_LDAP_USER_FLAGS_BY_GROUP`
  of `django_auth_ldap` is capable of setting the `is_active`,
  `is_staff` and `is_superuser` as well.
  But it got overridden by the receiver `staff_superuser_populate`
- Now both flavours are possible

### DIFF
--- a/src/ralph/accounts/ldap.py
+++ b/src/ralph/accounts/ldap.py
@@ -21,11 +21,15 @@ LDAPSettings.defaults['GROUP_MAPPING'] = {}
 
 @receiver(populate_user)
 def staff_superuser_populate(sender, user, ldap_user, **kwargs):
-    user.is_superuser = 'superuser' in ldap_user.group_names
+    user.is_superuser = True if 'superuser' in ldap_user.group_names \
+        else user.is_superuser
+
     # only staff users will have access to ralph now,
     # because ralph using django admin panel
-    user.is_staff = 'active' in ldap_user.group_names
-    user.is_active = 'active' in ldap_user.group_names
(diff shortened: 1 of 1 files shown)


vi4m (approved):
LGTM

mkurek:
I'm a little bit sceptical about this change - we won't be able to revoke superuser/if_staff/active flags from user - in other words, if user became superuser, it will be superuser forever (unless someone will edit this user manually).

@ni-christopher as far as i understand, the problem is here: https://github.com/allegro/ralph/blob/ng/src/ralph/accounts/ldap.py#L160 (`memberOf` attribute) - do you have any equivalent attribute in OpenLDAP? ~~If so, could you extract it to settings?~~

EDIT: btw, i see in [django-auth-ldap](https://bitbucket.org/psagers/django-auth-ldap/src/e7bbb9698f1466658127f4cb92099b9eaaa576e0/django_auth_ldap/config.py?at=default&fileviewer=file-view-default#config.py-527) source code, that there is `member_attr` - so instead of using settings, you could just pass this param to `MappedGroupOfNamesType` constructor.

romcheg:
@ni-christopher this PR is pretty old, are you going to do a fixup for @mkurek's concern?

ni-cwinkelmann:
Hey there. I'm sorry for not taking care of this that long. Totally got distracted. I will close it as it may not be relevant anymore.

PR #3081, opened 2017-05-24
Replaced metrology by statsd client

mkurek:
Metrology has to much stuff (with threads, locks, multiple unecessary reporter, statsd reporter not working correctly) and require system libraries (libffi). Moreover it aggregated stats locally (per process), which does not work well with gunicorn. It's now replaced by simple statsd client (and metrics are aggregated in statsd).


andrzej-jankowski (approved):

vi4m (approved):

vi4m:
@todo: update documentation

PR #3083, opened 2017-05-25
Fix randomness of test_options_filtering

ghost:


mkurek:
Consider https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual instead of casting to set

mkurek (approved):

PR #3085, opened 2017-06-02
Allow ordering DC assets by location

romcheg:
Since location is a complex field Ralph will order assets
by their datacenter, server room, rack, position and slot
in the specified order if those attributes are available.


src/ralph/data_center/admin.py
```diff
@@ -286,6 +286,41 @@ class DataCenterAssetSecurityInfo(SecurityInfo):
     url_name = 'datacenter_asset_security_info'
 
 
+class DataCenterAssetChangeList(ChangeList):
```
mkurek: use `RalphChangeList` (`ralph.admin.views.main.RalphChangeList`) here

mkurek (changes requested):

mkurek (approved):

ar4s (approved):

PR #3086, opened 2017-06-02
Make rack attribute mandatory for DC assets

romcheg:


src/ralph/data_center/api/serializers.py
```diff
@@ -116,13 +117,24 @@ class Meta:
 
 
 class DataCenterAssetSerializer(ComponentSerializerMixin, AssetSerializer):
-    rack = SimpleRackSerializer()
+    rack = SimpleRackSerializer(required=True, allow_null=False)
```
mkurek: since this is now serializer for read only, this won't be necessary

mkurek (changes requested):

mkurek (approved):

ar4s (approved):

PR #3090, opened 2017-06-12
Introduce SCM scan capabilities

romcheg:
- Add models with appropriate migrations to store SCM
   scan results for different objects.
 - Append above mentioned objects' admins with the last
   SCM scan status.
 - Allow filtering by last SCM status where applicable.
 - Introduce an API endpoint to set SCM status for objects.

### DIFF
--- a/src/ralph/configuration_management/api.py
+++ b/src/ralph/configuration_management/api.py
@@ -48,7 +48,7 @@ class SCMScanViewSet(RalphAPIViewSet):
             reduce(operator.or_, queries)
         ).distinct().first()
 
-    def set(self, request, hostname):
+    def create(self, request, hostname):
         """Sets a SCM scan record for an object having matching hostname."""
 
         bo = self.get_baseobject(hostname)
@@ -81,7 +81,7 @@ router.register('scm-scan', SCMScanViewSet)
 urlpatterns = [
     url(
             r'^scm-scan/(?P<hostname>[\w\.-]+)',
-            SCMScanViewSet.as_view({'post': 'set'}),
+            SCMScanViewSet.as_view({'post': 'create'}),
             name='scm-scan-post'
     )
 ]
--- a/src/ralph/configuration_management/api.py
+++ b/src/ralph/configuration_management/api.py
@@ -9,23 +9,23 @@ from rest_framework.response import Response
 from ralph.api import RalphAPISerializer, RalphAPIViewSet, router
 from ralph.api.serializers import RalphAPISaveSerializer
 from ralph.assets.models import BaseObject
-from ralph.configuration_management.models import SCMScan
+from ralph.configuration_management.models import SCMStatusCheck
 
 
 class SCMScanSerializer(RalphAPISerializer):
     class Meta:
-        model = SCMScan
+        model = SCMStatusCheck
 
 
 class SCMScanSaveSerializer(RalphAPISaveSerializer):
 
     class Meta:
-        fields = ('last_scan_date', 'scan_status')
-        model = SCMScan
+        fields = ('last_checked', 'check_result')
+        model = SCMStatusCheck
 
 
 class SCMScanViewSet(RalphAPIViewSet):
-    queryset = SCMScan.objects.all()
+    queryset = SCMStatusCheck.objects.all()
     serializer_class = SCMScanSerializer
     save_serializer_class = SCMScanSaveSerializer
 
@@ -63,11 +63,11 @@ class SCMScanViewSet(RalphAPIViewSet):
         serializer.is_valid(raise_exception=True)
 
         update_data = {
-            'last_scan_date': serializer.validated_data['last_scan_date'],
-            'scan_status': serializer.validated_data['scan_status']
+            'last_checked': serializer.validated_data['last_checked'],
+            'check_result': serializer.validated_data['check_result']
         }
 
-        scan, created = SCMScan.objects.update_or_create(
+        scan, created = SCMStatusCheck.objects.update_or_create(
             base_object_id=bo.id,
             defaults=update_data
         )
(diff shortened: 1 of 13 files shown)


romcheg:
The list of objects with different SCM scan statuses looks like this

screen_shot_2017-06-14_at_10_34_02

romcheg:
I would like to also center those icons both vertically and horizontally, however, my CSS skills are close to CSS skills of a copy-paste monkey, so I would like to ask @ar4s for some help here.

src/ralph/configuration_management/views.py
```diff
@@ -0,0 +1,110 @@
...
+                'scmscan__scan_status',
+                self.custom_title_filter('SCM scan status')
+            ),
+            (
+                'scmscan__last_scan_date',
+                self.custom_title_filter('Last SCM scan date')
+            )
+        ]
+
+    # NOTE(romcheg): This could be probably moved to a common place.
```
mkurek: ralph/admin/filters.py?

src/ralph/configuration_management/views.py
```diff
@@ -0,0 +1,110 @@
...
+    def custom_title_filter(self, title):
+
+        class CustomTitledFilter(FieldListFilter):
+            def __new__(cls, *args, **kwargs):
+                filter_instance = FieldListFilter.create(*args, **kwargs)
+                filter_instance.title = title
```
mkurek: why this `title` cannot be param to `__init__` (not passed to `super().__init__`)?

src/ralph/configuration_management/templates/configuration_management/scminfo/scm_info.html
```diff
@@ -0,0 +1,31 @@
...
+        <td>{{ scm_scan.get_scan_status_display }}</td>
+      </tr>
+      <tr>
+        <td>{% trans "Details" %}</td>
+        <td><a href="{{ scm_tool_url|default:"#" }}">{{ scm_tool_url|default:"-" }}</a></td>
+      </tr>
+    </table>
+
+  {% else %}
+    <p>{% trans "The status of the software configuration management is not verified yet." %}</p>
```
mkurek: Maybe sth like `The status of the software configuration management for this host is unknown.`?

src/ralph/configuration_management/api.py
```diff
@@ -0,0 +1,87 @@
...
+        res_status = status.HTTP_201_CREATED if created else status.HTTP_200_OK
+
+        return Response(self.serializer_class(scan).data, status=res_status)
+
+
+router.register('scm-scan', SCMScanViewSet)
+urlpatterns = [
+    url(
+            r'^scm-scan/(?P<hostname>[\w\.-]+)',
+            SCMScanViewSet.as_view({'post': 'set'}),
```
mkurek: use `create` name for `post` instead of `set` to keep it compatible (at least at the level of the names) with DRF

mkurek (changes requested):

src/ralph/configuration_management/api.py
romcheg: Done

src/ralph/configuration_management/templates/configuration_management/scminfo/scm_info.html
romcheg: Done

src/ralph/configuration_management/views.py
romcheg: Because this filter must be an instance of FieldListFilter but in the upstream code they don't use `__init__` to create objects. They use `create()` instead. I cannot return stuff from `__init__` so I've used `__new__`.

src/ralph/configuration_management/views.py
romcheg: Done

src/ralph/configuration_management/models.py
```diff
@@ -0,0 +1,35 @@
...
+    models.Model
+):
+    """Represents software configuration management scan."""
+
+    base_object = models.OneToOneField(BaseObject, on_delete=models.CASCADE)
+    last_scan_date = models.DateTimeField(
+        blank=False, null=False,
+        verbose_name=_("Last SCM scan date")
+    )
+    scan_status = models.PositiveIntegerField(
```
mkurek: I'm not sure if there is any "scan" related with scm (maybe it should be just `status`)

src/ralph/configuration_management/models.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SCMScan(
+    PermByFieldMixin,
+    TimeStampMixin,
+    models.Model
+):
+    """Represents software configuration management scan."""
+
+    base_object = models.OneToOneField(BaseObject, on_delete=models.CASCADE)
+    last_scan_date = models.DateTimeField(
```
mkurek: consider `last_run_date` instead - @misiek08 what do you think?

romcheg: Perhaps it makes change to get rid of the word Scan for this thing. @mkurek, @misiek08 what do you think, if I call it SCMStatusCheck with the attributes called last_checked and check_result. Possible values for check_result could be scm_ok, smc_error, check_failed.

src/ralph/configuration_management/views.py
```diff
@@ -0,0 +1,99 @@
...
+                hostname=hostname
+            )
+
+        return context
+
+
+class SCMStatusCheckInChangeListMixin(object):
+
+    icon_no_scan = '-'
+    icon_scan_ok = '<i class="fa fa-check-circle" aria-hidden="true"></i>'
```
mkurek: dj.choices has a "little feature" called extra - https://github.com/ambv/dj.choices/#extra-attributes-on-choices - consider using it for classes for different icons (`fa-check-circle` etc)

romcheg: Done

ar4s (approved):

mkurek (approved):

PR #3105, opened 2017-06-26
Fix tests to run on local environment

romcheg:
* Fix generator misusage
 * Use operation_types fixtures for admin.tests.tests_views
 * Set proper charset and collation for test databases


ar4s (approved):

romcheg:
@ar4s I will have to upload a new version, because postgresql does not support setting both the collation and the charset at the same time. Thank you for the approve anyway :)

PR #3109, opened 2017-06-28
Migrate to Xenial and python 3.5

romcheg:


Dockerfile
```diff
@@ -1,15 +1,17 @@
-FROM ubuntu:14.04
+FROM ubuntu:xenial
 MAINTAINER PyLabs pylabs@allegrogroup.com
 
 # set UTF-8 locale
+RUN apt-get clean && apt-get update
```
mkurek, line 5: why?

Dockerfile
```diff
@@ -1,15 +1,17 @@
...
+RUN apt-get install locales
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 
 # set paths
 ENV RALPH_DIR=/opt/ralph
-ENV RALPH_EXEC=ralph
+ENV RALPH_EXEC=dev_ralph
```
mkurek: we shouldn't run this on prod :)

Dockerfile
```diff
@@ -1,15 +1,17 @@
-FROM ubuntu:14.04
+FROM ubuntu:xenial
 MAINTAINER PyLabs pylabs@allegrogroup.com
 
 # set UTF-8 locale
+RUN apt-get clean && apt-get update
+RUN apt-get install locales
```
mkurek, line 6: could you add it to `provision.sh`?

Dockerfile
romcheg: Whoops, that was done for debuging purposes only. Will change it back to prod

Dockerfile
romcheg: locale-gen is not installed in xenial by default and it's used here on L7

Dockerfile
romcheg: Damn, I was living my life for a few days being sure I've commented here. It comes, I missed to press one extra button to do that.

I have moved this from the provisioner to the Dockerfile because locale-gen is used in here but it's not installed by default in xenial.

ar4s (approved):

PR #3115, opened 2017-07-03
Use python-3.5 in the debian package

romcheg:


mkurek:
What about users of older ubuntu/debians which don't include python-3.5 in their packages?

romcheg:
@mkurek should we recommend them to update to the latest stable/lts version of their distributions? There is another patch that changes the version of python from 3.4 to 3.5 in the test environment, so shipping a deb package on the older version of python that works a bit differently could lead to certain problems.

mkurek (approved):

mkurek:
remember about the docs :)

vi4m:
@romcheg docs please

PR #3124, opened 2017-07-24
Buyout date as a field

ar4s:


src/ralph/assets/migrations/0027_asset_buyout_date.py
```diff
@@ -0,0 +1,52 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='asset',
+            name='buyout_date',
+            field=models.DateField(blank=True, db_index=True, null=True),
+        ),
+        migrations.RunPython(
+            update_buyout_date, reverse_code=noop
```
mkurek: you could use `RunPython.noop` for `noop`

src/ralph/back_office/admin.py
```diff
@@ -108,8 +109,8 @@ class BackOfficeAssetAdmin(
         'user', 'owner', 'user__segment', 'user__company', 'user__department',
         'user__employee_id', 'property_of', 'invoice_no', 'invoice_date',
         'order_no', 'provider', 'budget_info', 'depreciation_rate',
-        'depreciation_end_date', 'force_depreciation', LiquidatedStatusFilter,
-        TagsListFilter
+        'depreciation_end_date', 'force_depreciation', 'buyout_date',
```
mkurek: `buyout_date` filter should filter out assets which model's category has `show_buyout_date` set to `False`

mkurek (changes requested):

mkurek (dismissed):

mkurek (approved):

PR #3144, opened 2017-08-18
Remove unique_together constraint on min_ip/max_ip

### LINKED ISSUE #3143: min_ip, max_ip fields erroring despite being specified via JSON REST API
As of commit 4794c4c04fced29b397b58a7689ff725c20ff6bd I'm having a problem where the JSON rest API won't let me create networks - https://22.0.0.2/api/networks/ (docker container instance of Ralph)
a request sent to 
```json
{
   "name":"test3",
   "address":"2.0.0.0/8",
   "remarks":"",
   "vlan":1,
   "dhcp_broadcast":false,
   "reserved_from_beginning":0,
   "reserved_from_end":0,
   "gateway":null,
   "network_environment":null,
   "kind":null,
   "service_env":null,
   "terminators":[

   ],
   "racks":[

   ],
   "dns_servers":[

   ],
   "min_ip":16777216,
   "max_ip":33554431
}
```
produces the following result:
```json
{
  "min_ip": [
    "This field is required."
  ],
  "max_ip": [
    "This field is required."
  ]
}
```

This seems to be a known issue with Django REST API, and is possibly a result of min_ip and max_ip being set as non-editable in the model? 

The same request works via the form request mechanism.

wrouesnel:
This constraint breaks saving via the REST API due to the use of the save() method to calculate these fields values.

Resolves #3143.

From my reading of the code this constraint shouldn't be required so long as the model is calculating this field, since the validation is being enforced there. But otherwise the validator code needs to become custom in order to ignore the missing min_ip/max_ip fields when posting via REST.

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -397,7 +397,7 @@ class Network(
     class Meta:
         verbose_name = _('network')
         verbose_name_plural = _('networks')
-        unique_together = ('min_ip', 'max_ip')
+        #unique_together = ('min_ip', 'max_ip')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)


mkurek:
This constraint should **not** be removed. Try to add `min_ip` and `max_ip` to excluded fields here: https://github.com/allegro/ralph/blob/ng/src/ralph/networks/api.py#L38

wrouesnel:
@mkurek That was my first approach - it doesn't work (`unique_together` forces the present constraint). The only option (that I can see from reading the Django docs) is to cancel all the validators and code them manually (see: http://www.django-rest-framework.org/api-guide/validators/#uniquetogethervalidator)

Blejwi:
There is new PR with changes to min/max IP fields so I'm closing this PR.
See #3365.

PR #3150, opened 2017-08-28
Add ``display_name`` field to ``Vulnerability``

ar4s:


src/ralph/security/models.py
```diff
@@ -46,7 +46,13 @@ class Vulnerability(
     name = models.CharField(
         verbose_name=_("name"),
         max_length=1024,
-        unique=False)
+        unique=False
+    )
+    short_name = models.CharField(
+        verbose_name=_("short name"),
+        max_length=1024,
+        unique=False
```
romcheg: I would say it should be unique -- otherwise that means the same vulnarability

szok: delete unique=false

romcheg (changes requested):
One nit inline. Otherwise LGTM

szok (dismissed):

src/ralph/security/models.py
```diff
@@ -46,7 +46,12 @@ class Vulnerability(
     name = models.CharField(
         verbose_name=_("name"),
         max_length=1024,
-        unique=False)
+        unique=False
+    )
+    short_name = models.CharField(
+        verbose_name=_("short name"),
+        max_length=1024
```
romcheg: From what I see unique is False by default: https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L135

romcheg (changes requested):

src/ralph/security/models.py
ar4s: I renamed this field to ``display_name``

romcheg (dismissed):
All my concerns have been resolved. LGTM, thanks!

szok (approved):

PR #3149, opened 2017-08-28
Push graphs to statsd

ar4s:

PR #3155, opened 2017-09-01
Fix #3149

ar4s:
Fix missing statsd, which is optional and used only in "dashboards" area.

It's occuring just after fresh install of Ralph ```pip3 install -r requirements/base.txt``` 

```ModuleNotFoundError: No module named 'statsd'```

### DIFF
--- a/src/ralph/lib/metrics/collector.py
+++ b/src/ralph/lib/metrics/collector.py
@@ -31,6 +31,7 @@ class TimerMock(ContextDecorator):
     def stop(self, *args, **kwargs):
         pass
 
+
 class StatsdMockClient(object):
     def __init__(self, *args, **kwargs):
         logger.warning(
@@ -54,7 +55,6 @@ if STATSD_IS_INSTALLED:
     PREFIX = getattr(settings, 'STATSD_PREFIX', defaults.PREFIX)
     MAXUDPSIZE = getattr(settings, 'STATSD_MAXUDPSIZE', defaults.MAXUDPSIZE)
     IPV6 = getattr(settings, 'STATSD_IPV6', defaults.IPV6)
-
 
     def build_statsd_client(
         host=HOST, port=PORT, prefix=PREFIX, maxudpsize=MAXUDPSIZE, ipv6=IPV6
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/ralph/lib/metrics/collector.py [hunk 1]
```diff
@@ -2,63 +2,74 @@
 from contextlib import ContextDecorator
 
 from django.conf import settings
-from statsd import defaults, StatsClient
+
+STATSD_IS_INSTALLED = False
+try:
+    from statsd import defaults, StatsClient
+    STATSD_IS_INSTALLED = False
```
vi4m: should be `True`

vi4m (changes requested):

src/ralph/lib/metrics/collector.py
```diff
@@ -2,63 +2,74 @@
...
+    ):
+        return StatsClient(
+            host=host,
+            port=port,
+            prefix=prefix,
+            maxudpsize=maxudpsize,
+            ipv6=ipv6
+        )
+else:
+    def build_statsd_client(*args, **kwargs):
```
vi4m, line 70: Why not have one function (`def build_statsd_client(*args, **kwargs)`:) with conditional statements inside? Creating alternative functions at runtime, inside the conditional smells like anti-pattern

src/ralph/lib/metrics/collector.py
```diff
@@ -2,63 +2,74 @@
...
+    HOST = getattr(settings, 'STATSD_HOST', defaults.HOST)
+    PORT = getattr(settings, 'STATSD_PORT', defaults.PORT)
+    PREFIX = getattr(settings, 'STATSD_PREFIX', defaults.PREFIX)
+    MAXUDPSIZE = getattr(settings, 'STATSD_MAXUDPSIZE', defaults.MAXUDPSIZE)
+    IPV6 = getattr(settings, 'STATSD_IPV6', defaults.IPV6)
+
+    def build_statsd_client(
+        host=HOST, port=PORT, prefix=PREFIX, maxudpsize=MAXUDPSIZE, ipv6=IPV6
+    ):
+        return StatsClient(
```
pnuckowski, line 62: Why not to create custom "test" StatsClient?
```python
class TestStatsClient(StatsClient):
    def _after(self, data):
        pass
```
No need to create `TimerMock` etc

vi4m:
@ar4s it's taking too long to complete this. I'm closing this one.

PR #3172, opened 2017-09-26
Add extra aggregate function and add aggregate expresion

ar4s:


src/ralph/dashboards/models.py
```diff
@@ -13,6 +14,19 @@
 )
 
 
+def _unpack_series(series):
+    if not isinstance(series, str):
+        raise ValueError('Series should be string')
+    series_field, fn = _to_pair(series, '|')
+    if (series_field != fn) and (fn != 'distinct'):
```
mkurek, line 21: we should change logic of `_to_pair` asap - this condition is terrible - couldn't `_to_pair` return `None` if there is nothing to split?

mkurek (dismissed):

mkurek (approved):

PR #3178, opened 2017-10-06
Add constraint to ``dns_servers_group`` on ``Network``

ar4s:
To prevent delete all related objects with ``DNSServerGroup``, like
``Networks``.

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -320,6 +320,7 @@ class Network(
         'dhcp.DNSServerGroup',
         null=True,
         blank=True,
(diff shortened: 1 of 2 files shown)


src/ralph/networks/models/networks.py
```diff
@@ -320,6 +320,7 @@ class Network(
         'dhcp.DNSServerGroup',
         null=True,
         blank=True,
+        on_delete=models.SET_NULL,
```
mkurek: why not `models.PROTECT`?

mkurek (approved):

PR #3180, opened 2017-10-11
Add custom target model

ar4s:
Changes:
 - add new option - ``target``
 - improve performance by applying select_related from admin model

### DIFF
--- a/src/ralph/dashboards/models.py
+++ b/src/ralph/dashboards/models.py
@@ -1,12 +1,20 @@
 from dj.choices import Choices
+from dj.choices.fields import ChoiceField
 from django.contrib.contenttypes.models import ContentType
 from django.db import connection, models
 from django.db.models import Case, Count, IntegerField, Max, Q, Sum, Value, When
+from django.db.models.fields import BooleanField
 from django.db.models.functions import Coalesce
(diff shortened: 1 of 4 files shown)


mkurek (approved):

src/ralph/dashboards/models.py
```diff
@@ -283,3 +326,38 @@ def render(self, **context):
...
+        elif isinstance(field, BooleanField):
+            graph_item = field.to_python(graph_item)
+        return graph_item
+
+    def get_queryset_for_filter(self, queryset, value):
+        filter_key = self.changelist_filter_key
+        value = self.normalize_changelist_value(value)
+        if self.custom_changelist_model:
+            value_param = self.params['target'].get('value', 'id')
+            values = self.build_queryset(annotated=False).filter(
```
mkurek, line 354: shouldn't queryset be passed here to `build_queryset`?

mkurek (changes requested):

src/ralph/dashboards/models.py
ar4s: only for line 359 is necessary

mkurek (approved):

PR #3188, opened 2017-11-13
Custom template for transtions

ar4s:


docs/development/transitions.md
```diff
@@ -1,6 +1,16 @@
 # Transitions
 The transition change state (e.g. status) of object from one to another - that's helps in product lifecycle management. For each object (asset, support, licence) you can define some workflow (set of transitions) and provide special actions for each transition.
 
+## Custom template for transition
+For each a transition you can separately specify Django's template in the transition's form, to do this add new item to ``TRANSITION_TEMPLATES`` in your settings file. For instance:
```
mkurek: For each transition

src/ralph/lib/transitions/checks.py
```diff
@@ -0,0 +1,36 @@
...
+                    ))
+                    continue
+                try:
+                    get_template(path)
+                except TemplateDoesNotExist:
+                    errors.append(Error(
+                        'Template {} ({}) doesn\'t exist'.format(
+                            template, path
+                        ),
+                        hint='Check TEMPLATES settings',
```
mkurek: did you mean `TRANSITION_TEMPLATES`?

src/ralph/lib/transitions/forms.py
```diff
@@ -13,6 +13,9 @@
 )
 
 
+TRANSITION_TEMPLATES = getattr(settings, 'TRANSITION_TEMPLATES', None)
```
mkurek: why getattr? add it to base settings and use `TRANSITION_TEMPLATES = settings.TRANSITION_TEMPLATES` here

src/ralph/lib/transitions/models.py
```diff
@@ -511,6 +511,7 @@ class Transition(models.Model):
     source = JSONField()
     target = models.CharField(max_length=50)
     actions = models.ManyToManyField('Action')
+    template_name = models.CharField(max_length=255, blank=True, default='')
```
mkurek, line 514: what will (should?) happen if I save template_name to DB and then remove it from settings?

mkurek (changes requested):

src/ralph/lib/transitions/models.py
ar4s: I will add some check for this which emit some warning.

mkurek (approved):

PR #3193, opened 2017-12-11
Change passed arguments to transitions

ar4s:
Main changes:
 - pass ``requester`` argument to precondition function
 - refactor - pass ``requester`` argument instead of ``request`` in transition function


src/ralph/lib/transitions/async.py
```diff
@@ -60,9 +60,14 @@ def run_async_transition(job_id):
 def _perform_async_transition(transition_job):
     transition = transition_job.transition
     obj = transition_job.obj
-    _check_instances_for_transition([obj], transition, check_async_job=False)
+    requester = transition.jobs.first().user
```
mkurek: why don't you use `transition_job` object?

src/ralph/lib/transitions/models.py
```diff
@@ -424,7 +436,7 @@ def run_field_transition(
     for instance in instances:
         _post_transition_instance_processing(
             instance, transition, data, history_kwargs=history_kwargs,
-            user=kwargs['request'].user, attachments=attachments,
+            user=requester, attachments=attachments,
```
mkurek: maybe `user` param could be renamed to `requester` as well to keep it consistent?

mkurek (commented):
So the main question here is, why (do you want to change it)? :)

mkurek (dismissed):

mkurek (dismissed):

mkurek (approved):

PR #3194, opened 2017-12-12
Add new precondition and action to back office asset

ar4s:


src/ralph/back_office/models.py
```diff
@@ -526,6 +537,12 @@ def _generate_report(cls, name, request, instances, language):
                 _('Document autogenerated by {} transition.').format(name)
             )
 
+    @classmethod
+    @transition_action(precondition=_check_assets_owner)
+    def must_be_owner_of_asset(cls, instances, request, **kwargs):
```
mkurek: do not define `request` here explicitly (because of #3193)

mkurek (changes requested):

mkurek (approved):

PR #3197, opened 2017-12-18
Remove request argument from transitions' actions

ar4s:


src/ralph/back_office/models.py
```diff
@@ -246,10 +245,6 @@ def render_template(template):
         self.hostname = last_hostname.formatted_hostname(fill=counter_length)
         if commit:
             self.save()
-        if request:
```
mkurek, line 249: remove request param in this method too

src/ralph/back_office/admin.py
```diff
@@ -166,6 +167,14 @@ class BackOfficeAssetAdmin(
         }),
     )
 
+    def save_model(self, request, obj, form, change):
+        hostname_changed = obj.hostname != form.cleaned_data['hostname']
```
mkurek: this will not work when hostname is changed by transition

mkurek (changes requested):

mkurek (dismissed):

mkurek (approved):

PR #3201, opened 2017-12-20
Refactor ``graph-query`` in dashboards

ar4s:


mkurek (dismissed):

src/ralph/dashboards/helpers.py
```diff
@@ -10,3 +15,23 @@ def decode_params(value):
     if not value:
         return {}
     return json.loads(base64.urlsafe_b64decode(value).decode())
+
+
+def normalize_value(model_class, labels, value, separator):
```
mkurek: any description?

src/ralph/dashboards/helpers.py
```diff
@@ -10,3 +15,23 @@ def decode_params(value):
...
+    field = get_field_by_relation_path(
+        model_class,
+        labels.split(separator)[0]
```
mkurek: instead of passing here labels and separator just to split it here, maybe pass preprocessed value?

src/ralph/dashboards/models.py
```diff
@@ -327,37 +319,17 @@ def render(self, **context):
...
-        value = self.normalize_changelist_value(value)
         if self.custom_changelist_model:
             value_param = self.params['target'].get('value', 'id')
             values = self.build_queryset(annotated=False).filter(
-                **{self.params['labels']: value}
-            ).values_list(value_param, flat=True)
-            queryset = queryset.filter(**{filter_key: values})
+                **filters
+            ).distinct().values_list(value_param, flat=True)
+            queryset = queryset.distinct().filter(**{filter_key: values})
```
mkurek: are you sure that using distinct here (every time) is necessary?

src/ralph/dashboards/models.py
```diff
@@ -327,37 +319,17 @@ def render(self, **context):
...
         else:
+            # value = self.normalize_changelist_value(value)
```
mkurek: is it necessary?

mkurek (changes requested):

src/ralph/dashboards/models.py
ar4s: is necessary

mkurek (approved):

PR #3209, opened 2018-01-11
Add basic status endpoints (ping and health)

mkurek:


src/ralph/health_check.py
```diff
@@ -0,0 +1,86 @@
...
+                }
+            )
+            messages.append(msg)
+    return messages
+
+
+@api_view(['GET'])
+@permission_classes((AllowAny,))
+@renderer_classes((PlainTextRenderer,))
+def status_ping(request):
```
ar4s: it will be better if you use vanilla views without decorators :)

mkurek: done

ar4s (approved):

Issue #3215, opened 2018-01-17
no write access to /home/ubuntu when vm_user = "ubuntu" in Vagrantfile


lowang-bh:
-    vm_user = "ubuntu"
+    vm_user = "vagrant"
If vm_user is "ubuntu", then VM_USER=${VM_USER:-ubuntu} will be ubuntu, but the login user is vagrant actually when setup the develop environment. This will cause no write access to /home/ubuntu with login user of vagrant. Change it to "vagrant" it  recover from error.

rcsheets:
I believe I've reproduced this issue. The following `vagrant up` output seems relevant:

    default: Collecting virtualenv
    default:   Downloading virtualenv-15.1.0-py2.py3-none-any.whl (1.8MB)
    default: Installing collected packages: virtualenv
    default: Successfully installed virtualenv-15.1.0
    default: Already using interpreter /usr/bin/python3
    default: Using base prefix '/usr'
    default: Traceback (most recent call last):
    default:   File "/usr/local/bin/virtualenv", line 11, in <module>
    default:     sys.exit(main())
    default:   File "/usr/local/lib/python3.5/dist-packages/virtualenv.py", line 713, in main
    default:     symlink=options.symlink)
    default:   File "/usr/local/lib/python3.5/dist-packages/virtualenv.py", line 925, in create_environment
    default:     site_packages=site_packages, clear=clear, symlink=symlink))
    default:   File "/usr/local/lib/python3.5/dist-packages/virtualenv.py", line 1110, in install_python
    default:     mkdir(lib_dir)
    default:   File "/usr/local/lib/python3.5/dist-packages/virtualenv.py", line 323, in mkdir
    default:     os.makedirs(path)
    default:   File "/usr/lib/python3.5/os.py", line 231, in makedirs
    default:     makedirs(head, mode, exist_ok)
    default:   File "/usr/lib/python3.5/os.py", line 231, in makedirs
    default:     makedirs(head, mode, exist_ok)
    default:   File "/usr/lib/python3.5/os.py", line 241, in makedirs
    default:     mkdir(name, mode)
    default: PermissionError: [Errno 13] Permission denied: '/home/ubuntu/venv'

rcsheets:
This seems like it could be closed now.

PR #3219, opened 2018-01-21
Fix the issue that vagrant user has no write access to /home/ubuntu

lowang-bh:
The default user is vagrant when using "vagrant up", so vagrant user will have no access to "/home/ubuntu" when create venv directory or any other operations need write access.
When change "ubuntu" into "vagrant", the error disappeared.


mkurek (approved):

rcsheets:
Just for reference, this should fix #3215

PR #3220, opened 2018-01-22
Add relations view for data center assets

romcheg:
At the moment only related cloud hosts are shown. If the idea works
in practice, more relations will be rendered.

### DIFF
--- a/src/ralph/data_center/views.py
+++ b/src/ralph/data_center/views.py
@@ -1,0 +1,23 @@
+from ralph.admin.views.extra import RalphDetailView
+
+
+class RelationsView(RalphDetailView):
+    icon = 'shekel'
+    label = 'Relations'
+    name = 'relations'
+    url_name = 'relations'
+    template_name = 'data_center/datacenterasset/relations.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        related_objects = {}
+
+        if self.object.cloudhost_set.exists():
+            related_objects['cloud_hosts'] = [
+                i for i in self.object.cloudhost_set.all()
+            ]
+
+        context['related_objects'] = related_objects
+
+        return context
--- a/src/ralph/data_center/views.py
+++ b/src/ralph/data_center/views.py
@@ -18,9 +18,7 @@ class RelationsView(RalphDetailView):
         context = super().get_context_data(**kwargs)
 
         related_objects = {}
-
         self._add_cloud_hosts(related_objects)
-
         context['related_objects'] = related_objects
 
         return context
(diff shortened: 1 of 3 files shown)


ar4s (dismissed):

src/ralph/data_center/views.py
```diff
@@ -0,0 +1,23 @@
...
+    url_name = 'relations'
+    template_name = 'data_center/datacenterasset/relations.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        related_objects = {}
+
+        if self.object.cloudhost_set.exists():
+            related_objects['cloud_hosts'] = [
```
mkurek: why not `related_objects['cloud_hosts'] = list(self.object.cloudhost_set.all())`?

src/ralph/data_center/views.py
```diff
@@ -0,0 +1,23 @@
...
+    name = 'relations'
+    url_name = 'relations'
+    template_name = 'data_center/datacenterasset/relations.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        related_objects = {}
+
+        if self.object.cloudhost_set.exists():
```
mkurek: couldn't you just fetch all cloudhosts (even if there is none) - now you're making two queries, while you could do this in one?

mkurek (commented):
Did you consider using `ralph.lib.Table` (https://github.com/allegro/ralph/blob/ng/src/ralph/lib/table/table.py)?

src/ralph/data_center/views.py
romcheg: I should probably do that because I suspect exists() will actually rise an exception, if the DCA does not host any CH.

mkurek: if `.exists()` will raise an exception if there is non CH attached to DCA, you'll get 500 instead of `No related objects found.`, wouldn't you?

romcheg: yup, that's why I've changed the current approach in my next patch set

src/ralph/data_center/views.py
romcheg: Done

ar4s (dismissed):

src/ralph/data_center/views.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if cloud_hosts:
+            related_objects['cloud_hosts'] = cloud_hosts
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        related_objects = {}
+
+        self._add_cloud_hosts(related_objects)
```
mkurek: are these empty lines necessary here?

ar4s (approved):

mkurek (approved):

PR #3222, opened 2018-01-23
Change to wider exception - DatabaseError instead of ProgrammingError

### LINKED ISSUE #3216: Table 'ralph_ng.transitions_transition' doesn't exist
when I follow the document to setup a develop environment, I met the error" default: django.db.utils.ProgrammingError: (1146, "Table 'ralph_ng.transitions_transition' doesn't exist") ". I think it is because there are no such tables when newly install ralph3 develop environment but ralph3 try to migrate them(from ralph2). I am on mac and have download the default box manually which will  be used in vagrant up.

ar4s:
Fix #3216

### DIFF
--- a/src/ralph/lib/transitions/checks.py
+++ b/src/ralph/lib/transitions/checks.py
@@ -1,7 +1,12 @@
+import logging
+
 from django.core.checks import Error
-from django.db.utils import OperationalError
+from django.db.utils import DatabaseError
 from django.template.base import TemplateDoesNotExist
 from django.template.loader import get_template
+
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


mkurek (approved):

lowang-bh (commented):
How about including line 53 to line 55 due to "Transition.objects" is the root cause, but as the lazy-update of ORM, the exception just raise in the iteration

ar4s:
@lowang-bh queryset in ``transitions_with_custom_templates`` are evaluated in the for loop (according to https://docs.djangoproject.com/en/1.8/ref/models/querysets/#when-querysets-are-evaluated) and I tested this PR on Vagrant machine provisioned from scratch.

PR #3229, opened 2018-02-06
Refactor tiles on dashboard

ar4s:
Changes:
 - delete cache
 - add two new tiles - Your Equipment and Equipments to accept
 - tiles respect user's permission


MarekBleschke (approved):
looks ok

PR #3233, opened 2018-02-19
Add property_of_id to invoice report

ar4s:


MarekBleschke (approved):

src/ralph/assets/invoice_report.py
```diff
@@ -106,6 +106,10 @@ def _get_report_data(self, request, queryset):
         first_item = queryset[0]
         data = {
             'id': str(slugify(first_item.invoice_no)),
+            'property_of_id': (
+                first_item.property_of.id
```
mkurek, line 110: you could use `first_item.property_of_id` here (now you have to fetch related object from db)

PR #3235, opened 2018-02-21
additional ip and hostname transition

szymi-:


src/ralph/data_center/models/physical.py
```diff
@@ -52,6 +53,30 @@
...
+    """
+    network_environments = []
+    networks = []
+    for obj in objects:
+        network_environments.append(
+            set(obj._get_available_network_environments())
+        )
+    network_environments = set.intersection(*network_environments)
+    for environment in network_environments:
+        for network in environment.network_set.all():
```
mkurek: shouldn't we allow only for networks which are available for all considered objects (not all networks in these networks environments)? I would not involve network environment here and focus only on networks - consider using `obj. _get_available_networks`

mkurek (dismissed):

ar4s (dismissed):

ar4s (approved):

PR #3239, opened 2018-03-01
Test for displaying a BO asset with limited permissions

romcheg:


src/ralph/back_office/tests/test_models.py
```diff
@@ -468,3 +470,39 @@ def test_requester_must_be_specified(self):
...
+        permissions = Permission.objects.exclude(
+            codename__in=[
+                'view_backofficeasset_hostname_field',
+                'view_backofficeasset_service_env_field',
+                'change_backofficeasset_hostname_field',
+                'change_backofficeasset_service_env_field',
+            ]
+        ).all()
+
+        for p in permissions:
```
ar4s: try ``user.user_permissions.add(*permissions)``

src/ralph/back_office/tests/test_models.py
```diff
@@ -468,3 +470,39 @@ def test_requester_must_be_specified(self):
...
+        user = UserFactory()
+
+        user.is_superuse = False
+        user.is_staff = False
+        user.set_password(passwd)
+
+        user.save()
+
+        # Grant all permissions to the user
+        permissions = Permission.objects.exclude(
```
ar4s, line 488: please split the test into two separated tests: first for ``hostname``, second for ``service_env`` and cover a conditions in this block of code: https://github.com/allegro/ralph/blob/ng/src/ralph/back_office/admin.py#L71-L79

src/ralph/back_office/tests/test_models.py
```diff
@@ -468,3 +470,39 @@ def test_requester_must_be_specified(self):
...
+
+        self.login_as_user(user=user, password=passwd)
+
+        url = reverse(
+            'admin:back_office_backofficeasset_change',
+            args=(asset.pk,)
+        )
+        resp = self.client.get(url)
+
+        self.assertEqual(resp.status_code, 200)
```
ar4s: add ``self.assertTrue(resp.context['form']['service_env'])``

ar4s (changes requested):

src/ralph/back_office/tests/test_models.py
romcheg: Makes sense.

src/ralph/back_office/tests/test_models.py
romcheg: It worked.

src/ralph/back_office/tests/test_models.py
romcheg: @ar4s Please take a look at the new patch set.

ar4s (approved):

Issue #3244, opened 2018-03-13
#3240 breaks `ralph migrate`


rcsheets:
Django's `prepare_sql_script` function requires `sqlparse`, which was  removed in #3240.

There is an ability to fall back to some legacy code in Django 1.8, but that is removed in Django 1.9, so we should probably re-introduce `sqlparse` somewhere in the requirements.

ar4s:
I will create a pull request to fix it. I'm sorry for that :(

NOTE: https://github.com/django/django/blob/1.8.19/django/db/backends/base/operations.py#L268-L284

rcsheets:
Confirmed fixed. Thanks.

PR #3245, opened 2018-03-13
Add sqlpqarse to requirements/base.txt

ar4s:
Django requires sqlparse for DB's vendors (excluding PostgreSQL).
@rcsheets thanks for testing.
For futher information, see
https://github.com/django/django/blob/2162f0983de0dfe2178531638ce7ea56f54dd4e7/django/db/backends/base/operations.py#L289-L307


ar4s:
Fix #3244

mkurek (approved):

PR #3254, opened 2018-04-10
Handle non integer counter in assets' hostname

ar4s:


mkurek:
did you consider using `__iregex` lookup to fetch only hostname matching pattern like  `{self.hostname_template_prefix}[0-9]+{self.hostname_template_postfix}`?

ar4s:
No, I will try.

mkurek (dismissed):

src/ralph/networks/models/networks.py
```diff
@@ -165,23 +165,26 @@ def current_counter_without_model(self):
...
-                hostnames.append(
-                    item.hostname[
-                        len(self.hostname_template_prefix):
-                        -len(self.hostname_template_postfix)
-                    ]
-                )
-
+                try:
+                    counter = int(item.hostname[start:stop])
+                except ValueError:
```
MarekBleschke: I think it would be good to add here comment that there can be non integer counters therefore we skip this error.

MarekBleschke (approved):

PR #3255, opened 2018-04-17
Introduce cloud synchronisation

romcheg:
This patch extends cloud providers to contain necessary settings for cloud syncronisation such as switchable option for enabling or disabling the syncronisation for every specific provider and pluggable syncronisation event processor.  The cloud syncronisation endpoint accepts POST requests for every cloud provider for which the syncronisation is enabled and an event processor is selected. Events are routed for necessary processors in the context of a selected cloud provider.

### DIFF
--- a/src/ralph/apps.py
+++ b/src/ralph/apps.py
@@ -14,6 +14,7 @@ class RalphAppConfig(AppConfig):
         when app is ready.
         """
         super().ready()
+
         package = self.module.__name__
         for module in self.get_load_modules_when_ready():
             try:
--- a/src/ralph/virtual/cloudsync.py
+++ b/src/ralph/virtual/cloudsync.py
@@ -1,0 +1,82 @@
+import json
+import logging
+import threading
+
+import pkg_resources
+from django.db.models import Q
+from django.http import (
+    HttpResponse,
+    HttpResponseBadRequest,
+    HttpResponseNotFound
+)
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+
+from ralph.virtual.models import CloudProvider
+
+logger = logging.getLogger(__name__)
+
+
+ENTRY_POINTS_GROUP = 'ralph.cloud_sync_processors'
+CLOUD_SYNC_DRIVERS = None
+LOCK = threading.Lock()
+
+
+def load_processors():
+    global CLOUD_SYNC_DRIVERS
+
+    # NOTE(romcheg): Check if processors were loaded prior to locking anything.
+    if CLOUD_SYNC_DRIVERS is not None:
+        return
+
+    with LOCK:
+        # NOTE(romcheg): Double check in order to avoid race conditions.
+        if CLOUD_SYNC_DRIVERS is not None:
+            return
+
+        logger.info('Loading cloud sync processors.')
+
+        CLOUD_SYNC_DRIVERS = {}
+
+        for ep in pkg_resources.iter_entry_points(ENTRY_POINTS_GROUP):
+            try:
+                CLOUD_SYNC_DRIVERS[ep.name] = ep.resolve()
+            except ImportError:
+                logger.error(
+                    'Could not import DC asset event processor from {}.'
+                    ''.format(ep.module_name)
+                )
+
+
+@csrf_exempt
+@require_POST
+def cloud_sync_router(request, cloud_provider_id):
+    load_processors()
+
+    raw_data = request.read().decode('utf-8')
+
+    try:
+        event_data = json.loads(raw_data)
+    except ValueError:
+        return HttpResponseBadRequest('Content must be a valid JSON text.')
+
+    try:
+        cloud_provider = CloudProvider.objects.get(
+            Q(
+                pk=cloud_provider_id,
+                cloud_sync_enabled=True,
+                cloud_sync_driver__isnull=False
+            ) & ~Q(cloud_sync_driver='')
+        )
+
+        processor = CLOUD_SYNC_DRIVERS[
+            cloud_provider.cloud_sync_driver
+        ]
+    except CloudProvider.DoesNotExist:
+        return HttpResponseNotFound()
+    except KeyError:
+        return HttpResponse('Specified processor is not available', status=501)
+
+    processor(cloud_provider, event_data)
+
+    return HttpResponse(status=204)
--- a/src/ralph/virtual/models.py
+++ b/src/ralph/virtual/models.py
@@ -40,6 +40,11 @@ class CloudProvider(AdminAbsoluteUrlMixin, NamedMixin):
         verbose_name = _('Cloud provider')
         verbose_name_plural = _('Cloud providers')
 
+    cloud_sync_enabled = models.BooleanField(
+        null=False, blank=False, default=False
+    )
+    cloud_sync_driver = models.CharField(max_length=128, null=True, blank=True)
+
 
 class CloudFlavor(AdminAbsoluteUrlMixin, BaseObject):
     name = models.CharField(_('name'), max_length=255)
--- a/src/ralph/virtual/processors/noop.py
+++ b/src/ralph/virtual/processors/noop.py
@@ -1,0 +1,9 @@
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def endpoint(cloud_provider, event_data):
+    """The endpoint for DC asset synchronisation that does nothing."""
+    logger.info('Received new DC asset synchronisation event. Doing nothing.')
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,9 @@ setup(
             'test_ralph = ralph.__main__:test',
             'validate_ralph = ralph.cross_validator.__main__:main',
         ],
+        'ralph.cloud_sync_processors': [
+            'noop=ralph.virtual.processors.noop:endpoint',
+        ],
     },
     classifiers=[
         'Development Status :: 4 - Beta',
--- a/src/ralph/admin/tests/tests_views.py
+++ b/src/ralph/admin/tests/tests_views.py
@@ -93,6 +93,7 @@ FACTORY_MAP = {
     'ralph.virtual.models.CloudHost': 'ralph.virtual.tests.factories.CloudHostFullFactory',  # noqa
     'ralph.virtual.models.CloudProject': 'ralph.virtual.tests.factories.CloudProjectFactory',  # noqa
     'ralph.virtual.models.CloudProvider': 'ralph.virtual.tests.factories.CloudProviderFactory',  # noqa
+    'ralph.virtual.models.CloudSyncProcessor': 'ralph.virtual.tests.factories.CloudSyncProcessorFactory',  # noqa
     'ralph.virtual.models.VirtualServer': 'ralph.virtual.tests.factories.VirtualServerFullFactory',  # noqa
     'ralph.virtual.models.VirtualServerType': 'ralph.virtual.tests.factories.VirtualServerTypeFactory',  # noqa
     'ralph.security.models.Vulnerability': 'ralph.security.tests.factories.VulnerabilityFactory',  # noqa
--- a/src/ralph/urls/base.py
+++ b/src/ralph/urls/base.py
@@ -51,6 +51,7 @@ urlpatterns = [
     url(r'^', include('ralph.admin.autocomplete_urls')),
     url(r'^dhcp/', include('ralph.dhcp.urls')),
     url(r'^deployment/', include('ralph.deployment.urls')),
+    url(r'^virtual/', include('ralph.virtual.urls')),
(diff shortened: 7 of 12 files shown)


src/ralph/virtual/processors/noop.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+
+LOGGER = logging.getLogger(__name__)
```
vi4m: logger = logging.getLogger(__name__)

src/ralph/settings/base.py
```diff
@@ -570,3 +570,7 @@ def str_to_bool(s: str) -> bool:
 STATSD_GRAPHS_PREFIX = 'ralph.graphs'
 
 TRANSITION_TEMPLATES = None
+# OpenStack event listeners
```
vi4m: What is this comment for?

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,40 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+ENTRY_POINTS_GROUP = 'ralph.dchost_processors'
+DCHOST_PROCESSORS = {}
+SUBSCRIPTIONS = {}
+
+
+def load_processors():
+    for ep in pkg_resources.iter_entry_points(ENTRY_POINTS_GROUP):
```
vi4m: s/ep/entry_point

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+
+
+def generate_listeners():
+    load_processors()
+
+    for subscription_config in settings.DCASSET_SYNC_ENDPOINTS:
+        sub_name = subscription_config['name']
+
+        # NOTE(romcheg): Work around the issue when ready() hook is called
```
vi4m: could you describe a little more what's happengine - or maybe provide issue link for this bug?

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,40 @@
+import logging
+
+import pkg_resources
+import pyhermes
+from django.conf import settings
+
+
+LOGGER = logging.getLogger(__name__)
```
vi4m: logger

src/ralph/virtual/processors/noop.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def endpoint(event_data):
+    """The endpoint for DC asset synchronisation that does nothing."""
+    LOGGER.info('Received new DC asset syncronisation event. Doing nothing.')
```
vi4m: synchronization

vi4m:
Any updates here?

src/ralph/virtual/models.py
```diff
@@ -35,11 +35,26 @@
 logger = logging.getLogger(__name__)
 
 
+class CloudSyncProcessor(models.Model, AdminAbsoluteUrlMixin):
```
mkurek: shouldn't `AdminAbsoluteUrlMixin`be inherited first?

src/ralph/apps.py
```diff
@@ -3,6 +3,8 @@
 
 from django.apps import AppConfig
 
+from ralph.virtual.subscriptions import generate_listeners
```
mkurek: wouldn't it be better to move this import into `ready` method and keep imports at the top of the file in `ralph.virtual.subscriptions`?

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,49 @@
...
+                'Could not import DC asset event processor from {}.'
+                ''.format(ep.module_name)
+            )
+
+
+def generate_listeners():
+    load_processors()
+
+    from ralph.virtual.models import CloudProvider
+    cloud_providers = CloudProvider.objects.filter(
```
mkurek: what if I add new cloud provider? why I'll have to restart ralph to create new subscriber?

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,49 @@
...
+import pyhermes
+from django.conf import settings
+from django.utils.text import slugify
+
+
+
+logger = logging.getLogger(__name__)
+
+
+ENTRY_POINTS_GROUP = 'ralph.dchost_processors'
```
mkurek: so is it dchost processor or cloud processor?

src/ralph/virtual/subscriptions.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def generate_listeners():
+    load_processors()
+
+    from ralph.virtual.models import CloudProvider
+    cloud_providers = CloudProvider.objects.filter(
+        sync_enabled=True,
+        sync_event_processor__isnull=False
+    ).all()
```
ar4s: please add ``.values_list('name', flat=True)`` to get only name as a flat list

src/ralph/virtual/subscriptions.py
ar4s: ``.select_related('sync_event_processor')``

src/ralph/virtual/subscriptions.py
romcheg: For readability purposes I have to leave it as is.

src/ralph/virtual/subscriptions.py
romcheg: For now yes -- it's necessary to restart Ralph in order to regenerate the list of listeners. The plan is to regenerate them on post_save signal.

src/ralph/virtual/subscriptions.py
romcheg: It's a cloud processor, I have to rename it.

src/ralph/virtual/subscriptions.py
romcheg: Done

src/ralph/virtual/models.py
romcheg: Done

src/ralph/virtual/subscriptions.py
romcheg: Right, for now it will be necessary to restart Ralph. I'm going to add a post_save signal subscriber in order to regenerate the list of subscribers every time something is done to a cloud provider. However, I'd rather did it in a separate patch.

mkurek: why couldn't you make one generic endpoint? or pass cloudprovider info as query param?

romcheg:
I have re-done the patch with a different approach -- now there is a custom Django view that routes sync events to appropriate cloud providers basing on URLs. Some tests were also added.

The patch was also rebased and squashed to avoid misunderstanding.

ar4s (dismissed):

ar4s (dismissed):

src/ralph/virtual/processors/noop.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def endpoint(cloud_provider, event_data):
+    """The endpoint for DC asset synchronisation that does nothing."""
+    logger.info('Received new DC asset syncronisation event. Doing nothing.')
```
mkurek: synchronisation

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,75 @@
...
+                logger.error(
+                    'Could not import DC asset event processor from {}.'
+                    ''.format(ep.module_name)
+                )
+
+
+@csrf_exempt
+@require_POST
+def cloud_sync_router(request, cloud_provider_id):
+    load_processors()
```
mkurek, line 54: can't we load it in `AppConfig.ready()` or sth like that?

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,75 @@
...
+
+    try:
+        cloud_provider = CloudProvider.objects.get(
+            pk=cloud_provider_id,
+            sync_enabled=True
+        )
+    except CloudProvider.DoesNotExist:
+        return HttpResponseNotFound()
+
+    CLOUD_SYNC_PROCESSORS[cloud_provider.sync_event_processor.module](
```
mkurek: what if `sync_event_processor` is not set?

src/ralph/virtual/models.py
```diff
@@ -35,11 +35,26 @@
 logger = logging.getLogger(__name__)
 
 
+class CloudSyncProcessor(AdminAbsoluteUrlMixin, models.Model):
```
mkurek: what's the point of this model? Couldn't we point processor by name in the url?

src/ralph/virtual/cloudsync.py
romcheg: Using AppConfig.ready() is bound with a few issues. One of them is that it's not even possible to import any object on a module level. In order to avoid that I have specifically re-written the patch to not use ready(). Loading the list of entry points on the first request won't hurt at all because it's won't bring any notable performance problems.

src/ralph/virtual/processors/noop.py
romcheg: Done

src/ralph/virtual/cloudsync.py
romcheg: Good point. I will add an extra DB filter for that.

src/ralph/virtual/models.py
romcheg: I wanter users to have a convenient admin pages where they can select processors from the list of available ones. On the other hand binding the field directly to the list of available entry points would require an automatic synchronisation of the data in the DB with the list of entry points.

It's a good point to extend the router with a check whether or not the specified processor is actually available.

MarekBleschke (dismissed):
looks ok

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def load_processors():
+    global CLOUD_SYNC_PROCESSORS
+
+    # NOTE(romcheg): Check if processors were loaded prior to locking anything.
+    if CLOUD_SYNC_PROCESSORS is not None:
+        return
+
+    with LOCK:
```
mkurek, line 32: why do you need this?

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,78 @@
...
+        # NOTE(romcheg): Double check in order to avoid race conditions.
+        if CLOUD_SYNC_PROCESSORS is not None:
+            return
+
+        logger.info('Loading cloud sync processors.')
+
+        CLOUD_SYNC_PROCESSORS = {}
```
mkurek: to avoid race condition, couldn't you use here some local variable, and at the end of this function, assign value of this local variable to `CLOUD_SYNC_PROCESSORS`?

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,78 @@
...
+            sync_enabled=True,
+            sync_event_processor__isnull=False
+        )
+        processor = CLOUD_SYNC_PROCESSORS[
+            cloud_provider.sync_event_processor.module
+        ]
+    except CloudProvider.DoesNotExist:
+        return HttpResponseNotFound()
+    except KeyError:
+        return HttpResponse('Specified processor is not available', status=501)
```
mkurek, line 78: doesn't 400 or 404 better fits here?

src/ralph/virtual/processors/noop.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def endpoint(cloud_provider, event_data):
```
mkurek, line 7: could you provide better name for this processor?

src/ralph/virtual/models.py
mkurek: But if processor is pointed in the url, user will "select" it in hermes console (by entering url). Right now you still have to sync data between db and code (ex. what if I change the name of the processor or remove one)? The compromise here would be storying `sync_event_processor` as string in `CloudProvider` model :)

src/ralph/virtual/cloudsync.py
```diff
@@ -0,0 +1,78 @@
...
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+
+from ralph.virtual.models import CloudProvider
+
+logger = logging.getLogger(__name__)
+
+
+ENTRY_POINTS_GROUP = 'ralph.cloud_sync_processors'
+CLOUD_SYNC_PROCESSORS = None
```
pnuckowski: To drop that global variable maybe use `nonlocal` ?
```python
def processor_factory():
  processors = None
  def inner():
    nonlocal processors
    if processors is None:
      processors = load_processors()
    return processors

get_processors = processor_factory()


# then in cloud_sync_router method
processors = get_processors()
```

Blejwi (approved):

mkurek (approved):

PR #3256, opened 2018-04-23
Add new model for SSL Certificates and management command.

smaciejs:


src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+    internal = _('CA ENT')
+
+
+class SSLCertificate(AdminAbsoluteUrlMixin, BaseObject):
+    certificate = models.CharField(
+        verbose_name=_('certificate name'),
+        help_text=_('Full certificate name'),
+        max_length=255
+    )
+
```
mkurek: please remove blank lines between fields definition

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+    ev = _('EV')
+    ov = _('OV')
+    dv = _('DV')
+    wildcard = _('Wildcard')
+    multisan = _('Multisan')
+    internal = _('CA ENT')
+
+
+class SSLCertificate(AdminAbsoluteUrlMixin, BaseObject):
+    certificate = models.CharField(
```
mkurek: the whole model name is SSL**Certificate** - couldn't this field be named just `name`?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+        help_text=_("Business contact person for a certificate")
+    )
+    technical_owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='certificates_technical_owner',
+        blank=True,
+        null=True,
+        help_text=_("Technical contact person for a certificate")
+    )
+    certificate_issued_by = models.ForeignKey(
```
mkurek: why not just `issued_by`?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+        AssetHolder,
+        blank=True,
+        null=True,
+        help_text=_("Company which receives certificate")
+    )
+
+    date_from = models.DateField(null=True, blank=True)
+
```
mkurek: rm blank lines

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+    date_to = models.DateField(null=False, blank=False)
+
```
mkurek: rm blank lines

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+    san = models.TextField(blank=True)
+    related_name = 'SAN'
+    help_text = _("All Subject Alternative Name")
```
mkurek: this should be part of field definition, isn't it?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,83 @@
...
+
+class SSLCertificateContract(
+    AdminAbsoluteUrlMixin,
+    PermByFieldMixin,
+    TimeStampMixin,
+    models.Model,
+):
+
+    certificate = models.ForeignKey(SSLCertificate)
+    price = models.DecimalField(
```
mkurek: what's the difference between cert price and contract price?

mkurek:
please merge all migrations into one (at the end of work on this PR)

mkurek (changes requested):

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def handle(self, *args, **options):
+        certs_dir = options['certs_dir']
+        for root, dirs, files in os.walk(certs_dir):
+            for filename in fnmatch.filter(files, '*.crt'):
+                san = ''
+                cert = None
+                extension = None
+                try:
+                    pem_data = open(os.path.join(root, filename)).read()
```
MarekBleschke: IMO it would be better to use `with` for opening files:
```python
with open(os.path.join(root, filename)) as f:
    pem_data = f.read()
```

That way you're not counting on garbage collector to realease file object: https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files.

src/ralph/ssl_certificates/models.py
smaciejs: Done

src/ralph/ssl_certificates/models.py
smaciejs: In point. I'll change it.

src/ralph/ssl_certificates/models.py
smaciejs: You are right. I'll change it.

src/ralph/ssl_certificates/models.py
smaciejs: done

src/ralph/ssl_certificates/models.py
smaciejs: done

src/ralph/ssl_certificates/models.py
smaciejs: Of cours it is. I'll make proper changes.

src/ralph/ssl_certificates/models.py
smaciejs: I really don't see any difference. Probably I wright it in some early step and completely forgot about it. I'll remove second "price".

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: It looks better. I'll implement it into.

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+                cert = None
+                extension = None
+                try:
+                    with open(os.path.join(root, filename)) as f:
+                        pem_data = f.read()
+                    cert = x509.load_pem_x509_certificate(
+                        pem_data.encode(), default_backend()
+                    )
+                except ValueError:
+                    print('{}/{} is not valid'.format(root, filename))
```
ar4s: use ``self.stderr.write`` (or ``logging`` module) instead of ``print`` for logging errors

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def handle(self, *args, **options):
+        certs_dir = options['certs_dir']
+        for root, dirs, files in os.walk(certs_dir):
+            for filename in fnmatch.filter(files, '*.crt'):
+                san = ''
+                cert = None
+                extension = None
+                try:
+                    with open(os.path.join(root, filename)) as f:
```
ar4s: move block responsible for read the certificate outside ``try`` block and you can add another ``try`` block and handle ``IOError`` and raise ``CommandError`` with an appropriate message

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+                    )
+                except ExtensionNotFound:
+                    pass
+
+                if extension and extension.value:
+                    san = extension.value.get_values_for_type(x509.DNSName)
+                issuer = cert.issuer.get_attributes_for_oid(
+                    NameOID.ORGANIZATION_NAME
+                )
+                domain = Path(filename).stem
```
ar4s: 👌

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,62 @@
...
+    certificate_type = models.PositiveIntegerField(
+        choices=CertificateType(),
+        default=CertificateType.ov.id,
+    )
+    business_owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='certificates_business_owner',
+        blank=True,
+        null=True,
+        help_text=_("Business contact person for a certificate")
```
ar4s: please don't mix ' and " for strings but is a niggling nuance and you don't need fix that

ar4s (changes requested):

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: I've add it

src/ralph/ssl_certificates/admin.py
```diff
@@ -0,0 +1,49 @@
...
+class SSLCertificateAdmin(AttachmentsMixin, RalphAdmin):
+    list_select_related = [
+        'technical_owner', 'business_owner',
+        'issued_by'
+    ]
+
+    list_filter = [
+        'name', 'certificate_type', 'business_owner',
+        'technical_owner', ('date_from', DateListFilter),
+        ('date_to', DateListFilter), 'issued_by',
```
mkurek, line 21: please keep consistent presence or absence of `,` after last item

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
+import datetime
```
mkurek: shouldn't this file be named `import_ssl_certificates.py`?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+
+    help = 'Import data to application from dir'
+
+    def add_arguments(self, parser):
+        parser.add_argument('certs_dir', type=str)
+
+    def handle(self, *args, **options):
+        certs_dir = options['certs_dir']
+        for root, dirs, files in os.walk(certs_dir):
+            for filename in fnmatch.filter(files, '*.crt'):
```
mkurek: shouldn't this regex (`*.crt`) be configurable?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+                except ExtensionNotFound:
+                    pass
+
+                if extension and extension.value:
+                    san = extension.value.get_values_for_type(x509.DNSName)
+                issuer = cert.issuer.get_attributes_for_oid(
+                    NameOID.ORGANIZATION_NAME
+                )
+                domain = Path(filename).stem
+                issuer_name = 'CA ENT'
```
mkurek: why? shouldn't this be configurable?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+                asset_holder = None
```
mkurek: do you need this statement?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,62 @@
...
+        AssetHolder,
+        blank=True,
+        null=True,
+        help_text=_("Company which receives certificate")
+    )
+    date_from = models.DateField(null=True, blank=True)
+    date_to = models.DateField(null=False, blank=False)
+    san = models.TextField(
+        blank=True,
+        help_text=_("All Subject Alternative Name"),
```
mkurek: name or names?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    price = models.DecimalField(
+        max_digits=10, decimal_places=2, default=0, null=True, blank=True,
```
mkurek: what's the point of null/blank in this field if default is 0?

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,62 @@
...
+        related_name='certificates_technical_owner',
+        blank=True,
+        null=True,
+        help_text=_("Technical contact person for a certificate")
+    )
+    issued_by = models.ForeignKey(
+        AssetHolder,
+        blank=True,
+        null=True,
+        help_text=_("Company which receives certificate")
```
mkurek: this help text is conflicting with field name (`issued_by`), isn't it?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Command(BaseCommand):
+
+    help = 'Import data to application from dir'
+
+    def add_arguments(self, parser):
+        parser.add_argument('certs_dir', type=str)
+
+    def handle(self, *args, **options):
```
mkurek: could you split it into multiple smaller functions?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: I've changed it.

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: (y)

src/ralph/ssl_certificates/models.py
smaciejs: OK

src/ralph/ssl_certificates/admin.py
smaciejs: Sorry for that. I'll fix it

src/ralph/ssl_certificates/models.py
smaciejs: I don't see conflict. Could you be more specific?

src/ralph/ssl_certificates/models.py
smaciejs: It's plural, in deed. I've corrected it

src/ralph/ssl_certificates/models.py
mkurek: as far as I understand it, issuer (issued_by) might be Digicert, thawte etc, but company which receives certificate looks to me like sth like Allegro, Github etc

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: As I try my code, I need this statement.

mkurek: why, what was the error? 3 lines below you overwrite it anyway..

src/ralph/ssl_certificates/models.py
smaciejs: Of course, my mistake. I all time thinking about "delivered". I've change it

src/ralph/ssl_certificates/models.py
smaciejs: I copy that from Ralph Licences Model. I fought, it should look like. Am I wrong?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: You have right, I can deleted it

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: If we have CA ENT certificate, certificate will by without this attribute. Thats why I wrote it in this way, otherwise we will have empty filed.

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: It could be configurable, but all our certs have this regex. Should I change it?

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: It should ;)

src/ralph/ssl_certificates/management/commands/importssl_certificates.py
smaciejs: As we talk, I'll try to do this in next pull request

src/ralph/ssl_certificates/models.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    technical_owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='certificates_technical_owner',
+        blank=True,
+        null=True,
+        help_text=_('Technical contact person for a certificate')
+    )
+    issued_by = models.ForeignKey(
+        AssetHolder,
```
mkurek: are you sure that `AssetHolder` is the right model here? We're using it for `property_of` like fields (for storying companies like Allegro and similar), ex. https://github.com/allegro/ralph/blob/685fbfc9adec31d577aa21f32f1d7bb1572521e0/src/ralph/supports/models.py#L112-L117

smaciejs: I see it. If I change it for:
issued_by= models.ForeignKey(
        Manufacturer,
        on_delete=models.PROTECT,
        blank=True,
        null=True,
    )

Then will be ok?

mkurek: yes

ar4s (approved):

mkurek (approved):

PR #3261, opened 2018-05-02
Fix build-depends for newer Debian

SesquipedalianDefenestrator:
Newer Debian uses the default-libmysqlclient-dev metapackage (which points to either MySQL or MariaDB) instead of libmysqlclient-dev, so accept either in Build-Depends.


SesquipedalianDefenestrator:
Test seems to have failed with a duplicate DB key for the psql back-end rather than anything related to the Debian package building or mysql.

mkurek (approved):

PR #3268, opened 2018-05-25
Make preoot configuration help text indepentent on settings

romcheg:
Fix the issue when Ralph reports a missing migration when its settings
change.


ar4s (dismissed):

src/ralph/deployment/models.py
```diff
@@ -87,7 +86,7 @@ def autocomplete_str(self):
 <br>  - ralph_instance (eg. '{ralph_instance}')
 <br>  - service_env (eg. 'Backup systems - prod')
 <br>  - service_uid (eg. 'sc-123')
-""".format(ralph_instance=settings.RALPH_INSTANCE.rstrip('/')).strip()  # noqa
+""".format(ralph_instance='http://ralph.example.com').strip()  # noqa
```
mkurek: you're fixing the effect, not the cause. Can't you do sth like [that](https://stackoverflow.com/a/29233569/761782) or move generating this help text to form field (not model field)? In my opinion we should provide proper ralph url in this help text, not example one.

mkurek (changes requested):

src/ralph/deployment/models.py
romcheg: Sounds sensible

MarekBleschke: I think it could be changed back to `settings.RALPH_INSTANCE.rstrip('/')` after the latest changes.

src/ralph/lib/mixins/fields.py
```diff
@@ -61,6 +61,63 @@ class NullableCharField(
...
+class NUMPFieldMixIn(object):
+
+    def __init__(self, fields_to_ignore, *args, **kwargs):
+        super(NUMPFieldMixIn, self).__init__(*args, **kwargs)
+        self.fields_to_ignore = fields_to_ignore
+
+    def deconstruct(self):
+        name, path, args, kwargs = super(NUMPFieldMixIn, self).deconstruct()
+
+        if not self.__class__.__mro__.index(NUMPFieldMixIn) > 1:
```
MarekBleschke: I think it would be good to add comment why we need to check this.

src/ralph/lib/mixins/fields.py
```diff
@@ -61,6 +61,63 @@ class NullableCharField(
...
+            # NOTE(romcheg): Exclude all fields that should not be concidered
+            #                when generating migrations.
+            kwargs = {
+                f: kwargs[f] for f in kwargs if f not in self.fields_to_ignore
+            }
+            path = '{}.{}'.format(NUMP.__module__, NUMP.__name__)
+
+            args = [self.base_class(*args, **kwargs)]
```
MarekBleschke: Maybe we could use names like `nump_args`, `nump_kwargs` here to additionally point that those are args/kwargs for NUMP?

src/ralph/deployment/models.py
romcheg: Done

src/ralph/lib/mixins/fields.py
romcheg: Done

src/ralph/lib/mixins/fields.py
romcheg: Done

MarekBleschke (approved):

src/ralph/lib/mixins/fields.py
```diff
@@ -61,6 +61,67 @@ class NullableCharField(
...
+    def __init__(self, fields_to_ignore, *args, **kwargs):
+        super(NUMPFieldMixIn, self).__init__(*args, **kwargs)
+        self.fields_to_ignore = fields_to_ignore
+
+    def deconstruct(self):
+        name, path, args, kwargs = super(NUMPFieldMixIn, self).deconstruct()
+
+        # NOTE(romcheg): Perform arguments substitution only for direct
+        #                descendants.
+        if not self.__class__.__mro__.index(NUMPFieldMixIn) > 1:
```
mkurek, line 78: why do you have to check this condition? why we only care about direct descendants?

src/ralph/lib/mixins/fields.py
```diff
@@ -61,6 +61,67 @@ class NullableCharField(
     pass
 
 
+# NOTE(romcheg): NUMP stands for No Useles Migrations Please.
+#                It is recommended to refrain from checking the other meaning
+#                in UD.
+class NUMPFieldMixIn(object):
```
mkurek, line 67: what's the benefit of all this magic below? Why it's better that simple inheriting from `models.TextField` and overwriting `deconstruct` method?

src/ralph/lib/mixins/fields.py
romcheg: Because it's only necessary to ignore specified fields for the class that is explicitly wrapped by users.

src/ralph/lib/mixins/fields.py
romcheg: Well, that would be basically the same but less generic. Actually you pointed me to the solution that looks like this one. Later I and @ar4s discussed that it would be nice to have a convenient decorator to avoid subclassing for every specific case.

mkurek (approved):
Maybe there is too much magic here, but in general, code looks ok.

PR #3271, opened 2018-06-15
email with attachments transition action

MarekBleschke:


ar4s (approved):

src/ralph/back_office/tests/test_models.py
```diff
@@ -464,6 +452,46 @@ def test_a_report_is_generated(self, mock_method):
         self.assertEqual(attachment.original_filename, correct_filename)
         self.assertEqual(attachment.file.read(), GENERATED_FILE_CONTENT)
 
+    def test_send_attachments_to_user_action_sends_email(self):
```
pnuckowski, line 455: Those two tests are similar, why not to use `ddt` and feed test with data (file content)?

MarekBleschke: They have different arguments in `send_attachments_to_user` and different assretion so I would need to make some `if` conditions to tests those two different things. IMO it's more readable to split the to two different cases.

PR #3272, opened 2018-06-19
display related virtual hosts

szymi-:
In Relations tab of DatacenterAsset, related virtual servers will be displayed.

### DIFF
--- a/src/ralph/data_center/views.py
+++ b/src/ralph/data_center/views.py
@@ -1,4 +1,7 @@
+from django.contrib.contenttypes.models import ContentType
+
 from ralph.admin.views.extra import RalphDetailView
+from ralph.virtual.models import VirtualServer
 
 
 class RelationsView(RalphDetailView):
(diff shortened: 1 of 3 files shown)


src/ralph/data_center/views.py
```diff
@@ -14,11 +14,17 @@ def _add_cloud_hosts(self, related_objects):
         if cloud_hosts:
             related_objects['cloud_hosts'] = cloud_hosts
 
+    def _add_virtual_hosts(self, related_objects):
+        virtual_hosts = list(self.object.children.all())
```
mkurek: blade could be child of DataCenterAsset as well - shouldn't we filter by type (content type?) of children?

szymi-: We should. I will add filtering.

mkurek (dismissed):

mkurek (commented):
Code is ok, but I'd appreciate some tests for `_add_virtual_hosts` and `_add_physical_hosts` :)

mkurek (approved):

PR #3269, opened 2018-05-30
SSL certs 2 step

smaciejs:
Add new management command to update data from DNSaaS

PR #3274, opened 2018-06-25
Add new fields in My equipment card RAS-1268

smaciejs:


smaciejs:
@ar4s could you add reviewers?

mkurek:
@smaciejs are all these commits valid in terms of this PR? a lot of them are commits from #3269 and the only ones reasonable here are 380e093 and 47f7a51 - am I right? If yes, please rebase this PR to contain only these two commits.

PR #3275, opened 2018-06-26
Add new fields in My equipment card RAS-1268

smaciejs:


src/ralph/accounts/admin.py
```diff
@@ -136,6 +136,30 @@ def escape_param(p):
...
+            placeholders = [
+                k[1] for k in Formatter().parse(url) if k[1] is not None
+            ]
+            item_dict.update({
+                k: getattr_dunder(item, k) for k in placeholders
+            })
+            if self.request and 'username' not in item_dict:
+                item_dict['username'] = self.request.user.username
+
+            def escape_param(p):
```
mkurek: why not use https://docs.python.org/3.6/library/urllib.parse.html#urllib.parse.urlencode ?

src/ralph/accounts/admin.py
```diff
@@ -136,6 +136,30 @@ def escape_param(p):
         return ''
     report_failure.title = ''
 
+    def report_buyout(self, item):
```
mkurek: his metod is almost exactly the same as `report_failure` - the only difference is formatted url and title - could you extract common content to another method?

mkurek (changes requested):

src/ralph/accounts/admin.py
smaciejs: I just do this in the same way as report_failure is done. Should i change it using urllib.parse.urlencode? In both cases?

src/ralph/accounts/admin.py
smaciejs: OK, I'll do this

src/ralph/accounts/admin.py
```diff
@@ -109,9 +109,9 @@ def user_licence(self, item):
         else:
             return []
 
-    def report_failure(self, item):
+    def create_report_link(self, url, url_title, item):
         item_dict = model_to_dict(item)
-        url = settings.MY_EQUIPMENT_REPORT_FAILURE_URL
+        url = url
```
MarekBleschke: it's not necessary

src/ralph/accounts/admin.py
```diff
@@ -127,15 +127,27 @@ def escape_param(p):
...
             return '<a href="{}" target="_blank">{}</a><br />'.format(
                 url.format(
                     **{k: escape_param(v) for (k, v) in item_dict.items()}
                 ),
-                _('Report failure')
+                _(url_title)
             )
         return ''
+
+    def report_failure(self, item):
```
MarekBleschke: IMO it's better to put report_failure and report_buyout above create_report_link, from readability perspective.
In this way you have more natural reading flow.

MarekBleschke (changes requested):

src/ralph/accounts/admin.py
smaciejs: Done

src/ralph/accounts/admin.py
smaciejs: Done

MarekBleschke (approved):

mkurek (approved):

PR #3278, opened 2018-07-03
[to close]Putted dictionaries from Back Office into Settings app tree RAS-526

smaciejs:


smaciejs:
@ar4s close it please

PR #3280, opened 2018-07-06
Add related hosts to api RAS-1267

m1kra:
Return the lists of CloudsHosts, VirtualServers and DataCenterAssets related with any DataCenterAsset in API.

### DIFF
--- a/src/ralph/data_center/api/serializers.py
+++ b/src/ralph/data_center/api/serializers.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
-from rest_framework import serializers
 from django.contrib.contenttypes.models import ContentType
+from rest_framework import serializers
 
 from ralph.api import RalphAPISerializer
 from ralph.api.serializers import RalphAPISaveSerializer
--- a/src/ralph/data_center/api/serializers.py
+++ b/src/ralph/data_center/api/serializers.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from django.contrib.contenttypes.models import ContentType
 from rest_framework import serializers
 
 from ralph.api import RalphAPISerializer
@@ -26,7 +25,6 @@ from ralph.data_center.models import (
     VIP
 )
 from ralph.security.api import SecurityScanSerializer
-from ralph.virtual.models import VirtualServer
 
 
 class ClusterTypeSerializer(RalphAPISerializer):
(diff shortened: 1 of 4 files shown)


src/ralph/data_center/tests/test_api.py
```diff
@@ -53,7 +55,7 @@ def setUp(self):
 
     def test_get_data_center_assets_list(self):
         url = reverse('datacenterasset-list')
-        with self.assertNumQueries(15):
+        with self.assertNumQueries(21):
```
mkurek: why number of queries has grown? Are you sure that it wouldn't grow linearly with more `DataCenterAsset`s on the list?

src/ralph/data_center/api/serializers.py
```diff
@@ -123,6 +125,34 @@ class DataCenterAssetSerializer(ComponentSerializerMixin, AssetSerializer):
     scmstatuscheck = SCMInfoSerializer()
     securityscan = SecurityScanSerializer()
 
+    related_hosts = serializers.SerializerMethodField()
+
+    def _get_physical_hosts(self, obj):
+        physical_server = ContentType.objects.get_for_model(DataCenterAsset)
+        return DataCenterAssetSimpleSerializer(
+            obj.children.filter(content_type=physical_server), many=True
```
mkurek: when you use filter, you are not using any prefetched data (by `prefetch_related` call) and you make here call(s) to the database foreach DataCenterAsset you're processing - I think you should filter out these objects in your code (not in database) or build prefetch new, filtered objects (for all `DataCenterAsset`s at once)

mkurek (changes requested):

m1kra:
@mkurek As you suggested, I have changed the way children are fetched from the database. I fetch all `VirtualServer`s and `DataCenterAsset`s  once in `DataCenterAssetViewSet` and pass them to the serialiser in context, overriding `get_serializer_context` method. I had been trying to do it using `prefetch_releted`, but I failed. I was also thinking of using `objects.all()` in serialiser, so that django would cache the request.

src/ralph/data_center/api/serializers.py
m1kra: Done.

src/ralph/data_center/tests/test_api.py
m1kra: That was because of using filter in serialiser, where it fetched data from database for each asset in the list.

src/ralph/data_center/tests/test_api.py
```diff
@@ -53,7 +55,7 @@ def setUp(self):
 
     def test_get_data_center_assets_list(self):
         url = reverse('datacenterasset-list')
-        with self.assertNumQueries(15):
+        with self.assertNumQueries(19):
```
mkurek: why 19, not 17?

src/ralph/data_center/api/serializers.py
```diff
@@ -123,6 +123,38 @@ class DataCenterAssetSerializer(ComponentSerializerMixin, AssetSerializer):
...
+            physical_hosts, many=True
+        ).data
+
+    def _get_virtual_hosts(self, obj):
+        if 'virtual_servers' not in self.context.keys():
+            return []
+        from ralph.virtual.api import VirtualServerSerializer
+        virtual_hosts = [vs for vs in self.context['virtual_servers']
+                         if vs.parent and vs.parent.pk == obj.pk]
+        return VirtualServerSerializer(virtual_hosts, many=True).data
```
mkurek: isn't there any simple serializer for `VirtualServer`?

mkurek (changes requested):

src/ralph/data_center/tests/test_api.py
m1kra: That was because of querying 'cloudhost_set' in Serializer, and not prefetching it in ViewSet.

src/ralph/data_center/api/serializers.py
m1kra: Done.

m1kra:
Now there are two additional queries - one for prefetching `children`, which comes from foreign key relation to `BaseObject` (field `parent`), and second for prefetching related `CloudHost`s (field `hypervisor`). If we didn't care about grouping those hosts in `related_hosts` JSON key, we could get rid of some code in serializer (see [this commit](https://github.com/m1kra/ralph/commit/08136e7bb3c376b8989d68fa21b48300aff4537b))

mkurek (approved):

PR #3282, opened 2018-07-09
Add cleanup security scans 

m1kra:
Transition action "cleanup security scans" should be available for VirtualServer and CloudHost, RAS-1215.

### DIFF
--- a/src/ralph/security/apps.py
+++ b/src/ralph/security/apps.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+
+from ralph.apps import RalphAppConfig
+
+
+class SecurityConfig(RalphAppConfig):
+    name = 'ralph.security'
+    verbose_name = 'Security'
+
+    def get_load_modules_when_ready(self):
+        """
+        security/transitions.py will be loaded alongside modules reurned by
+        super().get_load_modules_when_ready() when the app is ready.
+        This will add transition actions to appropriate models.
+        """
+        return super().get_load_modules_when_ready() + ['transitions']
--- a/src/ralph/security/apps.py
+++ b/src/ralph/security/apps.py
@@ -9,7 +9,7 @@ class SecurityConfig(RalphAppConfig):
 
     def get_load_modules_when_ready(self):
         """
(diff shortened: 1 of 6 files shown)


src/ralph/virtual/models.py
```diff
@@ -267,6 +268,18 @@ def ip_addresses(self, value):
             base_object=self, ipaddress__address__in=to_delete
         ).delete()
 
+    @classmethod
+    @transition_action(
+        verbose_name=_('Cleanup security scans'),
+    )
+    def cleanup_security_scans(cls, instances, **kwargs):
```
mkurek: the same method is defined already for `DataCenterAsset`. Please extract it to some common class or put somewhere as plain function - `transition_action` has additional argument - `models` - see https://github.com/allegro/ralph/blob/ng/src/ralph/deployment/deployment.py for examples.

vi4m:
@m1kra description is required for all pull requests (can re-paste from RAS-1215)

m1kra:
@vi4m Thank you, I have added the description.

@mkurek I have moved `cleanup_security_scans` to a new file in security app and added config loading it in `AppConfig.ready()`, following the approach in the deployment module. 
I am not quite sure if security module is the best location for it, but I find this solution logical. I was also thinking of adding it to transition app or base assets, so the additional config wouldn't be necessary.

mkurek (dismissed):

src/ralph/security/apps.py
```diff
@@ -0,0 +1,16 @@
...
+from ralph.apps import RalphAppConfig
+
+
+class SecurityConfig(RalphAppConfig):
+    name = 'ralph.security'
+    verbose_name = 'Security'
+
+    def get_load_modules_when_ready(self):
+        """
+        security/transitions.py will be loaded alongside modules reurned by
```
ar4s: reurned -> returned

ar4s (changes requested):

src/ralph/security/apps.py
m1kra: Done.

mkurek (approved):

ar4s (approved):

Issue #3259, opened 2018-04-24
Adding Virtual Server with Hypervisor with ServiceEnv


Chez:
As Above, when you add a virtual server with a Hypervisor set as a Service Environment:

    File "/home/vagrant/ralph/src/ralph/virtual/admin.py", line 156, in get_parent

    obj.parent.get_absolute_url(), obj.parent.hostname

    AttributeError: 'ServiceEnvironment' object has no attribute 'hostname'

[console ready]
>>> obj.parent.__dict__
{'modified': datetime.datetime(2018, 4, 24, 8, 52, 37, 972512), 'content_type_id': 26, 'service_env_id': None, 'parent_id': None, '_state': <django.db.models.base.ModelState object at 0x7f8eef976e80>, 'configuration_path_id': None, 'remarks': '', 'id': 1, 'environment_id': 1, 'baseobject_ptr_id': 1, 'created': datetime.datetime(2018, 4, 24, 8, 52, 37, 972479), 'service_id': 1  }
>>> 

If adding a Service Environment to a Hypervisor, then why are they listed as options as they do not have hostnames.

vi4m:
Which version of ralph is it?

Chez:
Ralph 3

vi4m:
Debian package? what does `dpkg -l "*ralph*"` show?

romcheg:
Closing a stale issue. If it is still relevant, please re-open it or start a topic at  Ralph forum: https://ralph.discourse.group

Issue #3236, opened 2018-02-22
ralph/virtual/admin.py in get_parent, line 168


ghost:
when we go to this url in ralph3 /virtual/virtualserver/ and create an virtual server.
and then we go to the listing "again to /virtual/virtualserver/" we got above error.

and this in de trace-back. i think this is a bug because we can simulate this with a fresh install.
see the trace-back below.
----------------------------------------

Request Method: | GET
-- | --
http://192.168.50.181:8000/virtual/virtualserver/
1.8.18
AttributeError
'ServiceEnvironment' object has no attribute 'hostname'
/opt/ralph/ralph-core/lib/python3.4/site-packages/ralph/virtual/admin.py in get_parent, line 168
/opt/ralph/ralph-core/bin/python
3.4.3
['/opt/ralph/ralph-core/bin',  '/opt/ralph/ralph-core/lib/python3.4',  '/opt/ralph/ralph-core/lib/python3.4/plat-x86_64-linux-gnu',  '/opt/ralph/ralph-core/lib/python3.4/lib-dynload',  '/usr/lib/python3.4',  '/usr/lib/python3.4/plat-x86_64-linux-gnu',  '/opt/ralph/ralph-core/lib/python3.4/site-packages',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf',  '/opt/ralph/ralph-core/lib/python3.4/site-packages/odf']
Thu, 22 Feb 2018 11:02:23 +0100

Chez:
I have also gotten the same error after adding a virtual server and then clicking virtual servers from the Data center Menu. Is this a recognised bug or a new bug?

Thathwamasi:
Not sure as this is a proper solution:
When you try to change the below line in python this resolved the issue
on /opt/ralph/ralph-core/lib/python3.4/site-packages/ralph/virtual/admin.py in get_parent, line 168


From :
 return '<a href="{}">{}</a>'.format(
            obj.parent.get_absolute_url(), obj.parent.hostname

TO:
 return '<a href="{}">{}</a>'.format(
[...]

PR #3284, opened 2018-07-12
Fix adding hypervisors

m1kra:
This pull request adds validation of assigned VM hypervisor, such that it can only be a ~~virtual server~~ cloud host or data center asset. This resolves issues #3259 and #3236 . Also see RAS-1181.

### DIFF
--- a/src/ralph/virtual/admin.py
+++ b/src/ralph/virtual/admin.py
@@ -56,27 +56,23 @@ class VirtualServerTypeForm(RalphAdmin):
 
 class VirtualServerForm(RalphAdminForm):
     HYPERVISOR_TYPE_ERR_MSG = 'Hypervisor must be one of \
-                                   DataCenterAsset or VirtualServer.'
-    HYPERVISOR_IS_SELF_ERR_MSG = 'Hypervisor and host must be different.'
+                                   DataCenterAsset or CloudHost'
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if 'parent' in self.fields:
             self.fields['parent'].required = True
-            self.fields['parent'].validators.extend((
-                self.validate_parent_type,
-                self.validate_parent_instance
-            ))
 
-    def validate_parent_type(self, value):
-        virtual_server = ContentType.objects.get_for_model(VirtualServer)
+    def clean_parent(self):
+        value = self.cleaned_data.get('parent')
+        self._validate_parent_type(value)
+        return value
+
+    def _validate_parent_type(self, value):
         dc_asset = ContentType.objects.get_for_model(DataCenterAsset)
-        if value.content_type not in (virtual_server, dc_asset):
+        cloud_host = ContentType.objects.get_for_model(CloudHost)
+        if value.content_type not in (dc_asset, cloud_host):
             raise ValidationError(_(self.HYPERVISOR_TYPE_ERR_MSG))
-
-    def validate_parent_instance(self, value):
-        if self.instance.pk == value.pk:
-            raise ValidationError(_(self.HYPERVISOR_IS_SELF_ERR_MSG))
 
     class Meta:
         labels = {'parent': _('Hypervisor')}
--- a/src/ralph/virtual/admin.py
+++ b/src/ralph/virtual/admin.py
@@ -55,8 +55,9 @@ class VirtualServerTypeForm(RalphAdmin):
 
 
 class VirtualServerForm(RalphAdminForm):
-    HYPERVISOR_TYPE_ERR_MSG = 'Hypervisor must be one of \
-                                   DataCenterAsset or CloudHost'
+    HYPERVISOR_TYPE_ERR_MSG = _(
+        'Hypervisor must be one of DataCenterAsset or CloudHost'
+    )
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


src/ralph/virtual/admin.py
```diff
@@ -52,11 +54,30 @@ class VirtualServerTypeForm(RalphAdmin):
...
+    HYPERVISOR_TYPE_ERR_MSG = 'Hypervisor must be one of \
+                                   DataCenterAsset or VirtualServer.'
+    HYPERVISOR_IS_SELF_ERR_MSG = 'Hypervisor and host must be different.'
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if 'parent' in self.fields:
             self.fields['parent'].required = True
 
+    def is_valid(self):
```
mkurek: isn't it possible to use (create) `validate_parent` method?

src/ralph/virtual/admin.py
```diff
@@ -52,11 +54,30 @@ class VirtualServerTypeForm(RalphAdmin):
...
+        valid = super().is_valid()
+        virtual_server = ContentType.objects.get_for_model(VirtualServer)
+        dc_asset = ContentType.objects.get_for_model(DataCenterAsset)
+        selected_object = self.cleaned_data['parent']
+        if selected_object.content_type not in (virtual_server, dc_asset):
```
mkurek: why do we allow to VM be parent of another VM?

src/ralph/virtual/admin.py
```diff
@@ -52,10 +55,28 @@ class VirtualServerTypeForm(RalphAdmin):
...
 class VirtualServerForm(RalphAdminForm):
+    HYPERVISOR_TYPE_ERR_MSG = 'Hypervisor must be one of \
+                                   DataCenterAsset or VirtualServer.'
+    HYPERVISOR_IS_SELF_ERR_MSG = 'Hypervisor and host must be different.'
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if 'parent' in self.fields:
             self.fields['parent'].required = True
+            self.fields['parent'].validators.extend((
```
mkurek: sorry for the confusion - I mean `clean_parent` - if you name a method like that, [it should work out of the box](https://docs.djangoproject.com/en/2.0/ref/forms/validation/), without adding it "manually" to validators

src/ralph/virtual/admin.py
m1kra: Done, now it's a lot cleaner.

src/ralph/virtual/admin.py
m1kra: Well, I have changed that to CloudHost. As you have said, `DataCenterAssets` are added as cloud hosts for Ironic.

src/ralph/virtual/tests/test_admin.py
```diff
@@ -0,0 +1,86 @@
...
+        self.assertTrue(form.is_valid())
+
+    def test_invalid_hypervisor_type(self):
+        form_data = self._get_basic_form_data()
+        for obj in (
+            BaseObjectFactory(),
+            ServiceEnvironmentFactory(),
+            VirtualServerFactory(),
+            LicenceFactory()
+            ):
```
mkurek: untab this line

src/ralph/virtual/admin.py
```diff
@@ -52,11 +55,25 @@ class VirtualServerTypeForm(RalphAdmin):
 
 
 class VirtualServerForm(RalphAdminForm):
+    HYPERVISOR_TYPE_ERR_MSG = 'Hypervisor must be one of \
```
mkurek: wouldn't it be better to translate it here?

src/ralph/virtual/admin.py
```diff
@@ -52,11 +55,25 @@ class VirtualServerTypeForm(RalphAdmin):
...
         if 'parent' in self.fields:
             self.fields['parent'].required = True
 
+    def clean_parent(self):
+        value = self.cleaned_data.get('parent')
+        self._validate_parent_type(value)
+        return value
+
+    def _validate_parent_type(self, value):
+        dc_asset = ContentType.objects.get_for_model(DataCenterAsset)
```
mkurek: you might use `ContentType.get_for_models`

m1kra:
I have been trying to solve it differently - filter out only valid hosts from ui. I didn't want to modify base object, adding `limit_choices_to`, so I tried doing it in the form. The field `parent` is `raw_id_field` in admin, and simple modification of queryset didn't work. I have tried adding custom `ForeignKeyRawIdWidget`, which is used by `raw_id_fields`, but it doesn't render correctly. I think I could try to solve that, but I would need some time to figure out how to do it.

src/ralph/virtual/admin.py
m1kra: Done.

src/ralph/virtual/tests/test_admin.py
```diff
@@ -0,0 +1,86 @@
...
+        return True
+
+
+class MockRequest:
+    pass
+
+
+class TestVirtualServerForm(RalphTestCase):
+    
+    def setUp(self):
```
mkurek: check out these places for inspiration how to make requests to admin in tests:
https://github.com/allegro/ralph/blob/ng/src/ralph/data_center/tests/test_admin.py#L29-L34
https://github.com/allegro/ralph/blob/ng/src/ralph/data_center/tests/test_admin.py#L69-L77

src/ralph/virtual/admin.py
m1kra: Done.

src/ralph/virtual/tests/test_admin.py
m1kra: Done.

src/ralph/virtual/tests/test_admin.py
m1kra: Done.

src/ralph/virtual/admin.py
m1kra: Done.

mkurek (approved):

PR #3292, opened 2018-07-26
Fix bower using environment variables

romcheg:


szymi- (approved):

romcheg:
Merging with one Review in order to unblock builds

PR #3293, opened 2018-07-27
transition for accepting back office assets for loan

MarekBleschke:


src/ralph/admin/templatetags/dashboard_tags.py
```diff
@@ -47,6 +47,20 @@ def get_user_equipment_to_accept_tile_data(user):
     }
 
 
+def get_user_equipment_to_accept_loan_tile_data(user):
+    from ralph.accounts.helpers import \
```
mkurek: why do you import it in function, not at the top of the module?

src/ralph/accounts/helpers.py
```diff
@@ -9,20 +10,38 @@
...
-        id=ACCEPTANCE_TRANSITION_ID
+        id=transition_id
     ).exists()
 
 
-def get_assets_to_accept(user):
+acceptance_transition_exists = partial(
+    transition_exists, ACCEPTANCE_TRANSITION_ID
+)
+loan_transition_exists = partial(
```
mkurek, line 26: is it used anywhere?

src/ralph/admin/templatetags/dashboard_tags.py
MarekBleschke: I was basing this on code for accepting equipment. Actually it seems to work with importing this  on module level so I've moved it there.

src/ralph/accounts/helpers.py
MarekBleschke: Good catch :)
I've missed part in accepting equipment on "my equipment" page. Added changes there for loan so this function is now used there.

mkurek:
ok for me, but travis has some complaints :)

MarekBleschke:
@mkurek fixed tests, travis is happy again :)

mkurek (approved):

ar4s (approved):

PR #3295, opened 2018-08-02
Fix duplicate hostname

m1kra:
If two IPs have the same hostname in DHCP, one is ignored. This PR adds validation of such scenario when "expose in DHCP" is selected while creating or updating IP address.

### DIFF
--- a/src/ralph/networks/models/networks.py
+++ b/src/ralph/networks/models/networks.py
@@ -6,9 +6,7 @@ import struct
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
-from django.contrib.contenttypes.models import ContentType
 from django.db import models, transaction
-from django.db.models import Q, Case, When
 from django.db.models.signals import post_migrate, pre_save
 from django.db.utils import ProgrammingError
 from django.dispatch import receiver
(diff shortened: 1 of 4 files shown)


src/ralph/networks/models/networks.py
```diff
@@ -662,6 +662,32 @@ class Meta:
     def __str__(self):
         return self.address
 
+    def hostname_is_unique_in_dc(self, hostname, dc):
+        from ralph.dhcp.models import DHCPEntry
+        entries_with_hostname = DHCPEntry.objects.filter(
+            hostname=hostname,
+            dhcp_expose=True,
```
mkurek: no need to filter it - it's already handled by DHCPEntry's manager

src/ralph/networks/models/networks.py
```diff
@@ -662,6 +662,32 @@ class Meta:
     def __str__(self):
         return self.address
 
+    def hostname_is_unique_in_dc(self, hostname, dc):
+        from ralph.dhcp.models import DHCPEntry
+        entries_with_hostname = DHCPEntry.objects.filter(
```
mkurek: shouldn't we exclude here current ip address (by its pk)?

mkurek (approved):

PR #3302, opened 2018-08-16
Securityscan transition fix

szymi-:


mkurek:
We'll need some test to confirm that it's working (by it I mean generating data for dc host update event after cleaning securityscan).

src/ralph/assets/api/serializers.py
```diff
@@ -437,9 +437,18 @@ class ComponentSerializerMixin(NetworkComponentSerializerMixin):
     processors = ProcessorSimpleSerializer(many=True, source='processor_set')
 
 
+class SecurityScanField(serializers.Field):
+
+    def to_representation(self, value):
+        if getattr(value, 'pk'):
+            return SecurityScanSerializer().to_representation(value)
+        else:
```
mkurek: what is (type of) value here? could we reverse this condition into sth like:
```
if not value:  # or if value is None
    return None
else:
    return SecurityScanSerializer().to_representation(value)
```

src/ralph/data_center/tests/test_publishers.py
```diff
@@ -0,0 +1,38 @@
+from ralph.data_center.publishers import _get_host_data
+from ralph.security.tests.factories import SecurityScanFactory
+from ralph.tests import RalphTestCase
+from ralph.virtual.tests.factories import VirtualServerFactory
+
+
+class PublisherTests(RalphTestCase):
+
+    def test_get_host_data(self):
```
mkurek: please rename it to `test_get_host_data_after_deleting_securityscan`

src/ralph/assets/api/serializers.py
szymi-: It could be None or a SecurityScan instance. I will add an extra check at the top to see if it is not none and then do as you say.

src/ralph/data_center/tests/test_publishers.py
szymi-: Sure.

src/ralph/assets/api/serializers.py
```diff
@@ -437,9 +437,19 @@ class ComponentSerializerMixin(NetworkComponentSerializerMixin):
     processors = ProcessorSimpleSerializer(many=True, source='processor_set')
 
 
+class SecurityScanField(serializers.Field):
+
+    def to_representation(self, value):
+        if value:
```
mkurek: maybe one condition will be enough here:
```
if value and value.pk:
    return SecurityScanSerializer().to_representation(value)
```
?

mkurek (approved):

PR #3307, opened 2018-08-31
Zero values as default in openstack_sync summary

szymi-:


src/ralph/virtual/management/commands/openstack_sync.py
```diff
@@ -57,11 +57,16 @@ def __str__(self):
         repr(self.value)
 
 
+class Summary(defaultdict):
+    def __getitem__(self, key):
+        return dict.get(self, key, 0)
```
mkurek: what's the difference between this and just `defaultdict(int)`?

szymi-: It returns 0 if key does not exist instead of raising KeyError

mkurek: Yes, but thats how defaultdict works:

```
>>> from collections import defaultdict
>>> d = defaultdict(int)
>>> d['non-existing-key']
0
```

mkurek (approved):

PR #3310, opened 2018-09-19
improve docker integration

dmitry-kr:


Dockerfile
```diff
@@ -1,54 +1,44 @@
...
+COPY src/ralph/static/src/ ./src/ralph/static/src/
+COPY src/ralph/admin/static/ ./src/ralph/admin/static/
+RUN npm install --silent -g bower && \
+    npm install --silent -g gulp && \
+    npm install --silent && \
+    bower --allow-root install && \
+    gulp build
+
+
+FROM python:3.4-stretch
```
romcheg: Ralph is compatible with Python 3.6 which is the version the previous base image (ubuntu bionic) supplied. I'd recommend sticking to that version.

anchiru: okay I'll check and update Dockerfile. tnx

dmitry-kr: okay I'll check and update Dockerfile. tnx

vi4m:
Don't merge it yet - need to discuss it internally.

romcheg:
@dmitry-kr sorry for not leaving any feedback on your PR. We were working on re-thinking our packaging options from scratch. Currently official docker images we provide use bare ubuntu base to which our official Debian package is installed. Since Ralph is not being built inside a container anymore, the changes you propose are no longer required. Thank you for your ideas and sorry for not keeping you up do date.

PR #3311, opened 2018-09-20
more generic createsuperuser.py

managai:
I would to automatize ralph installation on Ubuntu 18.04

this little change allow to create superuser even if ralph user is not yet configured by executing
python3 docker/createsuperuser.py

### DIFF
--- a/docker/createsuperuser.py
+++ b/docker/createsuperuser.py
@@ -7,8 +7,14 @@ django.setup()
 
 from django.contrib.auth import get_user_model
 print('Updating superuser info')
-u = get_user_model().objects.get(username='ralph')
-u.set_password('ralph')
-u.is_superuser = True
-u.is_staff = True
-u.save()
+try:
+    u = get_user_model().objects.get(username='ralph')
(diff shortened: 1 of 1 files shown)


docker/createsuperuser.py
```diff
@@ -7,8 +7,14 @@
 
 from django.contrib.auth import get_user_model
 print('Updating superuser info')
-u = get_user_model().objects.get(username='ralph')
-u.set_password('ralph')
-u.is_superuser = True
-u.is_staff = True
-u.save()
+try:
+    u = get_user_model().objects.get(username='ralph')
```
ar4s, line 11: try 
```
user, created = get_user_model().objects.get_or_create(
    username ='ralph',
    defaults={
        'is_superuser': True,
        'is_staff': True
    },
)
```

ar4s (changes requested):

romcheg:
Docker image was re-done from scratch. Proposed changes were included in a slightly different form. Thanks @managai !

PR #3312, opened 2018-09-21
prevent wrong asset model type assignment

szymi-:


src/ralph/back_office/admin.py
```diff
@@ -58,7 +59,8 @@ class BackOfficeAssetLicenceInline(RalphTabularInline):
     inlines = [BackOfficeAssetLicenceInline]
 
 
-class BackOfficeAssetAdminForm(RalphAdmin.form):
+class BackOfficeAssetAdminForm(AssetFormMixin, RalphAdmin.form):
+    MODEL_TYPE = 'back_office'
```
ar4s: use ``ObjectModelType.back_office`` instead of string

src/ralph/data_center/forms.py
```diff
@@ -5,10 +5,12 @@
 
 from ralph.admin.mixins import RalphAdminForm
 from ralph.data_center.models.physical import DataCenterAsset
+from ralph.lib.mixins.forms import AssetFormMixin
 from ralph.networks.models import IPAddress
 
 
-class DataCenterAssetForm(RalphAdminForm):
+class DataCenterAssetForm(AssetFormMixin, RalphAdminForm):
+    MODEL_TYPE = 'data_center'
```
ar4s: ditto

ar4s (changes requested):

src/ralph/lib/mixins/forms.py
```diff
@@ -94,3 +95,24 @@ def clean(self, value):
 
     def to_python(self, value):
         return value
+
+
+class AssetFormMixin(forms.ModelForm):
+    MODEL_TYPE = None
```
ar4s, line 101: IMO ``None`` is not valid value - consider some validation

ar4s (changes requested):

src/ralph/lib/mixins/forms.py
szymi-: What would you consider sufficient validation? Would checking if MODEL_TYPE is not None be enough?

ar4s (approved):

PR #3317, opened 2018-10-02
Intellectual property

smaciejs:


mkurek:
@smaciejs could you first try to rename model `TradeMarksLinkedDomains` to `TradeMarksLinkedDomain`? Django should ask you if you renamed this model and should prepare proper migration automatically - https://github.com/allegro/ralph/blob/981875c8bf1c26710d5d72de97ef0af2a9215d02/src/ralph/trade_marks/models.py#L110

smaciejs:
@mkurek  ok, I'll chagne it

mkurek:
migration 0001 was already merged - we should not modify it - there should be 0002 migration which just rename the model (and maybe some fields)

src/ralph/trade_marks/migrations/0002_auto_20181004_1417.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('domains', '0006_auto_20180725_1216'),
+        ('trade_marks', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
mkurek: didn't django ask about renaming existing model? :) we should use `RenameModel` here (and `RenameField` below as well )

https://docs.djangoproject.com/en/1.8/ref/migration-operations/#renamemodel
https://docs.djangoproject.com/en/1.8/ref/migration-operations/#renamefield

mkurek (changes requested):

src/ralph/trade_marks/migrations/0002_auto_20181004_1417.py
smaciejs: It didn't ask me about it. I'll do this in a proper way

mkurek:
@smaciejs the tests are not passing - shouldn't you rename some fields as well?

mkurek (approved):

MarekBleschke (approved):

PR #3321, opened 2018-10-09
Custom message for transitions

ar4s:


src/ralph/back_office/models.py
```diff
@@ -606,10 +607,11 @@ def release_report(cls, instances, requester, transition_id, **kwargs):
     def send_attachments_to_user(cls, requester, transition_id, **kwargs):
         if kwargs.get('attachments'):
             transition = Transition.objects.get(pk=transition_id)
+            context_func = get_hook(__name__, 'transition_email_context')
```
MarekBleschke: I think it's better to be explicit here and set module_name parameter as explicit string not relying on __name__.

src/ralph/back_office/hooks.py
```diff
@@ -0,0 +1,11 @@
+from collections import namedtuple
+
+EmailContext = namedtuple('EmailContext', 'subject body')
```
mkurek: i guess we want to be still compatible with python3.4?

src/ralph/lib/hooks.py
```diff
@@ -0,0 +1,21 @@
+import pkg_resources
+
+from django.conf import settings
+from django.utils import lru_cache
+
+
+@lru_cache.lru_cache(maxsize=None)
+def get_hook(module_name, name, variant=None):
```
MarekBleschke: IMO it would be nice to put here docstring documenting how to use this function and how to setup it in settings.

src/ralph/back_office/hooks.py
```diff
@@ -0,0 +1,11 @@
+from collections import namedtuple
```
mkurek: why hook(s)? isn't it just some (utility) function?

src/ralph/admin/management/commands/check_entry_points.py
```diff
@@ -0,0 +1,26 @@
...
+    def handle(self, *args, **kwargs):
+        self.stdout.write("Entry points:")
+        for key, active_variant in settings.ENTRY_POINTS_CONFIGURATION.items():
+            self.stdout.write("\t{}:".format(key))
+            for ep in pkg_resources.iter_entry_points(key):
+                ending = '\n'
+                try:
+                    ep.load()
+                except ImportError as e:
+                    ending = self.style.ERROR(" (error: {})\n".format(e))
```
mkurek: nice :)

src/ralph/back_office/models.py
mkurek: do we have to use module name here? if I move this piece of code to another module, it'll magically stop working. Can't we rely only on `name` (second param)?

src/ralph/settings/base.py
```diff
@@ -601,3 +601,9 @@ def str_to_bool(s: str) -> bool:
 STATSD_GRAPHS_PREFIX = 'ralph.graphs'
 
 TRANSITION_TEMPLATES = None
+
+ENTRY_POINTS_CONFIGURATION = {
```
mkurek: isn't it configuration for hooks?

src/ralph/lib/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+    loaded_func = None
+    full_name = '.'.join([module_name, name])
+
+    if variant is None:
+        variant = settings.ENTRY_POINTS_CONFIGURATION[full_name]
+    for entry_point in pkg_resources.iter_entry_points(full_name):
+        if variant == entry_point.name:
+            loaded_func = entry_point.load()
```
mkurek: what if `ImportError` will be raised there?

ar4s: I will write some checks using `System check framework` from Django

src/ralph/settings/base.py
ar4s: Yes, it is configuration for hooks. Should I change name? ``HOOKS_CONFIGURATION``?

src/ralph/back_office/models.py
ar4s: The reason why I used `__name__` is easy way to navigate where this hooks is used. I can change to fixed name.

docs/development/addons.md
```diff
@@ -112,6 +112,58 @@ class ExtraView(RalphDetailView):
...
+    ...
+)
+
+```
+
+
+### Available hooks
+| Name                                          | Description                                         |
+| --------------------------------------------- | --------------------------------------------------- |
+| `back_office.transition_action.email_context` | returns subject and body based on transition's name |
```
Blejwi: Maybe could we add new column with arguments that are passed to given context and expected return type/format? So a person who wants override some method is not forced to dive into source code?

Blejwi (dismissed):

src/ralph/lib/hooks/main.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Callable, Optional
```
mkurek, line 1: `typing` module is available since python3.5, but shouldn't ralph still work on python3.4?

ar4s: I added ``typing`` module to requirements.

Blejwi (dismissed):

Blejwi (approved):

PR #3325, opened 2018-10-18
Autocomplete endpoint to require authentication

romcheg:


Blejwi (dismissed):

ar4s (approved):

ar4s (changes requested):
Please add some test.

romcheg:
@ar4s done

src/ralph/admin/tests/tests_autocomplete.py
```diff
@@ -60,3 +61,17 @@ def test_not_or_empty(self):
...
+            kwargs={
+                'app': 'assets',
+                'field': 'service_env',
+                'model': 'BaseObject'
+            }
+        ) + '?q=foobar'
+        resp = self.client.get(url)
+
+        self.assertEqual(302, resp.status_code)
+        self.assertIn('login/?next=', resp.url)
```
ar4s: Did you consider ``assertRedirects``?

src/ralph/admin/tests/tests_autocomplete.py
```diff
@@ -60,3 +61,17 @@ def test_not_or_empty(self):
...
+            'autocomplete-list',
+            kwargs={
+                'app': 'assets',
+                'field': 'service_env',
+                'model': 'BaseObject'
+            }
+        ) + '?q=foobar'
+        resp = self.client.get(url)
+
+        self.assertEqual(302, resp.status_code)
```
Blejwi: `statuses` may be used here instead of hardcoded integer, it's more verbose to use constant IMO

src/ralph/admin/tests/tests_autocomplete.py
romcheg: Didn't know about it. Thanks!

src/ralph/admin/tests/tests_autocomplete.py
romcheg: After applying @ar4s's suggestion there's no need to check for a code here.

MarekBleschke (approved):

Blejwi (approved):

ar4s (approved):

PR #3329, opened 2018-10-25
Store cloud client config in cloud providers

romcheg:


src/ralph/virtual/forms.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class CloudProviderForm(RalphAdminFormMixin, forms.ModelForm):
+    class Meta:
+        model = CloudProvider
+        fields = (
+            'name', 'cloud_sync_enabled', 'cloud_sync_driver', 'client_config'
+        )
+
+    def clean_client_config(self):
```
Blejwi, line 17: I would consider using JSON field from django.extras package

Blejwi (changes requested):

src/ralph/virtual/forms.py
romcheg: Makes sense. However, I will have to keep the custom form because that class does not implement any validation.

MarekBleschke (dismissed):

mkurek (changes requested):
why do we need client_config, why is has to be encrypted and where it'll be used?

Blejwi (dismissed):

romcheg:
@mkurek That is a configuration for client tools. Currently it is stored in a configuration file as a huge JSON one-liner. I would like to move this configuration to closer to cloud providers so cloud administrators can adapt cloud configuration to the state of cloud providers in Ralph.

Another advantage of keeping these settings bound to cloud providers is that they can be easily accessible for, say, cloud synchronization.

romcheg:
Accidentally closed wrong PR :)

src/ralph/virtual/models.py
```diff
@@ -44,6 +46,13 @@ class Meta:
         null=False, blank=False, default=False
     )
     cloud_sync_driver = models.CharField(max_length=128, null=True, blank=True)
+    client_config = encrypt(
+        JSONField(
+            blank=True,
+            null=True,
+            verbose_name='client configuration',
```
mkurek, line 53: how the user should know what should he type into this field? text, json? if json, what keys?

mkurek (changes requested):

src/ralph/virtual/models.py
romcheg: The validator ensures an input is always a valid JSON. I would not force any restrictions for the content because the latter depends on the needs for every specific client. I think it's better to add the description of the configuration for every specific client to the documentation.

romcheg: Alternatively, it's possible to use YAML instead of JSON. However, that would be more error-prone due to YAML being dependent on whitespaces.

mkurek: Could you add some example and/or config for specific clients to help_text and/or docs?

romcheg: Will do

MarekBleschke: @romcheg how is it going with docs?

mkurek (approved):

PR #3338, opened 2018-11-07
Add docker-compose-dev file for developers

sekuranka:


docker-compose-dev.yml
```diff
@@ -0,0 +1,24 @@
...
+    MYSQL_DATABASE: ralph_ng
+    MYSQL_ROOT_PASSWORD: ralph_ng
+    MYSQL_USER: ralph_ng
+    MYSQL_PASSWORD: ralph_ng
+  volumes_from:
+    - data
+  command: mysqld --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
+redis:
+  image: redis:3.0
+  restart: always
```
ar4s: It's not good idea for development purpose.

ar4s (changes requested):

docker-compose-dev.yml
romcheg: +1, better stick to the no restart policy for development environments.

@sekuranka "no" is a default value for "restart", so you can just remove these "restart" parameters.

docker-compose-dev.yml
```diff
@@ -0,0 +1,24 @@
+db:
```
romcheg: I would also recommend to specify the version of the compose file explicitly and then stick to it's format, i. e.,

```suggestion
version: "3"
services:
  db:
```

You can look at [this example](https://docs.docker.com/compose/compose-file/#compose-file-structure-and-examples)

romcheg (changes requested):

docker-compose-dev.yml
```diff
@@ -0,0 +1,24 @@
...
+    MYSQL_PASSWORD: ralph_ng
+  volumes_from:
+    - data
+  command: mysqld --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
+redis:
+  image: redis:3.0
+  restart: always
+  ports:
+    - "6379:6379"
+data:
```
mkurek: there is no need to use data container anymore: https://docs.docker.com/compose/compose-file/#volumes

docker-compose-dev.yml
sekuranka: done

docker-compose-dev.yml
sekuranka: done

Blejwi (dismissed):

ar4s (dismissed):

romcheg (dismissed):
LGTM

docker-compose-dev.yml
```diff
@@ -0,0 +1,18 @@
...
+  image: mysql:5.7
+  ports:
+    - "3306:3306"
+  environment:
+    MYSQL_DATABASE: ralph_ng
+    MYSQL_ROOT_PASSWORD: ralph_ng
+    MYSQL_USER: ralph_ng
+    MYSQL_PASSWORD: ralph_ng
+  volumes:
+    - ./volumes/static:/opt/static
```
mkurek: do we really need to map this explicit to host filesystem? Why not just `- /opt/static`?

docker-compose-dev.yml
```diff
@@ -0,0 +1,18 @@
+db:
+  image: mysql:5.7
```
mkurek: another version of mysql? Why not 5.5 or 5.6 which are tested on travis?

romcheg: Looks like it's time to change travis because 5.7 is what we specify in out [production-ready docker-compose](https://github.com/allegro/ralph/blob/ng/contrib/docker-compose.yml.tmpl) in contrib/ :)

sekuranka: done

romcheg (approved):

Blejwi (approved):

PR #3340, opened 2018-11-09
Apache2 wsgi intergration on Debian 9

xsgb:
Hello,

That's what I've done to build the packages on debian stretch and load Ralph-ng behind Apache2 using mod_wsgi.

I've choosen to load the configuration form SetEnv directives and pass it from WSGI to django using this method: https://exceptionshub.com/access-apache-setenv-variable-from-django-wsgi-py-file.html

What do you think about that ? Isn't optimal, but for me it's really useful in our environment.

### DIFF
--- a/contrib/ralph.conf.apache2-wsgi
+++ b/contrib/ralph.conf.apache2-wsgi
@@ -1,0 +1,39 @@
+<IfModule mod_ssl.c>
+	<VirtualHost _default_:443>
+		ServerName  ralph.example.com
+		ServerAlias ralph
+		ServerAdmin webmaster@localhost
+		ErrorLog ${APACHE_LOG_DIR}/error.log
+		CustomLog ${APACHE_LOG_DIR}/access.log combined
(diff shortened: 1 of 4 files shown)


vi4m:
Thanks for contribution. I think it's something we want to incorporate into Ralph. 
I'm thinking about better way of handling settings environment, though. Maybe external file would be better in this case.

romcheg:
@xsgb we have re-done our deb package so no it's much easier to achieve what you intend to. Would you like to do it?

xsgb:
Super nice, no we can close the ticket! I think a yaml config file will be impeccable.

romcheg:
@xsgb after you install the new deb package, ralph's configuration files will be read from /etc/ralph.conf and /etc/ralph/conf.d/*.conf. The files there should be simple key=value text files, so no yaml is required.

xsgb:
Okay, didn't checked, pretty neat to know. Thanks you very much.

PR #3346, opened 2018-11-26
Buyout button in MyEquipmet view

smaciejs:


src/ralph/accounts/admin.py
```diff
@@ -120,18 +120,25 @@ def user_licence(self, item):
         else:
             return []
 
+    def buyout_ticket(self, item):
+        get_params = {"inventory_number": item.barcode,
```
mkurek: format it like that:
```
get_params = {
    "inventory_number": item.barcode,
    "serial_number": item.sn,
    ...
}
```

src/ralph/accounts/admin.py
```diff
@@ -120,18 +120,25 @@ def user_licence(self, item):
...
+                      "serial_number": item.sn,
+                      "model": item.model,
+                      "comment": item.buyout_date
+                      }
+        url = "?".join(
```
mkurek, line 130: wouldn't it be better to use `string.format` to create this url?

src/ralph/accounts/admin.py
smaciejs: Done

src/ralph/accounts/admin.py
smaciejs: with @ar4s we decide, that this form look better, but I can change it. Should I?

Blejwi (approved):

szymi- (approved):

PR #3360, opened 2018-12-28
New landing page

ar4s:
Preview:
landing-page


ar4s:
Update:
fireshot capture 010 - ralph - http___127 0 0 1_4000_

vi4m (changes requested):
1. Please change Damian -> Damian Blejwas in JSON. 
2. Replace entire current section "Flexible" with: 

Flexible

Ralph is built on top of Django and Python 3 and is easy to extend and customize without writing boilerplate code. 
REST API, Workflows code extensions allow for easy customization.

Furthermore, if you need to write something completely new - there are ready to use components - base models with permissions, regions, actions, custom fields and tags and even - PDF printouts.

vi4m (approved):

PR #3364, opened 2019-01-07
Match status if possible when converting asset type

szymi-:


src/ralph/back_office/helpers.py
```diff
@@ -10,3 +15,36 @@ def get_email_context_for_transition(transition_name: str) -> EmailContext:
         'body': 'Please see documents provided in attachments for "{}".'.format(transition_name)  # noqa
     }
     return EmailContext(**default)
+
+
+def status_converter(
+    old_status_id, target_status_id, from_status_cls=BackOfficeAssetStatus,
```
MarekBleschke: I think it would be good to add some assert to check if `from_status_cls` and `to_status_cls` are one of expected classes. It's easy to miss description in dcostring.

src/ralph/back_office/helpers.py
```diff
@@ -10,3 +15,36 @@ def get_email_context_for_transition(transition_name: str) -> EmailContext:
...
+        old_status_id: BO status (int)
+        target_status_id: ID of default status assigned to transition
+        from_status_cls: Source status class
+        to_status_cls: Destination status class
+
+    Returns:
+        new_status_id: DC status (int)
+    """
+    try:
+        if not target_status_id:
```
MarekBleschke: IMO it's better to move all code that can't cause exception outside of `try` block.

MarekBleschke (changes requested):

src/ralph/back_office/helpers.py
szymi-: I was considering a flag (maybe `reverse`?) that would specify from what type of status to what type of status to convert and then basing on that flag, set from_status_cls and to_status_cls inside the method. Default would be from BackOfficeAssetStatus to DataCenterAssetStatus, and `reverse` would be the other way round. What do you think?

romcheg: I can suggest to make this method internal and create two explicit wrappers for both cases that will have the params hardcoded.

MarekBleschke (approved):

romcheg (approved):
LGTM, thanks!

PR #3365, opened 2019-01-08
Network min and max IP, block update via API

### LINKED ISSUE #3143: min_ip, max_ip fields erroring despite being specified via JSON REST API

Blejwi:
Resolves #3143.

### DIFF
--- a/src/ralph/networks/api.py
+++ b/src/ralph/networks/api.py
@@ -34,6 +34,13 @@ class NetworkSimpleSerializer(RalphAPISerializer):
         )
 
 
+class NetworkSaveSerializer(RalphAPISerializer):
+    class Meta:
+        model = Network
+        depth = 1
+        exclude = ('min_ip', 'max_ip')
+
+
 class NetworkSerializer(RalphAPISerializer):
     class Meta:
         model = Network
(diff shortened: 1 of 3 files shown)


romcheg (dismissed):
LGTM

src/ralph/networks/api.py
```diff
@@ -94,14 +101,19 @@ def destroy(self, request, *args, **kwargs):
...
     queryset = Network.objects.all()
-    serializer_class = NetworkSerializer
     select_related = ['network_environment', 'kind']
     prefetch_related = ['racks']
     extended_filter_fields = {
         # workaround for custom field for address field defined in admin
         'address': ['address'],
     }
 
+    def get_serializer_class(self):
```
mkurek: use `serializer_class` and `save_serializer_class` class attributes instead

Blejwi: Fixed

mkurek (approved):

fabiankrack:
Thx a lot for the quick fix!

PR #3367, opened 2019-01-08
from_email in EmailContext

maho:
extend EmailContext by from_email field, allowing sending from hooks with different senders.


Makefile
```diff
@@ -13,7 +14,7 @@ upload-package:
 
 install-js:
 	npm install
-	bower install
+	# bower install
```
Blejwi: Is it commented by an accident?

mkurek (approved):

MarekBleschke (approved):

PR #3369, opened 2019-01-09
inkpy in docker-compose

maho:
Easy run inkpy and let it connect to the same redis which is in settings


contrib/ralph.env
```diff
@@ -6,7 +6,7 @@ DATABASE_HOST=db
 STORE_SESSIONS_IN_REDIS=1
 REDIS_HOST=redis
 REDIS_PORT=6379
-REDIS_DB=3
```
vi4m: Why was it 3 before?

docker-compose-dev.yml
```diff
@@ -18,3 +18,11 @@ services:
     image: redis:3.0
     ports:
       - "6379:6379"
+  inkpy:
+    image: ar4s/inkpy_jinja:latest
```
mkurek: use allegro's image: https://hub.docker.com/r/allegro/inkpy

mkurek (changes requested):

docker-compose-dev.yml
maho: just did it.

contrib/ralph.env
maho: I believe it was some remains after some testing. I think that it should be consistent, so same port in env as in settings. Otherwise it won't run.

maho:
merged master again, to avoid conflicts

mkurek (approved):

MarekBleschke (approved):

PR #3373, opened 2019-01-10
RAS-1338 - Better wording of back office front boxes

maho:
- better wording
- few pl translation
- instruction in readme how to see the boxes


README.md
```diff
@@ -88,3 +88,44 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
 ## Getting help
 
 Ralph community will answer your questions on a chat: [![Gitter](https://img.shields.io/gitter/room/gitterHQ/gitter.svg)](https://gitter.im/allegro/ralph?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+
+# Dev Tips & Tricks
+
+## how to see asset boxes in front dashboard
+
```
vi4m: One paragraph of intro required - for example: 

```

By default, when you visit front front dashboard, ... missing. 

To overcome this problem, ...```

File: README.md [hunk 2]
```diff
@@ -88,3 +88,55 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
...
+By default, when you visit front dashboard, you see no "Hardware loan", "Hardware release" nor "Hardware return" boxes. They should be visible only when you have assigned asset in back office, which is in 'in progress' or 'in use' state. However - when you add such item on empty database, you will ecounter error: 
+
+```
+NoReverseMatch at /
```
mkurek: couldn't we make this working out-of-the-box? The solution below is too complex (or even too complicated) in my opinion.

README.md
```diff
@@ -88,3 +88,55 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
...
+    ```
+    mysql> SELECT * FROM transitions_transition;
+    +----+------+--------+--------+----------+--------------------+--------------------+---------------+-------------+
+    | id | name | source | target | model_id | async_service_name | run_asynchronously | template_name | success_url |
+    +----+------+--------+--------+----------+--------------------+--------------------+---------------+-------------+
+    |  1 | foo  | ["2"]  | 4      |        1 |                    |                  0 |               | NULL        |
+    +----+------+--------+--------+----------+--------------------+--------------------+---------------+-------------+
+    1 row in set (0,00 sec)
+    ```
+* write following lines into settings/dev.py:
```
mkurek: there is no info, that the value for those settings should be the one I fetched from db

mkurek (approved):

mkurek (changes requested):
-

README.md
maho: I admit, it would be awesome. However, sounds like "big thing" and subject of separate issue. 

This too complicated instruction is better than WTF feeling when you ecounter "NoReverseMatch" :).

README.md
maho: I think it is there, but I just made it more visible.

mkurek (dismissed):

README.md
```diff
@@ -88,3 +88,58 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
 ## Getting help
 
 Ralph community will answer your questions on a chat: [![Gitter](https://img.shields.io/gitter/room/gitterHQ/gitter.svg)](https://gitter.im/allegro/ralph?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+
+# Dev Tips & Tricks
```
MarekBleschke: I don't think it's good place to add such info here. IMO would be better to move it somewhere to `ralph/docs/development`.

MarekBleschke (changes requested):
Just one comment about moving added docummentation to different place.

Blejwi (approved):

MarekBleschke (approved):

PR #3374, opened 2019-01-11
ralphctl fixed sourcing of conf.d/*

engelant:
soource /etc/ralph/conf.d/* didn't work, @rmocheg provided solution on gitter

### DIFF
--- a/contrib/common/usr/bin/ralphctl
+++ b/contrib/common/usr/bin/ralphctl
@@ -3,7 +3,9 @@
 set -aue
 
 source /etc/ralph/ralph.conf
-source /etc/ralph/conf.d/*
+for f in /etc/ralph/conf.d/*.conf; do
+    source $f
+done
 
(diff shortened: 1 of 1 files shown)


contrib/common/usr/bin/ralphctl
```diff
@@ -3,7 +3,9 @@
 set -aue
 
 source /etc/ralph/ralph.conf
-source /etc/ralph/conf.d/*
+for f in /etc/ralph/conf.d/*.conf; do
+    source $f
```
romcheg: Since the content of conf.d is never known it's safer to source files like:

```suggestion
    source "$f"
```

romcheg (changes requested):
Mentioned one minor nit inline. Otherwise LGTM, thanks!

romcheg (approved):

szymi- (approved):

PR #3376, opened 2019-01-14
Added SIM Card info to My Equipment List

sekuranka:


src/ralph/accounts/admin.py
```diff
@@ -208,6 +209,19 @@ def url(self, item):
     url.title = _('Link')
 
 
+class AssignedSimcardsList(Table):
+
+    def url(self, item):
```
romcheg: Regular users with regular permissions won't be able to go any of their sim cards. It's basically the same issue as happens when someone clicks on "go to asset" from My Equipment. I would suggest to remove the link.

P. S. It's fine, If removing the link here will also remove the "go to asset link".

src/ralph/sim_cards/admin.py
```diff
@@ -42,6 +43,7 @@ class SIMCardAdmin(RalphAdmin):
             )
         }),
     )
+    search_fields = ('card_number', 'phone_number', 'pin1', 'puk1')
```
romcheg: Perhaps, it would be reasonable to also display carriers.

romcheg (changes requested):
Left two comments inline.

src/ralph/accounts/admin.py
sekuranka: I removed url. It's not necessary in this case.

src/ralph/sim_cards/admin.py
sekuranka: I added carriers

MarekBleschke (approved):

src/ralph/accounts/views.py
```diff
@@ -220,6 +225,17 @@ def get_context_data(self, **kwargs):
             request=self.request,
         )
 
+        context['simcard_list'] = AssignedSimcardsList(
+            self.get_simcard_queryset(),
+            [
+                ('phone_number', _('Phone Number')),
+                ('card_number', _('Card Number')),
+                ('pin1', _('PIN 1'))
```
romcheg, line 233: Perhaps puk1 and carrier name could be useful for users as well.

szymi-: I agree, I bet pin is not something users will have to check for, especially that users will be likely to change it on the physical card. Most likely it will be puk that they will need, after entering wrong pin 3 times.

szymi- (approved):

romcheg (approved):

PR #3378, opened 2019-01-15
RAS-1327| Add field "start_usage" in asset model

smaciejs:


MarekBleschke (dismissed):

src/ralph/assets/models/assets.py
```diff
@@ -454,6 +454,7 @@ class Asset(AdminAbsoluteUrlMixin, BaseObject):
         null=True,
         blank=True,
     )
+    start_usage = models.DateField(blank=True, null=True)
```
MarekBleschke: On second thought I would add some help text about this field.

MarekBleschke (dismissed):

src/ralph/assets/models/assets.py
smaciejs: Fixed.

MarekBleschke (approved):

szymi- (approved):

PR #3393, opened 2019-01-24
Add new field "IMEI2"

look77:
I have a request to add new field in assets named "IMEI2"
Now most of phones is dualsim and have two numbers of IMEI


Blejwi:
Could you rebase your changes to ng branch? Seems that some old changes are here by accident :)

PR #3395, opened 2019-01-24
Switch to the new deb repository

romcheg:


Blejwi (approved):

contrib/common/apt/ralph.list
```diff
@@ -1 +1 @@
-deb https://dl.bintray.com/vi4m/ralph bionic main
+deb https://dl.bintray.com/allegro/debng bionic main
```
szymi-, line 1: debng - where does the name come from?

romcheg: deb - new generation

romcheg: The old repo "deb" has a deprecated layout so we have created a new one.

szymi- (approved):

PR #3401, opened 2019-01-30
Give credits to Łukasz and Radomir :)

vi4m:
Give credits to the (old fathers)  ❤️ team :)  🥇

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,3 +88,7 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
 ## Getting help
 
 Ralph community will answer your questions on a chat: [![Gitter](https://img.shields.io/gitter/room/gitterHQ/gitter.svg)](https://gitter.im/allegro/ralph?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+## Acknowledgements
+
+Ralph 3 is a from-scratch rewrite of original project created by small team including Łukasz Langa, and Radomir Dopieralski which was further developed by Marcin Kliks's teams. Old codebase is still available, for reference, under branch: [historical]


romcheg:
Please don't use anything non-ASCII either in the commit message title or the commit message body.

romcheg (changes requested):
.

README.md
```diff
@@ -88,3 +88,7 @@ Visit our documentation on [readthedocs.org](http://ralph-ng.readthedocs.org)
 ## Getting help
 
 Ralph community will answer your questions on a chat: [![Gitter](https://img.shields.io/gitter/room/gitterHQ/gitter.svg)](https://gitter.im/allegro/ralph?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
+## Acknowledgements
+
+Ralph 3 is a from-scratch rewrite of original project created by small team including Łukasz Langa, and Radomir Dopieralski which was further developed by Marcin Kliks's teams. Old codebase is still available, for reference, under branch: [historical]
```
romcheg, line 94: Frankly I'm not sure how to feel about this. For sure the founding fathers were important for this project. However, Ralph is what it is because of a combination of factors like the core team, contributors outside of that core team and the rest of the community from and outside Allegro that in one way or another helps shaping and sculpting Ralph into one of the best DCIM available.

While individual contributions are important, they aren't the complete story; and a story only makes sense when it's told as a whole. Therefore it would only make sense to acknowledge every single contribution (code or non-code) every individual has ever made to reflect their true value. With that in mind I propose to refrain from hiding the real story of Ralph, the DCIM, behind three names to be fair to code contributors, users, testers, translators and other members of the community.

9b5, line 94: fralychen

vi4m: Ok, I agree with your argumentation to include whole story.

But I think we should keep common standard to include this information in form of top-level file/files in repository(e.g. AUTHORS file which includes core team, contributors, and thanks for inspiration).  Currently this information is buried somewhere in docs showed on 3-rd party site server.

 zrzut ekranu 2019-01-31 o 11 51 38

vi4m: ?

romcheg: Just for the information -- my opinion on this topic is not strong (except the non-ASCII characters in the commit message -- that opinion is concrete-strong), so I will let the rest of the maintainers to say their word.

Re: AUTHORS file -- IMO, it's better to rename it to MAINTAINERS and keep it up to date with current maintainers names so the community members know whom to poke on Gitter.

vi4m:
Will do another PR for that matter.

PR #3403, opened 2019-02-05
Docker installation docs

szymi-:


szymi-:
@romcheg please take a look at this PR when you get a chance.

mkurek (approved):

romcheg:
I have some serious concerns regarding the approach. We need to discuss that. Please do not merge yet

5t0ne:
Could you sort out the issues already? As the current version is working like that, either the update should be merged, or the whole approach should be pulled back. If not final, you could leave the (experimental) tag.

Issue #3077, opened 2017-05-22
LDAP Authentication with OpenLDAP and Ralph3


ni-cwinkelmann:
Hi there,

before I start to pull my hair off - I should really start to ask for help. So here I am and really thankful for any advice.

## What do I want to archive?
User Authentication using OpenLDAP

## What did I do?
Followed the manual from [here](https://ralph-ng.readthedocs.io/en/latest/installation/configuration/).

### modified ralph.settings.prod.py
    ...
    from ralph.accounts.ldap import MappedGroupOfNamesType

    from django_auth_ldap.config import LDAPSearch
    from django_auth_ldap.config import GroupOfNamesType
    from django_auth_ldap.config import PosixGroupType

    ...

    # LDAP Config - Server
    AUTH_LDAP_START_TLS = True
    AUTH_LDAP_SERVER_URI = "ldap://ldap.company.domain"
    AUTH_LDAP_BIND_DN = "uid=ldap-account,ou=systems,ou=people,dc=company,dc=domain"
    AUTH_LDAP_BIND_PASSWORD = "mypassword"
    AUTH_LDAP_PROTOCOL_VERSION = 3

    # LDAP Config - User
    AUTH_LDAP_USER_USERNAME_ATTR = "uid"
    AUTH_LDAP_USER_OBJECT_CLASS = "posixAccount"
    AUTH_LDAP_USER_SEARCH_BASE = "ou=people,dc=company,dc=domain"
    AUTH_LDAP_USER_SEARCH_FILTER = '(&(objectClass={0})({1}=%(user)s))'.format(
        AUTH_LDAP_USER_OBJECT_CLASS,
        AUTH_LDAP_USER_USERNAME_ATTR
    )

    AUTH_LDAP_USER_SEARCH = LDAPSearch(
        AUTH_LDAP_USER_SEARCH_BASE,
        ldap.SCOPE_SUBTREE,
        AUTH_LDAP_USER_SEARCH_FILTER
    )
    AUTH_LDAP_USER_ATTR_MAP = {
        "first_name": "givenName",
        "last_name": "sn",
        "email": "mail",
        "employee_id": "employeeNumber",
        "location": "roomNumber",
        "city": "st",
        "company": "o",
        "cost_center": "CostCenter",
    }

    # LDAP Config - Groups
    AUTH_LDAP_MIRROR_GROUPS = True
    AUTH_LDAP_GROUP_TYPE = MappedGroupOfNamesType(name_attr="cn")
[...]

PR #3405, opened 2019-02-07
Support Debian Buster and MariaDB

CaptainSifff:
Curerntly the debian scripts prevent a build on Debian Buster.
Among the notable changes are:
- Debian Buster ships mariadb instead of mysql, hence some package names have changed
- Python-3.7 is the default.


romcheg:
@CaptainSifff Ralph is not tested against python-3.7 and MariaDB and most likely the tests will fail. Could you please verify this?

CaptainSifff:
Sure, I can check. does ralphctl expose some kind of testsuite?
I had to fiddle a bit with the package version in requirements.txt (rq >= 0.12 and django-rq >= 1.2.0)
But so far, I found that I can populate the database with the demo data and play with 
that dataset through the browser without noticing any issues so far.

CaptainSifff:
I tested some of the ralphctl commands:
ralphctl demodata: seems to work - database is populated and looks like I think it is expected.
ralphctl sqlflush - lots of output, no error
ralphctl ldap_sync - fails, currently not built with ldap(comes next)
ralphctl push_graphs_to_statd - no output, no error
ralphctl check - System check identified no issues (0 silenced).
ralphctl compilemessages - no error
ralphctl createcachetable - no error, no output
ralphctl dbshell - Mariadb shell preset to the ralph_ng database
ralphctl shell - Python3.7 shell
ralphctl test - Creating test database for alias 'default'...
Got an error creating the test database: (1044, "Access denied for user 'ralph_ng'@'localhost' to database 'test_ralph_ng'")
Which is expected that In the spirit of the installation docs I only gave permission to acces the database ralph_ng
ralphctl rqstats - Connection refused, somehow expected since I didn't set up redis

Anything else that you are interested in?

CaptainSifff:
I also got ldap and openldap sufficiently working for my needs by following the steps from the doc:
https://ralph-ng.readthedocs.io/en/latest/installation/configuration/
I also needed the patch from #3080 for patching ldap.py and the documentation given in 
#3077 for patching prod.py.

romcheg:
@CaptainSifff thank you for the research you have made. Before we start shipping the package for debian, we should enable automatic tests for 3.7 and mariadb. A spare pair of hands (you :) ) would be helpful for that atm. Let's discuss details in our Gitter channel.

requirements/base.txt
```diff
@@ -7,7 +7,7 @@ django-filter==0.13.0
 django-import-export==0.4.2
 django-mptt==0.8.7
 django-reversion==1.8.6
-django-rq==0.9.0
+django-rq>=1.2.0
```
romcheg: I would recommend to freeze the version to 1.2.0 to avoid random failures when django-rq or one of its dependencies becomes incompatible with Ralph. Since Ralph is not a library but an application, freezing dependencies does not hurt anyone.

requirements/base.txt
```diff
@@ -24,7 +24,7 @@ python-dateutil==2.4.2
 pytz==2015.4
 redis==2.10.3
 requests==2.20.0
-rq==0.5.6
+rq>=0.12
```
romcheg: ditto

romcheg (changes requested):
Left a few comments inline. Could you also please change mysql to mysql/mariadb in the [debconf](https://github.com/allegro/ralph/blob/5ed85d35d692392dc434ac32919555fa4ef3c3bf/debian/ralph-core.templates#L9)? You may want to test that after introducing the changes -- to do that build the package using ```make build-snapshot-package``` command and then install the package somewhere.

requirements/base.txt
CaptainSifff: OK, will do. I chose those versions since at those releases the libraries became compatible with Python-3.7 (async is a keyword now). Do you happen to know whether it is possible to request those versions only in case of Python3.7?

romcheg: It's possible to do  requirements/py37.txt and requirements/py-other.txt but I would rather test everything with the same set of requirements than build more different builds of Ralph.

szymi- (dismissed):

requirements/base.txt
CaptainSifff: OK, that's how we have it now.

CaptainSifff:
I also modified now the template as pointed out by @romcheg so that mariadb should be a possibility.

debian/ralph-core.templates
```diff
@@ -6,7 +6,7 @@ Description: Use default database settings:
 
 Template: ralphsrv/database_engine
 Type: select
-Choices: mysql, postgresql
+Choices: mysql, postgresql, mariadb
```
romcheg, line 9: there are references for ```ralphsrv/database_engine``` in ```ralph-core.config``` and ```ralph-core.postinst```. It is necessary to make appropriate changes at least in ```ralph-core.config```

romcheg (changes requested):

requirements/base.txt
CaptainSifff: should be resolved

debian/ralph-core.templates
CaptainSifff: should be resolved.

vi4m:
@romcheg Do we wait for something here? is it RTM?

romcheg:
@vi4m the changes do not allow to build the package for Buster within the existing build scripts, so we cannot merge this.

CaptainSifff:
Can somebody remind me what I would have to check...? The pipeline still goes through or did I miss a point?

PR #3413, opened 2019-02-17
SSL Certificates API

vi4m:


Blejwi (dismissed):

src/ralph/ssl_certificates/api.py
```diff
@@ -0,0 +1,36 @@
...
+        'service_env__service__uid',
+        'service_env__service__name',
+        'service_env__service__id',
+    ]
+    select_related = [
+        'service_env__service', 'service_env__environment'
+    ]
+
+
+class SaveSSLCertificate(RalphAPISaveSerializer):
```
MarekBleschke: I think it's not used anywhere.

MarekBleschke (changes requested):

vi4m:
@MarekBleschke removed unneded class

MarekBleschke (approved):

PR #3414, opened 2019-02-18
extend entrypoint to allow ralphctl

szymi-:


docker/provision/docker-entrypoint.sh
```diff
@@ -51,6 +52,9 @@ case "$ARG" in
     upgrade)
         "${RALPH_LOCAL_DIR}/init-ralph.sh" upgrade
         ;;
+    ralphctl)
```
romcheg, line 55: Isn't it goint to do the same thing as 

`docker run --entrypoint ralphctl <container> <extra_args>
`
?

szymi-: Yes, but before that it will load config from the config files.

romcheg: The line above will also read configuration files -- that's what ralphctl does.

szymi-: I need the stuff that happens from line 40 of this file to have db connection configured for instance.

romcheg: ralphctl does that on its own: https://github.com/allegro/ralph/blob/63316f4d9938c42cebe88ebdc1210a829b4b1dae/contrib/common/usr/bin/ralphctl#L5

szymi-: Ok, then this is uselss.

PR #3415, opened 2019-02-18
RAS-1390|Add Registring Institiotion and Country into Trade Marks model

smaciejs:


src/ralph/trade_marks/models.py
```diff
@@ -118,6 +142,11 @@ class TradeMark(Regionalizable, AdminAbsoluteUrlMixin, BaseObject):
         related_name='+',
         through='TradeMarksLinkedDomains',
     )
+    registering_institution = models.ForeignKey(
```
szymi-: Maybe `registrar` would be better? More concise and probably more natural. Google search for registering institution brings up no institution, only instructions on how to register an institution :)

src/ralph/trade_marks/models.py
```diff
@@ -118,6 +142,11 @@ class TradeMark(Regionalizable, AdminAbsoluteUrlMixin, BaseObject):
...
+        TradeMarkRegisteringInstitution,
+        verbose_name=_('Registering Institution'),
```
szymi-: ditto

szymi- (changes requested):
Also, there is still some reference to the old "region" field in views, which is causing tests to fail.

src/ralph/trade_marks/models.py
smaciejs: Convinced me, I'll change it.

src/ralph/trade_marks/models.py
smaciejs: Deleted

src/ralph/trade_marks/admin.py
```diff
@@ -62,10 +68,26 @@ class TradeMarkAdmin(AttachmentsMixin, RalphAdmin):
             )
         })
     )
-    search_fields = ['name', 'id', ]
-    readonly_fields = ['image_tag']
+
+    class TradeMarksAdditionalCountyInline(RalphTabularInline):
```
Blejwi: I guess there is a typo in `County`

Blejwi (changes requested):

src/ralph/trade_marks/admin.py
smaciejs: Fixed

Blejwi (approved):

szymi- (approved):

PR #3424, opened 2019-03-02
Update installation.md

Russell-IO:
The Key used to sign packages in https://dl.bintray.com/allegro/debng has been updated,
This PR updates the value used in the docs


romcheg (approved):
Good catch. Thanks!

Blejwi (approved):

PR #3426, opened 2019-03-04
make network vlan searchable

ahaupt:
It would be much more convenient for us to directly search a network by its VLAN tag. This simple patch extends the list of searchable items by the vlan key.

### DIFF
--- a/src/ralph/networks/admin.py
+++ b/src/ralph/networks/admin.py
@@ -111,7 +111,7 @@ class NetworkRalphChangeList(RalphChangeList):
 class NetworkAdmin(RalphMPTTAdmin):
     ordering = ['min_ip', '-max_ip']
     change_form_template = 'admin/data_center/network/change_form.html'
-    search_fields = ['name', 'address', 'remarks']
+    search_fields = ['name', 'address', 'remarks', 'vlan']
     list_display = [
         'name', 'address', 'kind', 'vlan', 'network_environment',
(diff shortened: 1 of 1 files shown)


src/ralph/networks/admin.py
```diff
@@ -111,7 +111,7 @@ def get_queryset(self, request):
 class NetworkAdmin(RalphMPTTAdmin):
     ordering = ['min_ip', '-max_ip']
     change_form_template = 'admin/data_center/network/change_form.html'
-    search_fields = ['name', 'address', 'remarks']
+    search_fields = ['name', 'address', 'remarks', 'vlan']
```
romcheg, line 114: Could you please also make it possible to filter by VLAN range by extending ```list_filter``` property with ```'vlan'```? You may put it into any position that you think will be the best.

romcheg (changes requested):
Thanks for your contribution! I have mentioned one small change inline, otherwise LGTM!

src/ralph/networks/admin.py
ahaupt: OK, should be in 37415d16bc04fbe41bcf4fe754131b601eb234b0

romcheg: Thanks!

romcheg (approved):
I verified the changes on my local environment and can confirm everything works as expected. Thanks!

szymi- (approved):

PR #3427, opened 2019-03-05
Document development environment

romcheg:


Blejwi (approved):

docs/development/development_environment.md
```diff
@@ -0,0 +1,134 @@
...
+The environment will be activated upon creation.  
+1. Install required python packages  
+        ```
+        $ pip install -r requirements/dev.txt
+        ```  
+Please note, it may be necessary to install appropriate shared libraries and
+build tools in order for some python packages to be built.
+1. Install JS requirements and build static files  
+        ```
+        $ npm install
```
vi4m, line 54: This list of commands is not maintainable in this documentation. We used to have simple wrappers on series of commands, inside Makefile. For example: something like `make install-js` for `npm` related stuff. What happended to that?

vi4m (changes requested):

docs/development/development_environment.md
MarekBleschke: We've discussed this on hackhaton and decided that we want to remove non-used stuff from Makefile. They were not maintained as nobody used them in day to day work. As a result sooner or later commands were obsolete and not working.

vi4m: Who exactly is "nobody"? I think it's worth keeping couple(3?) of targets in Makefile, to avoid repeating the same commands in:

- documentation
- travis.yml

It doesn't required much effort, and give benefits to everyone.

zrzut ekranu 2019-03-6 o 11 18 22

romcheg: In our experience in every project where there is a target like this it got broken at some point. Right here this one you posted won't work for everyone and is missing one command. On the other hand we use these individual commands to perform specific tasks, so we sort of maintain them on regular basis.

The point is that npm install, gulp, and collectstatic are familiar to every django developer, but our own targets having meaningless names like install-js are not. It also works similar with packaging scripts -- most of contributors who were trying to do something with the packaging were trying to run debuild before they found out there is a series of make targets we maintain.

MarekBleschke: install-js would be one time only command which would be used only when setting up environment for the first time. It isn't used in build process and it isn't used in day to day work. 

We've made a decision that we want to remove this types of commands. From our experience they are not maintainable (this conclusions is from last time someone was installing Ralph and SC locally). As this didn't work for us, we want to try another approach where we don't create Makefile commands that are not often used.

Blejwi: `install-js` is actually wrong name IMO. It actually installs node modules and uses gulp to compile it (which is not installation anymore).

If we want to be strict with naming, we should separate install-js into 2 commands: `install-js-dependencies`, `compile-statics` (or `compile-js`).

Where each command wraps only one command, so I don't see a point to keep a wrapper for just one command.
IMO it's ok to give a list of commands to run for contributors, instead of wrappers, at least we are sure that contributors know what tools we are using :)

vi4m (approved):

docs/development/development_environment.md
vi4m: Ok, given the fact "npm install and gulp" is de facto standard nowadays, and we use it without additional commandline switches I'm in favor of this proposition. 

If things get more complicated (additional commandline switches to npm, replacing npm with yarn back and forth) we should make it back to Makefile. For now - let's remove it.

Thanks for discussing it in details.

PR #3443, opened 2019-04-09
RAS-1448 | Eliminate MX and TXT records from checking_type function

smaciejs:
Eliminate MX and TXT records from checking_type function

### DIFF
--- a/src/ralph/ssl_certificates/management/commands/update_dns_service_env_from_dnsaas.py
+++ b/src/ralph/ssl_certificates/management/commands/update_dns_service_env_from_dnsaas.py
@@ -8,9 +8,8 @@ from ralph.ssl_certificates.models import SSLCertificate
 
 
 def checking_type(value):
-    """
-    if record type is conected with mail records, then return empty string
-    """
+    # if record type is conected with mail records, then return empty string.
+    # When we get empty string, management command did not change service env.
     if value['type'] in {'MX', 'TXT'}:
         domain = ''
     elif value['type'] in {'CNAME', 'PTR'}:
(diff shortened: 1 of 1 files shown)


src/ralph/ssl_certificates/management/commands/update_dns_service_env_from_dnsaas.py
```diff
@@ -8,7 +8,10 @@
 
 
 def checking_type(value):
-    if value['type'] in {'CNAME', 'PTR'}:
+    domain = ''
+    if value['type'] in {'MX', 'TXT'}:
+        pass
```
mkurek:
```
if value['type'] in {'MX', 'TXT'}:
    domain = ''
```

mkurek (changes requested):

src/ralph/ssl_certificates/management/commands/update_dns_service_env_from_dnsaas.py
smaciejs: Fair point

mkurek (dismissed):

src/ralph/ssl_certificates/management/commands/update_dns_service_env_from_dnsaas.py
vi4m: I think we definitely need comment here - why is this change now required? @smaciejs

smaciejs: Bacause all our MX and TX records have the same service (email) and this this two records gives us false data

MarekBleschke: IMO it would be good to add this comment in code for future generations ;)

MarekBleschke (approved):

Blejwi (approved):

PR #3467, opened 2019-06-25
Cloud images - support duplicated names

tomez:


src/ralph/virtual/admin.py
```diff
@@ -514,7 +514,7 @@ class CloudProviderAdmin(RalphAdmin):
 @register(CloudImage)
 class CloudImageAdmin(RalphAdmin):
     list_display = ['name', 'image_id']
-    list_filter = ['name', 'image_id']
+    list_filter = ['image_id']
```
romcheg: After giving it a thought I think it's actually better to leave the name here and get rid of the uuid -- I can imagine someone recalling a part of the image name and trying to filter by that, however, remembering an image uuid is unlikely.

src/ralph/virtual/migrations/0012_cloudimage.py
```diff
@@ -16,13 +16,9 @@ class Migration(migrations.Migration):
             name='CloudImage',
```
romcheg: You cannot change already merged migrations -- they are already applied somewhere. Just generate a new one.

src/ralph/virtual/admin.py
romcheg: Another point is that it would be useful to enable API filtering by both name and uuid fields. API filters are defined on view sets and by default are copied from admin. So in this case it's necessary to add a ```list_filter``` property to ```CloudImageViewSet```. Here you can find an example: https://github.com/allegro/ralph/blob/ng/src/ralph/virtual/api.py#L181

romcheg (changes requested):
Left several comments inline.

src/ralph/virtual/admin.py
tomez: Done.

src/ralph/virtual/migrations/0012_cloudimage.py
tomez: Done.

romcheg (approved):
LGTM, thanks!

szymi- (approved):

PR #3495, opened 2019-12-11
Restricted custom fields

romcheg:
Add an optional managing group property to custom fields. If set,
adding, changing or deleting values of those custom fields will
be restricted to mebmerb of a specified groups only.

Custom field form will not show such custom field values to
unauthorized users.

### DIFF
--- a/src/ralph/lib/custom_fields/admin.py
+++ b/src/ralph/lib/custom_fields/admin.py
@@ -24,7 +24,7 @@ class CustomFieldAdmin(RalphAdmin):
     list_filter = ['type']
     fields = [
         'name', 'attribute_name', 'type', 'choices', 'default_value',
-        'use_as_configuration_variable'
+        'managing_group', 'use_as_configuration_variable'
     ]
     readonly_fields = ['attribute_name']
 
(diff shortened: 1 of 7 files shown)


src/ralph/lib/custom_fields/admin.py
```diff
@@ -135,3 +135,33 @@ def changeform_view(
         return super().changeform_view(
             request, object_id, form_url, extra_context
         )
+
+    def _create_formsets(self, request, obj, change):
+        """Helper function to generate formsets for add/change_view."""
```
MarekBleschke: Please add comment that this is 99% not altered internal django method + comment around lines that are altered and why.

MarekBleschke (changes requested):

src/ralph/lib/custom_fields/admin.py
romcheg: Done

Blejwi (dismissed):

src/ralph/lib/custom_fields/api/viewsets.py
```diff
@@ -42,6 +44,12 @@ def _get_related_model_info(self):
         }
         return info
 
+    def _user_can_manage_customfield(self, user, custom_field):
+        return (
+            custom_field.managing_group is None or
+            custom_field.managing_group in user.groups.all()
```
ar4s: Did you try ``user.groups.filter(pk=custom_field.managing_group.pk).exists()``?

src/ralph/lib/custom_fields/forms.py
```diff
@@ -39,6 +40,21 @@ def save(self, *args, **kwargs):
 
 
 class CustomFieldValueFormSet(BaseGenericInlineFormSet):
+
+    def __init__(
+        self, data=None, files=None, instance=None, save_as_new=None,
+        prefix=None, queryset=None, request=None, **kwargs
+    ):
```
ar4s: You can write ``def __init__(self, request=None, *args, **kwargs):`` if you only need ``request``. Also ``super`` clause should simplify to ``super().__init__(request=request, queryset=queryset, *args, **kwargs)``

src/ralph/lib/custom_fields/forms.py
```diff
@@ -39,6 +40,21 @@ def save(self, *args, **kwargs):
...
+        queryset = queryset.filter(
+            Q(custom_field__managing_group__isnull=True) |
+            Q(custom_field__managing_group__in=request.user.groups.all())
+        )
+
+        super().__init__(
+            data=data, files=files, instance=instance, save_as_new=save_as_new,
```
ar4s: Where is ``request``?

src/ralph/lib/custom_fields/admin.py
```diff
@@ -135,3 +135,42 @@ def changeform_view(
         return super().changeform_view(
             request, object_id, form_url, extra_context
         )
+
+    def _create_formsets(self, request, obj, change):
```
ar4s, line 139: Where did you use this function?

MarekBleschke (dismissed):

ar4s (changes requested):

src/ralph/lib/custom_fields/admin.py
romcheg: It it used internally by the ModelAdmin class

src/ralph/lib/custom_fields/api/viewsets.py
romcheg: Looks more efficient. Thanks!

src/ralph/lib/custom_fields/forms.py
romcheg: +1

src/ralph/lib/custom_fields/forms.py
romcheg: It is not expected there.

ar4s (dismissed):

Blejwi (dismissed):

Blejwi (approved):

romcheg:
The commits have been squashed. Since no other changes have been made, I'm going to merge this with a single approve.

PR #3481, opened 2019-08-27
Changes about electronic give out for SIMCards.

lukaszkarykowski:
Added modules from BackOffice transitions to SIMCards module.
This modules allows to accept give out of SIMCards electronically.

PR #3500, opened 2019-12-17
Put back "Changes about electronic give out for SIMCards. (#3481)""

romcheg:


romcheg:
The original author has created his own PR

PR #3527, opened 2020-04-27
Restrict deletion of used cloud flavors and providers

romcheg:


src/ralph/virtual/api.py
```diff
@@ -180,11 +183,48 @@ class CloudFlavorViewSet(RalphAPIViewSet):
...
     filter_fields = ['flavor_id']
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        force_delete = request.data.get('force', False)
+
+        if instance.cloudhost_set.count() != 0 and not force_delete:
+            raise Conflict(
+                _(
+                    "Cloud flavor is in use and hence is not delible. "
```
MarekBleschke: Shouldn't it be `is not deletable`?

MarekBleschke (changes requested):

src/ralph/virtual/api.py
romcheg: Frankly, I am not sure either. I have checked several dictionaries and found out some of them recommend using delible while the others use deletable. @szymi- your English is better -- could you please help us here?

szymi-: I'd stick to deletable. Delible sounds a little archaic to me.

MarekBleschke (approved):

PR #3531, opened 2020-05-11
Added new transition, and new status for Access Cards.

lukaszkarykowski:


src/ralph/access_cards/models.py
```diff
@@ -25,7 +25,8 @@ class AccessCardStatus(Choices):
     damaged = _('damaged')
     used = _('in use')
     free = _('free')
-    return_in_progress = _('return in progres')
+    reserved = _("reserved")
+    return_in_progress = _('return in progress')
```
romcheg, line 28: Always add new statuses to the end of the list.

romcheg (changes requested):

src/ralph/access_cards/models.py
lukaszkarykowski: ok. Done.

romcheg (commented):
Looks like you have to generate migrations to make CI happy again. LGTM otherwise.

szymi- (approved):

romcheg (approved):
LGTM, thanks for following up!

PR #3549, opened 2020-07-22
Add asset model to dchost serializer

awieckowski:


src/ralph/assets/api/serializers.py
```diff
@@ -456,3 +456,24 @@ class Meta:
             'created', 'modified', 'remarks', 'parent', 'object_type',
             'configuration_variables', 'securityscan',
         ]
+
+
+class DCHostPhysicalSerializer(ComponentSerializerMixin, BaseObjectSerializer):
```
szymi-: `DCHostPhysicalSerializer` could extend `DCHostSerializer` since they are very similar.

szymi- (changes requested):

szymi- (dismissed):

szymi- (approved):

PR #3559, opened 2020-09-08
money field for licence price

szymi-:


romcheg (changes requested):
Could you please add a migration?

UPD: Sorry — did not notice it somehow.

szymi-:
I added the migration while you were reviewing. Then I changed it again. Now it should be ok.

src/ralph/licences/migrations/0006_auto_20200908_0847.py
```diff
@@ -0,0 +1,26 @@
...
+        ),
+        migrations.AlterField(
```
romcheg: Will this convert existing values with no data loss?

src/ralph/licences/models.py
```diff
@@ -191,8 +192,8 @@ class Licence(Regionalizable, AdminAbsoluteUrlMixin, BaseObject):
         help_text="Leave blank if this licence is perpetual",
     )
     order_no = models.CharField(max_length=50, null=True, blank=True)
-    price = models.DecimalField(
-        max_digits=10, decimal_places=2, default=0, null=True, blank=True,
+    price = MoneyField(
```
romcheg: Shouldn't the `price_currency` field be also reflected on the model?

src/ralph/licences/migrations/0006_auto_20200908_0847.py
szymi-: Yes. `MoneyField` extends `DecimalField`. Precision and default value stay the same; field name (`price`) also remains the same. In the database, currency is stored in a separate field (`price_currency`). Also, in the REST API field `price` looks the same as before; currency is displayed in a separate `price_currency` fideld.

romcheg:
@szymi- I've commented in a few places I'm concerned about since I'm not aware how this django-money works. If those concerns are invalid and the CI is happy I'll be happy to approve.

src/ralph/licences/models.py
szymi-: The field `price_currency` is automatically created by the library. `MoneyField` combines the two fields (`price` and `price_currency`) into one model.

szymi- (commented):
> @szymi- I've commented in a few places I'm concerned about since I'm not aware how this django-money works. If those concerns are invalid and the CI is happy I'll be happy to approve.

Thanks. I considered creating a separate field for currency while keeping the original `price` field, however this library does exactly the same in the background with some additional bonuses (most importantly, admin form and REST API presentation are dealt with by the library). I verified the library with exactly the issues you rose in mind - and came to the conclusion that it does the job well.

romcheg (dismissed):
Thanks for resolving my concerns. LGTM

romcheg:
@szymi- I've just got one more concern -- does this change the format of an existing object field that Ralph returns via the API?

szymi-:
> @szymi- I've just got one more concern -- does this change the format of an existing object field that Ralph returns via the API?

Field `price` remains the same. New field `price_currency` is added.

szymi-:
@romcheg @MarekBleschke I am considering unifying `price` field in all models across the entire project, possibly via a `PriceMixin`. I have two reasons for that: to unify the way reports are generated (at the moment reports mixing `Licence` price with prices of other models will not work). The second reason is to unify the approach for price in Ralph. Please let me know what you think about that.

MarekBleschke:
> @romcheg @MarekBleschke I am considering unifying `price` field in all models across the entire project, possibly via a `PriceMixin`. I have two reasons for that: to unify the way reports are generated (at the moment reports mixing `Licence` price with prices of other models will not work). The second reason is to unify the approach for price in Ralph. Please let me know what you think about that.

Yes, unifying price field in all models IMO is the way to go.

MarekBleschke (dismissed):

src/ralph/licences/tests/factories.py
```diff
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
+import random
 from datetime import date, datetime, timedelta
 
 import factory
+from djmoney.money import Money
 from factory.django import DjangoModelFactory
-from factory.fuzzy import FuzzyText
+from factory.fuzzy import FuzzyDecimal, FuzzyText
```
MarekBleschke: `FuzzyDecimal` looks unused

MarekBleschke (approved):

PR #3562, opened 2020-09-13
Run tests with GitHub Workflows

romcheg:
The new workflow runs the following tests in parallel:
- Code style
- Missing migrations for MySQL and PostgreSQL
- Settings tests
- Python tests for Ralph and ralph_dhcp_agent on both MySQL
  and PostgreSQL


romcheg:
As you can see there are plenty of repetitions in the workflow file. Unfortunately GitHub workflows does not support YAML anchors at this moment. As soon as they are supported, it will be possible to reduce the size of the workflow file significantly.

romcheg:
Another thing: GitHub runs workflows for pull requests if there's at least one workflow in the base branch. For that reason it won't run anything for this PR. I have tested these changes in my own repo: https://github.com/romcheg/ralph/actions/runs/252639137

MarekBleschke (dismissed):

.github/workflows/main.yml
```diff
@@ -0,0 +1,252 @@
...
+on:
+  push:
+    branches: [ ng ]
+  pull_request:
+    branches: [ ng ]
+
+jobs:
+  code_style:
+    name: Code style
+    runs-on: ubuntu-18.04
```
szymi-: Why not Ubuntu 20.04?

.github/workflows/main.yml
```diff
@@ -0,0 +1,252 @@
...
+            ${{ runner.os }}-
+      - name: Install Ralph's python dependencies
+        run: pip3 install -r "${GITHUB_WORKSPACE}/requirements/code_style.txt"
+
+      - name: Run flake8
+        run: make flake
+
+  missing_migrations:
+    name: Missing migrations
+    runs-on: ubuntu-18.04
```
szymi-: Ditto

.github/workflows/main.yml
```diff
@@ -0,0 +1,252 @@
...
+        if: ${{ matrix.db-engine == 'psql' }}
+        run: pip3 install psycopg2-binary
+
+      # Runs a set of commands using the runners shell
+      - name: Check missings migrations
+        run: "${GITHUB_WORKSPACE}/scripts/check_missing_migrations.sh"
+
+  settings_test:
+    name: Settings tests
+    runs-on: ubuntu-18.04
```
szymi-: Ditto

.github/workflows/main.yml
```diff
@@ -0,0 +1,252 @@
...
+        run: |
+          pip3 install -r "${GITHUB_WORKSPACE}/requirements/test.txt"
+          pip3 install -r "${GITHUB_WORKSPACE}/requirements/openstack.txt"
+      
+      - name: Test Ralph's settings
+        run:  ./scripts/test_settings.sh prod test >/dev/null
+
+  python_tests:
+    name: Python tests
+    runs-on: ubuntu-18.04
```
szymi-: Ditto

szymi- (changes requested):

szymi-:
I am ready to approve - I just thought Ubuntu 20.04 will save us one extra change in the future.

szymi-:
Just to clarify my previous comment: I know we use Ubuntu 18.04 as Ralph's container. But the images specified in this change are something different - they are the environment in which the tests run. So when in the future Github drops support for Ubuntu 18 (regardless of the fact that we might still be using it then), the jobs will start to fail for a reason unrelated to the changes in code, which is always annoying.

romcheg:
@szymi- I chose 18.04 simply because it's the version marked as "latest" on github runners. However, 20.04 is also available and you've made a good point about dropping the support. I'll try to drop-in replace 18.04 by 20.04 and will make changes, if required.

romcheg:
@szymi- I have updated ubuntu on runners to 20.04. You can see the test results here: https://github.com/romcheg/ralph/pull/3/checks?check_run_id=1199731071

szymi- (approved):

romcheg:
@MarekBleschke ;)

MarekBleschke (approved):

MarekBleschke (approved):

PR #3578, opened 2020-10-21
Added Filter Fields by objects ID to Transitions History Serializer

lukaszkarykowski:


lukaszkarykowski:
cannot resolv conflict.

PR #3586, opened 2020-11-20
disable support for deprecated DNSaaS

MarekBleschke:


src/ralph/settings/base.py
```diff
@@ -511,8 +511,38 @@ def get_sentinels(sentinels_string):
 
 # disable integration with DNSaaS as it's no longer supported
 # https://github.com/allegro/django-powerdns-dnssec
+# This settings will be deleted in future release.
 ENABLE_DNSAAS_INTEGRATION = False
+DNSAAS_URL = os.environ.get('DNSAAS_URL', '')
```
romcheg, line 516: I'm not sure I'm getting the idea behind this PR right. If the goal is to disable DNSaaS support then why are these settings needed here?

MarekBleschke: It will be two step process:
1. Disabling DNSaaS support as DNSaaS service is not maintained for some time and its repo is archived.
2. Removing all DNSaaS code from Ralph repository.

This will be done in two steps as we need to use DNSaaS client internally (in short-term with some minor modifications in separate PR) but I want to drop support for it sooner.

For reason that we will keep DNSaaS client code a little bit longer, we also need it's settigns so that tests will be working properly.

romcheg: Makes sense. Thanks for the explanation!

romcheg (dismissed):
LGTM

szymi- (approved):

romcheg (approved):

PR #3606, opened 2021-02-25
Add google tag manager

awieckowski:


src/ralph/settings/prod.py
```diff
@@ -86,3 +86,6 @@
         STATSD_GRAPHS_PREFIX = os.environ.get(
             'STATSD_GRAPHS_PREFIX', 'ralph.graphs'
         )
+
+if os.environ.get('TAG_MANAGER_TAG_ID'):
```
MarekBleschke: This setting is not needed in `prod.py` as `TAG_MANAGER_TAG_ID` is set in `base.py`.

src/ralph/settings/base.py
```diff
@@ -663,3 +663,5 @@ def get_sentinels(sentinels_string):
 
 CONVERT_TO_DATACENTER_ASSET_DEFAULT_STATUS_ID = 1
 CONVERT_TO_BACKOFFICE_ASSET_DEFAULT_STATUS_ID = 1
+
+TAG_MANAGER_TAG_ID = os.environ.get('TAG_MANAGER_TAG_ID', None)
```
MarekBleschke: I think it would be more readable to change name this setting to be more specific: `GOOGLE_TAG_MANAGER_TAG_ID` or `GOOGLE_ANALYTICS_ID`.

MarekBleschke (changes requested):

MarekBleschke (approved):

PR #3608, opened 2021-03-02
Feature/update vip event support

mzdeb:
`updateVipEvent` was never supported properly, this  PR adds basic support of such events.


szymi- (approved):

src/ralph/data_center/subscribers.py
```diff
@@ -141,18 +133,54 @@ def handle_update_vip_event(data):
...
+    # update cluster.
+    cluster_type, _ = ClusterType.objects.get_or_create(
+        name=data['load_balancer_type']
+    )
+    cluster, _ = Cluster.objects.get_or_create(
+        name=data['load_balancer'],
+        type=cluster_type,
+    )
+
+    if ip.ethernet.base_object != cluster:
```
szymi-: What if ip does not have `ethernet` or `ethernet` does not have a `base_object`? Some validation for that would be useful, ethernet field is not required for an IP address.

src/ralph/data_center/subscribers.py
```diff
@@ -141,18 +133,54 @@ def handle_update_vip_event(data):
...
+    ip, _ = IPAddress.objects.get_or_create(address=data['ip'])
     protocol = VIPProtocol.from_name(data['protocol'].upper())
-    vip = get_vip(ip.address, data['port'], protocol.name)
+    vip = get_vip(ip, data['port'], protocol)
     if vip is None:
+        # VIP not found, should create new one.
+        return handle_create_vip_event(data)
+
+    # update cluster.
+    cluster_type, _ = ClusterType.objects.get_or_create(
```
szymi-, line 144: Not sure if I like the idea of creating cluster types this way. Would it not be better to log the fact that a given cluster type does not exist and then create it if required?

src/ralph/data_center/subscribers.py
```diff
@@ -141,18 +133,54 @@ def handle_update_vip_event(data):
...
+        name=data['load_balancer_type']
+    )
+    cluster, _ = Cluster.objects.get_or_create(
```
szymi-, line 147: Same here.

szymi- (changes requested):

src/ralph/data_center/subscribers.py
szymi-: I think we're missing two very important checks here:

1. check here if the original `ip.ethernet.base_object` is a cluster - we don't want to take an IP assigned to a non-cluster host and assign it to a cluster
2. check here if `ip.ethernet` is exposed in DHCP. If it is, refuse to assign it to a VIP.

szymi-: And one last thing: VIP is missing assignment of parent - see line 116 in `handle_create_vip_event`.

src/ralph/data_center/subscribers.py
```diff
@@ -119,19 +127,39 @@ def handle_create_vip_event(data):
     logger.debug('VIP %s created successfully.', vip.name)
 
 
+def migrate_vip_to_cluster(vip, cluster, protocol):
+    msg = (
```
szymi-, line 132:
```suggestion
    cluster_content_type = ContentType.objects.get_for_model(Cluster)
    msg = (
```

src/ralph/data_center/subscribers.py
```diff
@@ -119,19 +127,39 @@ def handle_create_vip_event(data):
...
+        'failed: %s'.format(vip.ip.address, vip.port, protocol.name)
+    )
+    ethernet = vip.ip.ethernet
+    if not ethernet:
+        logger.error(msg, 'no `Ethernet` object found')
+        return
+    if ethernet.base_object != vip.parent:
+        logger.error(msg, '`Ethernet` base_object differs from `VIP` parent')
+        return
+    if not isinstance(ethernet.base_object, Cluster):
```
szymi-: This does not work (tested):

```
(Pdb++) type(ethernet.base_object)
<class 'ralph.assets.models.base.BaseObject'>
(Pdb++) type(vip.parent)
<class 'ralph.assets.models.base.BaseObject'>

```

```suggestion
    if ethernet.base_object.content_type != cluster_content_type:
```

src/ralph/data_center/subscribers.py
```diff
@@ -119,19 +127,39 @@ def handle_create_vip_event(data):
...
+        logger.error(msg, '`Ethernet` base_object is not `Cluster` instance')
+        return
+    if not isinstance(vip.parent, Cluster):
```
szymi-:
```suggestion
    if vip.parent.content_type != cluster_content_type:
```

src/ralph/data_center/subscribers.py
```diff
@@ -3,6 +3,7 @@
 
 import pyhermes
 from django.core.exceptions import ValidationError
```
szymi-, line 6:
```suggestion
from django.contrib.contenttypes.models import ContentType
from django.core.exceptions import ValidationError
```

szymi- (changes requested):

szymi- (approved):

PR #3612, opened 2021-03-12
Added a condition for the buyout ticket link. 

lukaszkarykowski:
if asset have no buyout date - you cant generate ticket

### DIFF
--- a/src/ralph/accounts/admin.py
+++ b/src/ralph/accounts/admin.py
@@ -117,21 +117,20 @@ class AssetList(Table):
             return []
 
     def buyout_ticket(self, item):
-        get_params = {
-            "inventory_number": item.barcode,
-            "serial_number": item.sn,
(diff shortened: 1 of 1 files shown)


src/ralph/accounts/admin.py
```diff
@@ -127,7 +127,10 @@ def buyout_ticket(self, item):
             [settings.MY_EQUIPMENT_BUYOUT_URL, urlencode(get_params)]
         )
         url_title = 'Report buyout'
-        return self.create_report_link(url, url_title, item)
+        if not item.model.category.show_buyout_date:
```
MarekBleschke, line 120: We can return fast here, as we have access to `item` object at the beginning of the function. I propose to move this `if` statement as the first line in this function.

lukaszkarykowski: Done.

MarekBleschke (approved):

PR #3625, opened 2021-04-26
convert string from environment to int in settings

szymi-:


vi4m (commented):
Does it fix some bug in code, where NoneType was incorrectly handled? If so, describe the proper reason in the commit message.

hipek8 (approved):

PR #3652, opened 2021-08-11
feat: added helm chart

wid:
Hello,

Please accept my contribution of a helm chart to deploy ralph on a kubernetes cluster.

Once you install the prerequisites (see README.md), it will start ralph along with nginx, mysql and redis with the helm install.

Thank you for this tool !

### DIFF
--- a/charts/ralph/README.md
+++ b/charts/ralph/README.md
@@ -1,0 +1,37 @@
+
+# Prerequistes
+
+ - Kubernetes cluster >= v1.19.11
+ - Helm 3 
+Find instructions here: https://helm.sh/docs/intro/install/)
+ - KubeDB Community Edition (optional but easier)
+Find instructions here: 
+https://kubedb.com/docs/v2021.06.23/setup/install/community/
+
+# How to install this Helm chart
+
+## Edit values in values.yaml
+
+Default should be fine except the ingress hostname
+
+## Install the chart & Configure database
+
+```
+helm install ralph charts/ralph # from ralph git root directory
+kubectl get pods
+NAME                     READY   STATUS    RESTARTS   AGE
+mysql-0                  1/1     Running   0          104s
+ralph-58f4968585-f7cm7   2/2     Running   0          105s  <------ralph pod
+redis-0                  1/1     Running   0          104s
+```
+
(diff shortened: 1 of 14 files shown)


charts/ralph/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  port: 80
+
+ingress:
+  enabled: true
+  className: ""
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: ralph-demo.allegro.tech
```
vi4m, line 50: Not sure about that. Why this particular domain?

vi4m (changes requested):

charts/ralph/values.yaml
wid: Well it's if you want to use it yourself.
You can put anything else there.

vi4m:
I'm thinking about moving this configuration to the "contrib" directory (which already has infrastructure things like `docker-compose`, `dhcp` etc). What do you think?

romcheg:
@wid thanks for the great chart!

I agree with @vi4m that this should perhaps be in the `contrib` catalogue.

At the same time I kindly ask to not merge it right now. I agreed with @szymi- to do a little of a follow-up changes on the top in order to make it a production-grade chart including automatic testing

wid:
Yes, you can do that.

charts/ralph/README.md
```diff
@@ -0,0 +1,37 @@
+
+# Prerequistes
+
+ - Kubernetes cluster >= v1.19.11
+ - Helm 3 
+Find instructions here: https://helm.sh/docs/intro/install/)
+ - KubeDB Community Edition (optional but easier)
+Find instructions here: 
+https://kubedb.com/docs/v2021.06.23/setup/install/community/
```
stxm, line 9: Using kubedb seems like a uncommon choice to me. Suggestion to use mysql and redis as subchart.

wid:
Yes, mysql seems a better choice, specially that latest version do not have
basic needed features in the latest versions.

PR #3664, opened 2021-10-11
Copied missing images to `docs/user/img` to fix missing images in Documenmtation

RuttJen:
Hello,

the images were still not displayed in the documentation despite the correct path, this was due to missing files in the folder `docs/user/img`. This pull request fixes this issue by copying the existing images from `docs/img` to `docs/user/img`.

Another possibility for a fix would probably have been to use the `docs/img` folder as reference for the images.

Greetings


vi4m:
Yes, that was the case, exactly! Can you move it instead of copying ,using `git mv`? It should relink nicely in the git history...

RuttJen:
Sorry for the delay, I had a some issues with my time management ;-)

I moved the images from `/docs/img` to `/docs/user/img` and kept the images used in `index.md`.

Additionally I have

 * created a folder `img` under `/docs/development` and
 * moved the required images in this chapter to this folder (`add_transitions.png` and `extra_params.png`)
 * I also adjusted the files under `/docs/develoment` so that the paths for the images to be displayed now point to the files in `/docs/development/img`.

PR #3666, opened 2021-10-15
nginx service waiting for web:8000 at startup

pinoatrome:
the docker  compose command docker-compose up -d can result in a nginx container exited after error connecting to web:8000 (proxy_pass http://web:8000; in nginx conf file) when web container is still not available.
using ralf.entrypoint.nginx script as entrypoint.sh for the nginx image will start nginx process only after service web is active and responding on port 8000

### DIFF
--- a/contrib/docker/ralph.entrypoint.nginx
+++ b/contrib/docker/ralph.entrypoint.nginx
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+export RALPH_URL="web:8000"
+
+function ralph_ready(){
+  curl -s "${RALPH_URL}"
+}
+
+>&2 echo "Checking Ralph (at ${RALPH_URL})..."
+until ralph_ready; do
+  >&2 echo "Ralph is unavailable - waiting 1 sec for ${RALPH_URL}"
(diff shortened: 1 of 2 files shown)


pinoatrome:
the error is
`nginx: [emerg] host not found in upstream "web" in /etc/nginx/conf.d/default.conf:22`

and this is the output of "docker-compose ps" command (note nginx service down) after "docker-compose up -d" command to start ralph.

```
$:~/projects/ralph$ docker-compose ps
      Name                    Command               State            Ports
---------------------------------------------------------------------------------------------
ralph_db_1      docker-entrypoint.sh mysql ...   Up      0.0.0.0:3306->3306/tcp, 33060/tcp
ralph_inkpy_1   /bin/sh -c supervisord -n        Up
ralph_nginx_1   /docker-entrypoint.sh ngin ...   Exit 1
ralph_redis_1   docker-entrypoint.sh redis ...   Up       0.0.0.0:9079->6379/tcp
ralph_web_1     /var/local/ralph/docker-en ...   Up       0.0.0.0:9000->8000/tcp
```

vi4m (approved):

PR #3680, opened 2021-11-21
initial data commands

szymi-:


hipek8 (dismissed):

src/ralph/data_importer/management/commands/create_network.py
```diff
@@ -0,0 +1,143 @@
...
+    Generate a single, production ready network
+    """
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-d', '--dc-name',
+            default='dc1',
+            dest='dc_name',
+            help='Data center name.'
+        )
+        parser.add_argument(
```
MarekBleschke, line 53: I think allowing only for two DNS servers (no more, no less) is too restrictive. Maybe we could allow for `--dns` to be list of arguments?

src/ralph/data_importer/management/commands/create_network.py
```diff
@@ -0,0 +1,143 @@
...
+            dns1_address=dns1_address,
+            dns2_address=dns2_address,
+            gateway_address=gateway_address,
+            dc_name=dc_name,
+            server_room_name=server_room_name,
+            create_rack=create_rack,
+        )
+
+
+def get_or_create(model, **kwargs):
```
MarekBleschke: Why not just use `get_or_create` on model? I don't see much benefit in doing `obj = get_or_create(model, xyz)` vs `obj, _ = model.objects.get_or_create(xyc)` but it clouds standard Django method with our custom function (we need to check what it does and how to invoke it).

src/ralph/data_importer/management/commands/create_preboot_configuration.py
```diff
@@ -0,0 +1,78 @@
+from django.core.management import BaseCommand
+from django.db import transaction
+
+from ralph.data_importer.management.commands.create_network import get_or_create
```
MarekBleschke: IMO importing functions between commands it's susceptible to breaking stuff and not knowing about it :)
I wouldn't expect that `get_or_create` from `create_network.py` is used anywhere else. If it is it should be in separate file, not in command file.

src/ralph/data_importer/management/commands/create_transitions.py
```diff
@@ -0,0 +1,87 @@
...
+from ralph.data_center.models import DataCenterAsset, DataCenterAssetStatus
+from ralph.lib.transitions.conf import DEFAULT_ASYNC_TRANSITION_SERVICE_NAME
+from ralph.lib.transitions.models import Action, Transition, TransitionModel
+
+
+class Command(BaseCommand):
+    """
+    Generate production ready Data Center Asset Transitions
+    """
+    def handle(self, *args, **options):
```
MarekBleschke, line 13: I think this is more readable order of methods in class: first `handle` than methods invoked in `handle` in order of their invocation. Can we apply this approach to the rest of commands in this PR? :)

src/ralph/data_importer/management/commands/initial_data.py
```diff
@@ -0,0 +1,165 @@
...
+            network_address = parent_network.network_address
+            dns_1 = ipaddress.ip_address(options.get('dns_1'))
+            dns_2 = ipaddress.ip_address(options.get('dns_2'))
+            number_of_subnets = int(options.get('number_of_subnets'))
+            region = options.get('region')
+            configuration_path = options.get('configuration_path')
+        except ValueError as e:
+            raise CommandError(e)
+
+        self._validate_network(parent_network, number_of_subnets)
```
MarekBleschke, line 38: IMO these validations should be in corresponding command classes.

MarekBleschke (changes requested):

src/ralph/data_importer/management/commands/create_network.py
szymi-: This script is designed as a quick start enabling users to boot Ralph with some basic data. Any extra servers can be added later, in a standard way (via the GUI). Two DNS servers is a good starting point and pretty much a standard configuration configuration for any DHCP server (one primary DNS and one for redundancy as a fail over).

src/ralph/data_importer/management/commands/create_transitions.py
szymi-: Sure.

src/ralph/data_importer/management/commands/initial_data.py
szymi-: Ok.

src/ralph/data_importer/management/commands/create_network.py
szymi-: I will change it to get_or_create on model.

src/ralph/data_importer/management/commands/create_preboot_configuration.py
szymi-: Ditto.

src/ralph/data_importer/management/commands/initial_data.py
szymi-: This validation is specific for this command. It checks sanity of the combination of specified arguments. I would rather keep it here.

MarekBleschke (approved):

PR #3682, opened 2021-11-23
refactor openstack client

szymi-:


vi4m:
Can we remove direct dependency on openstack from base.txt ? Not sure why it's there (-r openstack.txt), since we still check imports during the startup. Setting it up(installing openstack dependencies) is nightmarish experience for the developers.

szymi-:
> Can we remove direct dependency on openstack from base.txt ?

Ok, but I will do it in a separate change after I've finished all modifications in the script, after I make sure everything works as expected.

hipek8 (approved):

PR #3697, opened 2021-12-20
Openstack sync changed since parameter 2

szymi-:


src/ralph/lib/openstack/client.py
```diff
@@ -193,3 +201,151 @@ def get_keystone_projects(self):
         else:
             projects_resource = self.keystone_client.tenants
         yield from projects_resource.list()
+
+
+class RalphOpenStackInfrastructureClient:
```
MarekBleschke, line 210: Having both `RalphOpenStackInfrastructureClient` and `RalphOpenstackClient` is little confusing. How `RalphOpenStackInfrastructureClient` is different from `RalphOpenstackClient` and which one should I use? We should have doscstrings for this classes describing their purpose and how to use them.

src/ralph/lib/openstack/client.py
```diff
@@ -193,3 +201,151 @@ def get_keystone_projects(self):
...
+    def _add_flavor(cls, client, flavor, openstack_flavors):
+        flavor_id = flavor['id']
+        new_flavor = {
+            'name': flavor['name'],
+            'cores': flavor['vcpus'],
+            'memory': flavor['ram'],
+            'disk': flavor['disk'] * 1024,
+            'tag': client.site['tag'],
+        }
+        openstack_flavors[flavor_id] = new_flavor
```
MarekBleschke: I think this method shouldn't modify `openstack_flavors`. It's hard to track what is happening with flavors :)
```
openstack_flavors[flavor['id']] = self._make_flavor(client, flavor)
```

MarekBleschke (changes requested):

src/ralph/lib/openstack/client.py
```diff
@@ -66,7 +66,11 @@ def get_nodes_list(self, associated=True, fields=None):
 
 
 class RalphOpenstackClient:
-
+    """
+    This is a single OpenStack instance client. It handles keystone
+    authentication. It has methods methods to fetch OpenStack data useful from
```
MarekBleschke: `It has methods methods` - minor typo, doubled `methods` :)

MarekBleschke (approved):

PR #3707, opened 2022-02-13
Small docker fixes

ConsoleCatzirl:
Add a `demodata` target to the docker entrypoint script, and update the
documenation to use the entrypoint arguments.


fajfer:
I just wanted to make exactly the same changes. Shame it wasn't reviewed for 1.5 year. @romcheg I know Ralph isn't super top priority thing in Allegro but this fixes both docs and docker setup. This fix would be very welcome

fajfer:
@hipek8 any comments on that? was it implemented one way or another?

PR #3734, opened 2022-06-28
fix vip update event handling

szymi-:


src/ralph/data_center/subscribers.py
```diff
@@ -138,10 +138,6 @@ def migrate_vip_to_cluster(vip, cluster, protocol):
     if not ethernet:
         logger.error(msg[0], *msg[1], 'no `Ethernet` object found')
         return
-    if ethernet.base_object != vip.parent:
```
szymi-, line 141: This check was unnecessary. A few lines up, we are making sure that the ethernet before change belonged to a cluster, which is sufficient to prevent accidental DHCP configuration modification (we assume that an ethernet belonging to a cluster is not managed by a DHCP server). Moreover, at the end of this function, `ethernet.base_object` and `vip.parent` are set the same value (`cluster).

awieckowski (approved):

PR #3759, opened 2023-03-02
HAL-2320 | added  new field to Backoffice assets and 2 new transition actions

lukaszkarykowski:


lukaszkarykowski:
I will create new one. Here are some errors.

PR #3785, opened 2023-09-22
ITS-454743 | added service calculator to Backoffice assets page.

lukaszkarykowski:


src/ralph/back_office/models.py
```diff
@@ -214,11 +215,32 @@ class BackOfficeAsset(Regionalizable, Asset):
...
+    @property
+    def service_calculation(self):
+        if self.service_cost is None:
+            return None
+        else:
+            today = date.today()
+            timedelta = (self.buyout_date - today)
+            months = timedelta.days // 30
+            service = (int(self.service_cost) / months)
+            buy_rate = (self.price.amount / self.depreciation_rate) + 50
```
hipek8, line 238: why/what is 50?

lukaszkarykowski: this is pattern what I receive to calculate profitability.

We can discuss this hole pattern on PM, if needed.

PR #3810, opened 2024-06-11
Optimize hosts endpoints

hipek8:
Add missing prefetches
Fix tests


src/ralph/assets/api/views.py
```diff
@@ -98,6 +97,8 @@ class Meta(NetworkableObjectFilters.Meta):
     Prefetch('licences', queryset=BaseObjectLicence.objects.select_related(
         *BaseObjectLicenceViewSet.select_related
     )),
+    'securityscan__tags',
+    'securityscan__vulnerabilities__tags',
```
hipek8, line 101: This is then used in multiple viewsets

matyldv (approved):

PR #3862, opened 2024-11-07
Fix patch dc hosts

hipek8:


src/ralph/assets/api/views.py
```diff
@@ -296,6 +301,26 @@ class DCHostViewSet(BaseObjectViewSetMixin, RalphAPIViewSet):
...
+                    return VirtualServerSaveSerializer
+                elif isinstance(obj_, DataCenterAsset):
+                    from ralph.data_center.api.serializers import DataCenterAssetSaveSerializer
+                    return DataCenterAssetSaveSerializer
+                elif isinstance(obj_, CloudHost):
+                    from ralph.virtual.api import SaveCloudHostSerializer
+                    return SaveCloudHostSerializer
+                elif isinstance(obj_, Cluster):
+                    from ralph.data_center.api.serializers import ClusterSerializer
+                    return ClusterSerializer
```
awieckowski, line 319: shouldn't we also add some "none of the above" case?

awieckowski (dismissed):

matyldv (dismissed):

matyldv (approved):

PR #3884, opened 2025-03-03
Add auto hostname

hipek8:


src/ralph/data_center/admin.py
```diff
@@ -572,6 +573,9 @@ class DataCenterAssetAdmin(
         ),
     )
 
+    def assign_mgmt_hostname(self, *args, **kwargs):
+        return assign_management_hostname_and_ip(self, *args, **kwargs)
+
```
hipek8, line 557: This is weird but for a reason - if I added original action directly to actions list the the tests wouldn't work.
So if I did
```
actions = ["bulk_edit_action", "invoice_report", assign_management_hostname_and_ip]
```
I wouldn't be able to call
```
admin.assign_management_hostname_and_ip [...]
```
Weird but hope it is fixed in some future django version

src/ralph/data_center/admin_actions.py
```diff
@@ -0,0 +1,81 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from ralph.data_center.models import DataCenterAsset
+
+
+def assign_management_hostname_and_ip(modeladmin, request, queryset):
+    for dca in queryset:
+        try:
+            if not modeladmin.has_change_permission(request, obj=dca):
+                raise RuntimeError("insufficient permissions")
```
hipek8, line 13: Because of mess with actions this action is visible for everyone so I guard here

matyldv (approved):

PR #3887, opened 2025-03-18
Information bubble

hipek8:
Limit visibility of hosts (data center asset, virtual server, cluster, cloud host) for users in information bubble
User in an information bubble or in a group which is in an information bubble will see limited assets - asset must belong to the service env of a service which belongs to at least 1 of user's information bubble.
BaseObjectsSupport visibility is also limited (but not Support visibility as it's less clear how to do that)


matyldv (dismissed):

src/ralph/lib/information_bubble/migrations/0003_auto_20250312_1327.py
```diff
@@ -0,0 +1,24 @@
...
+    dependencies = [
+        ('auth', '0011_update_proxy_permissions'),
+        ('information_bubble', '0002_servicebasedinformationbubble_groups'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='servicebasedinformationbubble',
+            name='groups',
+        ),
```
awieckowski: so migration 0002 is not really needed?

vi4m:
I know this PR is almost finished.  I'm afraid, Information Bubble is non standard and quite confusing name, isn't it ? Did we evaluate other choices ? 

A few options: 

- visibilityScope (it's precise and fits the purpose I think) 

Alternatives 

- backOfficeRegion
- accessZone

src/ralph/lib/information_bubble/migrations/0003_auto_20250312_1327.py
hipek8: Yeah, they could've been squashed into 1. Now it's 1

awieckowski (approved):

vi4m:
+1

PR #3905, opened 2025-05-20
Add filter "ne" - allow filtering out a single value

hipek8:
Also: Run tests in parallel


pyproject.toml
```diff
@@ -96,6 +96,7 @@ dependencies = [
   "six==1.16.0",
   "statsd==3.2.1",
   "stevedore",
+  "tblib>=3.1.0",
```
hipek8, line 99: This is needed for traceback to be reported when parallel test fails

awieckowski (approved):

kvehy (approved):

PR #3909, opened 2025-08-05
Add external depot id

hipek8:


src/ralph/back_office/models.py
```diff
@@ -48,6 +48,7 @@ class Warehouse(AdminAbsoluteUrlMixin, NamedMixin, TimeStampMixin, models.Model)
     _allow_in_dashboard = True
     stocktaking_enabled = models.BooleanField(default=False)
     stocktaking_tag_suffix = models.CharField(max_length=8, default="", blank=True)
+    external_depot_id = models.IntegerField(null=True, blank=True)
```
matyldv: It should be a positive integer

matyldv (changes requested):

matyldv (approved):

PR #3938, opened 2026-03-13
Improve test factories

hipek8:
Factories now initialize more fields (still some work to do).
Removed some 3-item iterators (this would prevent tests from detecting N+1 queries).
Factories are now used where .objects.create was used before.
Now tests are better at detecting problems with SQL queries and are less flaky.

### DIFF
--- a/src/ralph/api/tests/test_rendering.py
+++ b/src/ralph/api/tests/test_rendering.py
@@ -84,7 +84,7 @@ ALL_API_ENDPOINTS = {
     "racks": "/api/racks/",
     "regions": "/api/regions/",
     "server-rooms": "/api/server-rooms/",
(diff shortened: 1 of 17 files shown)


src/ralph/api/tests/test_rendering.py
```diff
@@ -84,7 +84,7 @@ class APIBrowsableClient(APIClient):
     "racks": "/api/racks/",
     "regions": "/api/regions/",
     "server-rooms": "/api/server-rooms/",
-    "services": "/api/services/",
+    "services": ("/api/services/", 25),
```
hipek8, line 87: This will go back to regular 20 in the next PR

awieckowski (approved):

PR #3951, opened 2026-07-24
Add experimental switchports feature

hipek8:
BREAKING: Removed unused data_center.Connection model

Can represent connections between data center assets (switch - server).

Steps:
 - configure rack switches (e.g. eth1 - switch123)
 - enter switch port for given server in rack
 - (optionally) validate with backend (internal tool – netmaker – supported only)


matyldv (approved):
🚀 can't wait to test it

